PR #10, opened 2025-01-31
add initial spec of the merkletree

arnaucube:
Add initial spec of the merkletree. The diagrams are hand-drawn to not spend much time on it, we can make them with drawio/figma in the future once more consolidated.

Also, at the bottom of the merkletree spec section, it notes that for the first iteration the implementation uses a hash of the key-values concatenated, with the idea of replacing it by the MerkleTree approach described above. This with the intention to leave space for using the hash approach in the first iteration of the project (first weeks), since the merkletree approach will require of a bit of more time to implement it (requires implementing this specific design of merkletree in rust, and in the future also the plonky2 circuit)


book/src/merkletree.md
```diff
@@ -1 +1,41 @@
...
+
+Suppose that the binary representation of the key `userPk` is `1011...`. This uniquely defines the leaf position that contains the public key of the authenticated user.
+
+![](img/merkletree1.png)
+
+
+## Proofs of inclusion and non-inclusion
+Merkle proofs contain the siblings along the path from the leaf to the root, where the leaf position is determined by the key binary representation. 
+
+The same approach works for proving that a key is not in the tree, and furthermore, that a value is not in the tree. The main difference is that it is proving that in the leaf position there is no key or value stored, and since leaf positions are deterministic based on the key, if at the position of the key `signerPk` the leaf is empty (non-existent), it means that the tree does not contain any leaf with that key.
```
ed255, line 33: To prove non-existence you can ether supply a merkle proof that follows the path of the key and arrives to
- an empty node
- a node which has a key different than the key you're proving non-existence of

The second case is missing from this description.

book/src/merkletree.md
```diff
@@ -1 +1,41 @@
...
+
+
+## Temporary first version
+The first iteration of the implementation uses a hash of the key-values concatenated, with the idea of replacing it by the MerkleTree approach described above.
+
+## Resources
+- [https://docs.iden3.io/publications/pdfs/Merkle-Tree.pdf](https://docs.iden3.io/publications/pdfs/Merkle-Tree.pdf)
+- [https://eprint.iacr.org/2018/955](https://eprint.iacr.org/2018/955)
```
ed255, line 41: Can you add a section called "Encoding" with a TODO to explain who key-values are encoded, and how nodes are encoded?

For example, how do you turn a key into field values
How do you turn a value into field values
How do you turn those field values into a leaf node
How do you take 2 leaf nodes and make an intermediate node
How do you encode the empty merkle tree
How do you encode an empty leaf

book/src/merkletree.md
arnaucube: both were there in the shape of _"proving that a key is not in the tree, and furthermore, that a value is not in the tree"_, but good point, it was not much clear. I've [rephrased it](https://github.com/0xPARC/pod2/pull/10/commits/f699013e4507f4f972f4af3579ecbcfec4e92d72) into two separated bullet points to make it more explicit.

book/src/merkletree.md
arnaucube: Right! added

book/src/merkletree.md
arnaucube: Right, the section was not clear, updated it: https://github.com/0xPARC/pod2/pull/10/commits/01fdce32c2d8a89c2b64b9d158f6b857ed7e4914

ed255 (approved):
LGTM!

PR #11, opened 2025-01-31
feat: move code here

ed255:


Cargo.lock
```diff
@@ -0,0 +1,639 @@
+# This file is automatically @generated by Cargo.
```
arnaucube, line 1: proposal to remove `Cargo.lock` since this is a lib. Also adding it to be ignored in the `.gitignore` file

ed255: done!

arnaucube (approved):

Issue #12, opened 2025-02-03
Define a errors strategy


arnaucube:
Currently we're using [anyhow](https://docs.rs/anyhow/latest/anyhow/) since it is the one used in Plonky2, but maybe also worth considreing [thiserror](https://docs.rs/thiserror/latest/thiserror/).

gubsheep:
No opinion, happy to defer to you guys

ed255:
For now we'll use [anyhow](https://docs.rs/anyhow/latest/anyhow/).
Since we don't have integrations with external libraries I think `anyhow` should work for us and we just add string descriptions to the errors.
Once we have integrations we could switch to [thiserror](https://docs.rs/thiserror/latest/thiserror/) so that consumers can parse the error and recover gracefully from different situations.

Note: in https://github.com/0xPARC/pod2/pull/23 I've already changed the SignPod and MainPod traits to use anyhow Result.

Closing this for now.

Issue #6, opened 2025-01-30
Specify a Merkle Tree


ed255:
The Merkle Tree must serve various cases:
- a. Potential build of the SignaturePod (to get a cryptographic hash to identify the Pod and to be signed)
- b. Inclusion / Non-inclusion of an element in a set
- c. Inclusion / Non-inclusion of an element in a list
- d. Inclusion of a key-value / Non-inclusion of an key in a map 

Is it a Sparse Merkle Tree?

arnaucube:
With the aim to keep the discussion rolling, the key-values of a pod in POD1 uses a tree where each key and each value is placed in a leaf, and the tree is build from the leafs to the root.
For example, if there are 5 key-values, the tree would have 10 leaves, five for the keys and five for the values, where each key-value pair are placed consecutively in the leaves.

![Image](https://github.com/user-attachments/assets/f95915c5-1b4f-4bb6-97aa-ab4a1cefdcf0)

Current POD2 implementation hashes the key-values concatenated without a tree, but as a temporary choice, with the idea of using a merkletree shape soon.

An approach that we've started discussing today is following the tree design used by iden3 https://docs.iden3.io/publications/pdfs/Merkle-Tree.pdf (or at least the idea of the item positions), where each leaf contains both the key-and-the-value, and each key-value leaf position is defined by the binary encoding of the `key` parameter.
In this way, in order to generate the proof or verify the merkleproof, it does not require the index of the leaf being proved, since it is already the key content of the key-value itself.

![Image](https://github.com/user-attachments/assets/728a2f52-1ea2-4d17-819c-15bf246d4c60)

Which actually in practice it would look more like:

![Image](https://github.com/user-attachments/assets/9438b777-bf52-4242-bf75-b708854efb6f)

Another related document: https://eprint.iacr.org/2018/955.

ed255:
Technically we could use a different tree for the Signature Pod and for the Inclusion/Non-inclusion statements, but I would prefer having the same to simplify the spec and implementation.

- Cons of POD1 MT:
  - You can have duplicate keys at different paths.  This means that case d. (Inclusion/Non-inclusion in a map) doesn't support a succint proof.
- Pros of using a SMT that uses sequential indexes as paths (the path is determined by the index bits starting by lowest)
  - It's always perfectly balanced
- Cons of SMT where keys determine paths:
  - A malicious user could find keys that have prefix collisions, making the tree very deep.

arnaucube:
Regarding the last point
[...]

PR #13, opened 2025-02-03
add initial MerkleTree implementation

arnaucube:
Add initial MerkleTree implementation, which is a wrapper on top of Plonky2's MerkleTree, with the idea that the future iteration will replace it by the MerkleTree specified at
https://0xparc.github.io/pod2/merkletree.html .

This PR relates to #6 .

Notes:
- initially we mentioned about doing a 'fake merkletree' where it would be an interface which internally just does hashes of the keyvalues, where a 'merkleproof' would be just the entire keyvalues so that the verifier takes it all and hashes it to check that the specific keyvalue is there. But after checking the Plonky2 existing merkletree, seemed more straightforward (and more close to our final merkletree) to just use it
- currently it does not implement the non-inclusion proofs, which in this version are just mock functions. This is, we can still use them for tests and building the rest of the system, but their verification will always pass
- this PR also updates the spec side (`book/src/merkletree.rs`), to reflect that at the initial version we use this wrapper, the first version will implement the tree defined in the spec, and future versions might go with Starky lookups to implement the tree defined in Aard's hackmd.
- for the methods that can return errors, I've used `anyhow` since is the one used in Plonky2, but maybe we want to explore other options (eg. `thiserror`). Opened a issue to tackle that: https://github.com/0xPARC/pod2/issues/12

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,4 +2,5 @@
 
 Specification can be found at the [`book`](https://github.com/0xPARC/pod2/tree/main/book) directory, a live rendered version at: https://0xparc.github.io/pod2/
 
-Requires Rust nightly: `rustup override set nightly`.
+## Usage
+- Run tests: `cargo test --release`
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,13 +2,19 @@ use hex::{FromHex, FromHexError};
 use plonky2::field::goldilocks_field::GoldilocksField;
 use plonky2::field::types::{Field, PrimeField64};
 use plonky2::hash::poseidon::PoseidonHash;
-use plonky2::plonk::config::Hasher;
+use plonky2::plonk::config::{Hasher, PoseidonGoldilocksConfig};
+use std::cmp::Ordering;
 use std::fmt;
 
 pub mod backend;
 pub mod frontend;
+pub mod merkletree;
 
 pub type F = GoldilocksField;
+/// C is the Plonky2 config used in POD2 to work with Plonky2 recursion.
+pub type C = PoseidonGoldilocksConfig;
+/// D defines the extension degree of the field used in the Plonky2 proofs (quadratic extension).
+pub const D: usize = 2;
 
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Eq)]
 pub struct Hash(pub [F; 4]);
@@ -37,6 +43,17 @@ impl FromHex for Hash {
             inner[i] = F::from_canonical_u64(u64::from_le_bytes(buf));
         }
         Ok(Self(inner))
+    }
+}
+
+impl Ord for Hash {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.to_string().cmp(&other.to_string())
+    }
+}
+impl PartialOrd for Hash {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
     }
 }
 
--- a/src/merkletree.rs
+++ b/src/merkletree.rs
@@ -1,0 +1,163 @@
+/// MerkleTree implementation for POD2.
+///
+/// Current implementation is a wrapper on top of Plonky2's MerkleTree, but the future iteration
+/// will replace it by the MerkleTree specified at https://0xparc.github.io/pod2/merkletree.html .
+use anyhow::{anyhow, Result};
+use itertools::Itertools;
+use plonky2::field::types::Field;
+use plonky2::hash::{
+    hash_types::HashOut,
+    merkle_proofs::{verify_merkle_proof, MerkleProof as PlonkyMerkleProof},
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -2,4 +2,6 @@
 
 Specification can be found at the [`book`](https://github.com/0xPARC/pod2/tree/main/book) directory, a live rendered version at: https://0xparc.github.io/pod2/
 
-Requires Rust nightly: `rustup override set nightly`.
+## Usage
+- Requires Rust nightly: `rustup override set nightly`
```
ed255, line 6: We can add a file like this https://github.com/openvm-org/openvm/blob/main/rust-toolchain.toml to set a particular rust nightly version and make jumping into the repo to develop more smooth.

ed255:
>     * this PR also updates the spec side (`book/src/merkletree.rs`), to reflect that at the initial version we use this wrapper, the first version will implement the tree defined in the spec, and future versions might go with Starky lookups to implement the tree defined in Aard's hackmd.

I would like to say that even if we don't have lookups the spec of the merkle tree may be the same.  On one hand we can specify the merkle tree without discussing depth, and then set a max accepted depth on top of it due to implementation limitations (but otherwise the spec remains the same).  Also we can realize Aard's implementation without lookups, just attaching a poseidon to each statement/operation pair, instead of attaching a merkle tree verification to each statement/operation pair.

src/lib.rs
```diff
@@ -2,13 +2,16 @@ use hex::{FromHex, FromHexError};
...
+use plonky2::plonk::config::{Hasher, PoseidonGoldilocksConfig};
 use std::fmt;
 
 pub mod backend;
 pub mod frontend;
+pub mod merkletree;
 
 pub type F = GoldilocksField;
+pub type C = PoseidonGoldilocksConfig;
+pub const D: usize = 2;
```
ed255, line 14: Can you write a doc line about what this value is?
For `C` and `D`, maybe use a more descriptive name?

src/merkletree.rs
```diff
@@ -0,0 +1,152 @@
...
+    existence: bool,
+    index: usize,
+    proof: PlonkyMerkleProof<F, <C as GenericConfig<D>>::Hasher>,
+}
+
+impl MerkleTree {
+    pub fn new(kvs: HashMap<Hash, Value>) -> Self {
+        let mut keyindex: HashMap<Hash, usize> = HashMap::new();
+        let mut leaves: Vec<Vec<F>> = Vec::new();
+        for (i, (k, v)) in kvs.into_iter().enumerate() {
```
ed255, line 42: HashMap iteration is not stable.  Can you use this to guarantee we always get the same order?
```rust
use itertools::Itertools;

... kvs.into_iter().sorted_by_key(|kv| kv.0) ...
```

src/merkletree.rs
```diff
@@ -0,0 +1,152 @@
...
+        root: Hash,
+        proof: &MerkleProof,
+        _key: &Hash,
+        _value: &Value,
+    ) -> Result<()> {
+        // mock method
+        if proof.existence {
+            anyhow!("expected proof of non-existence, found proof of existence");
+        }
+        Ok(())
```
ed255, line 104: Could you add a print to show a warning, so that we don't forget about this?
Also a comment with "TODO"?

src/lib.rs
ed255: Also perhaps move this `D` and `C` to the merkletree file?  Do you think these type and constant need to be accessed generally?

arnaucube: Cool, will add a comment to bring context; regarding moving them to the merkletree file, I think that not worth, since at least once we start adding circuits we will need them too.

src/merkletree.rs
arnaucube: Initially tried, but since `Hash` does not implement the sort algorithm it didn't work; but now added the `Ord` impl for the `Hash` type to make it work.
As a note, once we plug in the [merkletree](https://0xparc.github.io/pod2/merkletree.html) will not need the sorted approach, since it will be a tree deterministic tree from the key values, not based on the key insertion order.

src/merkletree.rs
arnaucube: Added the print with the warning. Regarding adding a "TODO", I would say that the whole merkletree is a "TODO" not just this method; this is, implementing the merkletree [specified in the docs](https://0xparc.github.io/pod2/merkletree.html), the current one is just for the current initial weeks so that we can focus on the rest of the system. So prefer to not add a "TODO" to this specific method, to avoid somebody understanding it as a task to be done (implementing the non-inclusion proofs in the plonky2's tree wrapper).

README.md
arnaucube: right, updated (removed the line from readme, added the `rust-toolchain.toml` file)

src/merkletree.rs
ed255: That sounds good!

Issue #7, opened 2025-01-30
Specify the list of Native statements & Native operations


ed255:
These are the native statements that are hardcoded by the system.
Each statement can be generated (proved) by running an operation (deduced) using previously proved statements.

Each native statement & native operations should specify how they are encoded (for example with a numeric id)

ax0:
The Devcon implementation has the following statement types:
- `None` - 'no statement' (used for padding)
  - This arises from the `None` operation of no arguments.
- `ValueOf` - statement expressing the fact that an anchored key (argument 1) corresponds to a value (argument 2)
  - This arises from the unary `NewEntry` operation with a key-value pair as its argument.
- `Equal` - equality of values associated with two anchored keys (its arguments)
  - This arises from the following two operations:
    - `EqualityFromEntries` (binary operation checking that its arguments are `ValueOf` statements with equal values)
    - `TransitiveEqualityFromStatements` (binary operation checking that its arguments are `Equal` statements with equal 'inner' anchored keys)
- `NotEqual` - non-equality of values associated with two anchored keys (its arguments)
  - This arises from the following operations:
    - `NonequalityFromEntries` (binary operation checking that its arguments are `ValueOf` statements with different values)
    - `GtToNonequality` (unary operation checking that its argument is a `Gt` statement)
    - `LtToNonequality` (unary operation checking that its argument is a `Lt` statement)
- `Gt` - greater-than relation between the values associated with its two anchored key arguments
  - This arises from the binary `GtFromEntries` operation with two `ValueOf` statements as its arguments checking that the value of the first is greater than that of the second.
- `Lt` - less-than relation between the values associated with its two anchored key arguments
  - This arises from the binary `LtFromEntries` operation with two `ValueOf` statements as its arguments checking that the value of the first is less than the that of the second.
- `Contains` - 'contains' relation between a vector value associated with an anchored key (argument 1) and a scalar value associated with another anchored key (argument 2)
  - This arises from the following operations:
    - `ContainsFromEntries` (binary operation checking that its arguments are `ValueOf` statements with the (vector) value of the first containing the (scalar) value of the second.
    - `RenameContainedBy` (binary operation checking that its first argument is a `Contains` statement and that its second is an `Equal` statement expressing equality of the first argument of `Contains` with some other element).
- `SumOf` - ternary relation asserting that a value associated with an anchored key (argument 1) is the sum of the values associated with other anchored keys (arguments 2 and 3)
[...]

PR #17, opened 2025-02-03
add SignerPod.id computation (integrate kvs with MerkleTree)

arnaucube:
( depends on #7 )

- Add SignerPod.id computation (integrate kvs with MerkleTree).
- Updates frontend::SignerPod to be a wrapper on top of
backend::SignerPod with extra metadata (a keymap between hashes and
their strings).
- Get's rid of SignerPod.compile() since now the frontend::SignerPod
uses the method `::new()` which internally calls
backend::SignerPod::new which constructs the merkletree to use it's root
as PodID.


src/frontend.rs
```diff
@@ -71,27 +71,42 @@ impl fmt::Display for Value {
...
 #[derive(Clone, Debug)]
 pub struct SignedPod {
-    pub params: Params,
-    pub id: PodId,
-    pub kvs: HashMap<String, Value>,
+    pub pod: backend::SignedPod,
+    // `string_key_map` is a hashmap to store the relation between key strings and key hashes
+    // TODO review if maybe store it as <Hash, String>, so that when iterating we can get each
+    // hash respective string
+    string_key_map: HashMap<String, crate::Hash>,
```
arnaucube, line 82: added it as `HashMap<String, Hash>` but with the `TODO` note, since currently we're not using this mapping anywhere so didn't know what would be more useful

ed255 (approved):
LGTM!

Issue #8, opened 2025-01-30
What's first, the Statement or the Operation?


ed255:
Approach a) So far the design considers that the creator of a pod2 doesn't write statements, instead the creator performs operations using previous statements to generate new statements.  This reminds me of [constructive proofs](https://en.wikipedia.org/wiki/Constructive_proof)

Approach b) A different approach would be that the creator writes a statement, and then provides the proof for its truth (the operation that can generate it, or the sequence of operations that can generate it).

Which one is better for pod2?  Here are some thoughts:
- In approach b) for all the native operations that are of the form `...FromEntries` the operation could be generated automatically very easily:  it's just referencing the anchored values used in the statement and applying the matching `...FromEntries` operation.
- approach a) resembles very much the verification process of the statements
- approach a) may be more annoying for custom statements.  I would find it more natural to start writing the statement and then figure out how to arrive there.

ax0:
In the Devcon implementation we require that the user specify the operations they wish to apply to the statements of the input PODs, so essentially approach A, however the general idea was that you had some output statements in mind and devised the operations necessary to produce them, so also approach B. However, that implementation did not allow for intermediate statements (or custom statements) where I think approach B is more natural, so my vote goes to approach B.

gubsheep:
I think Approach A still works with intermediate statements if we have some way to specify that some operation execution results in private (intermediate) statements?

But overall, no strong opinions and happy to defer to you guys.

ed255:
In https://github.com/0xPARC/pod2/pull/18 I went for approach A (in terms of the facilities that the frontend library offers).
I can imagine in the future that PEX is used to build the statements+operations (either with approach A or B) and the frontend library just accepts pairs of statement,operation.  That would be very simple to have by using the method from the MainPodBuilder (or a similar one):
```rust
    fn push_st_op(&mut self, st: middleware::Statement, op: middleware::Operation) {
        self.statements.push(st);
        self.operations.push(op);
    }
```

I'll close this issue for now because we are going to use approach A.  Feel free to reopen in the future.

PR #18, opened 2025-02-03
feat: add middleware and signer traits

ed255:
I've restructured the code a little bit like this:
- middleware: types and traits that frontend and backends use to communicate between themselves
  - I've also added some small common functions, but maybe they should go to lib.rs?  I don't know
- frontend: accesses a backend via trait
- backends: signs/proves stuff that the frontend sends encoded via middleware types

Currently there's only 1 implementation of the backend: the MockSignedPod wich is a type of Signed Pod.  The trait may require a few touches after we start integrating things

Missing things from this PR
- ~~The traits for MainPods~~
- ~~Integrating the SignedPod trait into the frontend properly (so that creating a new MainPod takes SignedPod trait objects)~~
- Use the merkle tree for the id of the MockSignedPod (then it's more similar to the real thing)
- ~~Rewrite the MainPodBuilder~~


src/frontend.rs
```diff
@@ -72,271 +76,416 @@ impl fmt::Display for Value {
...
+//     ProductOf = 9,
+//     MaxOf = 10,
+// }
+
+// impl From<NativeStatement> for middleware::NativeStatement {
+//     fn from(v: NativeStatement) -> Self {
+//         Self::from_repr(v as usize).unwrap()
+//     }
+// }
+
```
arnaucube, line 155: maybe can be removed?

src/frontend.rs
```diff
@@ -72,271 +76,416 @@ impl fmt::Display for Value {
...
+    pub fn new(params: &Params) -> Self {
+        Self {
+            params: params.clone(),
+            kvs: HashMap::new(),
+        }
     }
-    pub fn id(&self) -> PodId {
-        self.pod.id
+
+    pub fn insert(&mut self, key: impl Into<String>, value: impl Into<Value>) {
```
arnaucube, line 93: do you envision some use case where it needs to insert new keyvalues after the pod creation? (after `SignedPodBuilder::new()`)

Furthermore, what happens when a Pod is signed and the user inserts a new key-value to it? the signature would not verify anymore for that new version of the pod

src/frontend.rs
```diff
@@ -72,271 +76,416 @@ impl fmt::Display for Value {
...
+pub struct SignedPod {
+    pub pod: Box<dyn middleware::SignedPod>,
+    /// HashMap to store the reverse relation between key strings and key hashes
+    pub key_string_map: HashMap<Hash, String>,
 }
 
-impl From<NativeStatement> for backend::NativeStatement {
-    fn from(v: NativeStatement) -> Self {
-        Self::from_repr(v as usize).unwrap()
+impl SignedPod {
```
arnaucube, line 122: How would you see merging the `SignedPodBuilder` into `SignedPod`?
where the `SignedPodBuilder::new()` and `SignedPodBuilder.sign()` methods would be just done as part of `SignedPod::new()`.
(and getting rid of the `.insert()` method, which introduces cases where an already signed pod gets new keyvalues inserted)

In this way the interface gets more direct to create pods, with just
```rust
pub struct SignedPod {
    pub fn new<S: PodSigner>(
        params: &Params,
        signer: &mut S,
        kvs: HashMap<String, Value>
    ) -> Self;

    // and the already existing methods
    // id(), origin(), verify(), kvs()
```

arnaucube (approved):
LGTM!
added some comments but are not blockers for merge, just to explore other code structures

src/frontend.rs
ed255: I would recommend taking a look at the entire file (not via diff).
A pod creation is done when calling `SignedPodBuilder::sign` which returns a `SignedPod`.
The `SignedPodBuilder` contains a SignedPod in "construction" and it's not signed, so you can't add a key-value to a signed pod with this API.
The reason to have the builder is to offer a nice API where you can do this:
```rust
        let mut gov_id = SignedPodBuilder::new(params);
        gov_id.insert("idNumber", "4242424242");
        gov_id.insert("dateOfBirth", 1169909384);
        gov_id.insert("socialSecurityNumber", "G2121210");
        let gov_id = gov_id.sign(&mut signer); // The type has changed now
```
Notice how in the insert method we can pass different types.  This is because each argument is generic `key: impl Into<String>, value: impl Into<Value>` so you don't have to write `.into()` for each argument and we can use the `From<...> for Value` trait implementations.

src/frontend.rs
ed255: Please see my comment https://github.com/0xPARC/pod2/pull/18#discussion_r1940820395

src/backend.rs
```diff
@@ -1,71 +1,15 @@
+// TODO: Move the SignedPod.id calculation to mock_signed
+// TODO: Move the MainPod logic to mock_main and implement the MainPod trait
+/*
```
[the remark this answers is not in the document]
ed255: But most of this file will actually go to `src/backends/mock_main.rs`, did you mean to move it there?

src/frontend.rs
arnaucube: (assuming that we can find a way for the hashmap to take as values the different types)

ed255: This doesn't work because kvs is type `HashMap<String, Value>` and you're passing keys as `&str` and values as `&str` and `integer`.
You'd have to do this:
```rust
        kvs.insert("idNumber".into(), "4242424242".into());
        kvs.insert("dateOfBirth".into(), 1169909384.into());
        kvs.insert("socialSecurityNumber".into(), "G2121210".into());
```

arnaucube: well, yes, even with the `.into()`, seems preferable to do it this way so that we don't need the "Builder" type and things are more simple, no?
To be clear, what I'm suggesting is that we can get rid of the `SignedPodBuilder` (with the methods `new`, `insert`, and `sign`), and just add a method `new` in the `SignedPod`, achieving a similar API as the one you showed in the comment, but with one struct less.

src/backend.rs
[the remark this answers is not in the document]
arnaucube: it was to avoid having a file that it's just commented out, but I guess that it's fine since it's a WIP process

src/frontend.rs
ed255: I personally like this pattern, and I like being able to insert the key values without the `.into()`.  For example if we didn't have the `SignedPodBuilder` I would probably write a small utility function to fill in the `HashMap` with `key: impl Into<String>, value: impl Into<Value>` for my convenience.
But I agree that things would be more simple without this type; it's just that I see it as a small type so the tradeoff looks good to me.
Also I implemented the same pattern with the `MainPod`.

What do you think about the `MainPodBuilder`?

arnaucube: +1 on the utility function to fill the hashmap with `impl Into<String/Value>` to avoid needing the `.into()` all the time.
Still see value on getting rid of the `SignedPodBuilder` type, but happy to merge this PR as it is and we can explore it in a future iteration.
Regarding the MainPod related structs, I don't have an opinion yet (I saw that there are even more helper types (builder and compiler), but so far it looks like it fits well (which then could justify following the same pattern for the SignedPod).

ax0 (approved):
LGTM.

src/frontend.rs
```diff
@@ -136,207 +140,333 @@ pub struct AnchoredKey(pub Origin, pub String);
...
-    pub fn origin(&self) -> Origin {
-        Origin(PodType::Main, self.id)
+    pub fn id(&self) -> PodId {
+        self.pod.id()
     }
-
-    pub fn compile(&self) -> Result<backend::MainPod> {
-        MainPodCompiler::new(self).compile()
+    pub fn origin(&self) -> Origin {
+        Origin(PodClass::Signed, self.id())
```
gubsheep, line 350: should this be PodClass::Main?

src/frontend.rs
```diff
@@ -136,207 +140,333 @@ pub struct AnchoredKey(pub Origin, pub String);
...
+pub struct Operation(pub NativeOperation, pub Vec<OperationArg>);
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct MainPodBuilder {
     pub params: Params,
+    pub input_signed_pods: Vec<SignedPod>,
+    pub input_main_pods: Vec<MainPod>,
     pub statements: Vec<Statement>,
     pub operations: Vec<Operation>,
```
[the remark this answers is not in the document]
ed255: I though that part of the `MainPodBuilder` could be agnostic to the decision from https://github.com/0xPARC/pod2/issues/8
So internally it keeps the statements and operations (which is what the backend will eventually need), but the current `MainPodBuilder` has a method where you just pass an operation and it stores the operation along with the statement (following approach A): 
https://github.com/0xPARC/pod2/blob/4d16647d10688eb003aaacc2a4610b6b3b11a1f7/src/frontend.rs#L281

There's also a small detail: in the current implementation you can use constants in all operations, and the `MainPodBuilder` will internally generate a `ValueOf` statement and reference it automatically.  For this reason it's also convenient to keep statements and operations as this processing can be done on the fly.

src/frontend.rs
ed255: yes!

ed255: I've pushed the fix in this PR https://github.com/0xPARC/pod2/pull/23
https://github.com/0xPARC/pod2/pull/23/commits/7dc9597de5d42195ccfdcc4a97e4e61b8297f4c4

Issue #5, opened 2025-01-30
Specify Value


ed255:
So far we have determined that at the frontend level we have 3 value types:
- Integer
- String
- Merkle Tree (related issue: https://github.com/0xPARC/pod2-docs/issues/6 )

What is the spec of each one?
How are they encoded for hashing purposes / circuit purposes?

ax0:
Some thoughts based on the Devcon implementation and POD 1:
- Strings and Merkle tree roots would be represented as hashes. In the context of Plonky2, this means 4 field elements for the required security level.
- Integers in the context of Plonky2 would have to contain less than 64 bits to fit in a single field element, but there are applications (e.g. involving timestamps) that will require 64 bits, so one approach would be to represent 64-bit integers as an appropriate pair of 32-bit field elements.

Keeping these two points in mind and assuming we are using Plonky2, I think it would make sense to encode all of these value types as 4 element arrays of field elements in little endian form.

ed255:
> Keeping these two points in mind and assuming we are using Plonky2, I think it would make sense to encode all of these value types as 4 element arrays of field elements in little endian form.

This sounds good to me.
I also prefer having 64 bit integers rather than 32 bit integers.  Encoding the 64 bit integer as 4 32-bit limbs (with the 2 upper limbs at 0) sounds good to me.
I was also considering defining the integer as signed so that we can cover more use cases.

gubsheep:
I don't have strong opinions here and am happy to defer to you guys.

One weak suggestion is that perhaps we can have a data type that refers to a "unique cryptographic identifier" - this means any public key, hash, merkle root, etc. It seems like this is the root abstraction that is independent of cryptographic protocol / hash function / etc.

gubsheep:
Now that I think about this more, I'm thinking that "Merkle Tree" should maybe not be a frontend data type but rather an underlying implementation detail. And the actual data type is something like "set" "ordered list" or "dictionary," it just happens to be merkelized under the hood.

ed255:
> Now that I think about this more, I'm thinking that "Merkle Tree" should maybe not be a frontend data type but rather an underlying implementation detail. And the actual data type is something like "set" "ordered list" or "dictionary," it just happens to be merkelized under the hood

That sounds great to me.

> with the merkelization scheme depending on the POD type.

I would try to have a single merkelization scheme, except for introductory gadgets which can use whatever fits.  But for our own design I would try to keep the same spec for a Merkle Tree everywhere.

[...]

PR #19, opened 2025-02-04
docs: add POD value types

ax0:
This PR documents the POD value types as discussed in #5.


book/src/values.md
```diff
@@ -0,0 +1,30 @@
...
+$$x = x_0 + x_1 \cdot 2^{32}$$
+with $0 \leq x_0, x_1 < 2^{32}$ and representing it as
+$$\texttt{map}\ \iota\ [x_0, x_1, 0, 0],$$
+where $\iota:\mathbb{N}\rightarrow\texttt{GoldilocksField}$ is the canonical projection.
+
+## `String`
+In the frontend, this type corresponds to the usual `String`. In the backend, the string will be mapped to a sequence of field elements[^String] and hashed with the hash function employed there, thus being represented by its hash.
+
+## `MerkleTree`
+In the front end, this type encapsulates a sparse Merkle tree. In the backend, this will be represented by the root hash of the tree, which will be of the type of the output of the hash function employed.
```
arnaucube, line 26: is it or maybe the reverse? like in the frontend the `MerkleTree` is just the root, and in the backend it is the whole merkletree structure

arnaucube (approved):
Nice!

book/src/values.md
ed255: I think it can be seen both ways.
If you think about the `Value`, in the backend a `MerkleTree` `Value` is just the root.
In the frontend when we have a `MerkleTree` Value we can attach metadata (the entire tree, which is actually implemented beyond the frontend, probably the backend)

PR #21, opened 2025-02-04
docs: add native statement table

ax0:
This PR documents the native statements as discussed in #7.

### DIFF
--- a/book/src/statements.md
+++ b/book/src/statements.md
@@ -1,1 +1,45 @@
 # Statements
+The claims asserted by a POD are referred to as its *statements*. These statements introduce values and express relations between them, where the values may or may not be part of the same POD. The mechanism for referring to values in arbitrary PODs is furnished by *anchored keys*.
+
+## Anchored keys
+Rather than dealing with just keys, we introduce the notion of an *anchored key*, which is a pair consisting of an origin specifier and a key, i.e.
+
+```
+type AnchoredKey = (Origin, Key)
+type Key = String
+```
+
+An *origin* is a triple consisting of a numeric identifier called the *origin ID*, a string called the *origin name* (omitted in the backend) and another numeric identifier called the *gadget ID*, which identifies the means by which the value corresponding to a given key is produced.
+
+The origin ID is defined to be 0 for 'no origin' and 1 for 'self origin', otherwise it is the content ID[^content-id] of the POD to which it refers. The origin name is not cryptographically significant and is merely a convenience for the frontend.
+
+The gadget ID takes on the values in the following table:
+
+| Gadget ID | Meaning                                                                                   |
+|-----------|-------------------------------------------------------------------------------------------|
+| 0         | no gadget                                                                                 |
+| 1         | `SignedPOD` gadget: The key-value pair was produced in the construction of a `SignedPOD`. |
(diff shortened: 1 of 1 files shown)


arnaucube (approved):
Great!
This spec would have been handy when I was starting to discover the PODs system hehe

book/src/statements.md
```diff
@@ -1 +1,45 @@
...
+```
+type AnchoredKey = (Origin, Key)
+type Key = String
+```
+
+An *origin* is a triple consisting of a numeric identifier called the *origin ID*, a string called the *origin name* (omitted in the backend) and another numeric identifier called the *gadget ID*, which identifies the means by which the value corresponding to a given key is produced.
+
+The origin ID is defined to be 0 for 'no origin' and 1 for 'self origin', otherwise it is the content ID[^content-id] of the POD to which it refers. The origin name is not cryptographically significant and is merely a convenience for the frontend.
+
+The gadget ID takes on the values in the following table:
```
ed255, line 16: I would like to discuss the idea of having Pod Class and Pod Type (Gadget Class and Gadget ID).
Gadget Class would be None, Signed, Main
Then within each class we have actual gadgets
- Signed: MockSignPod, SchnorrPod, FalconPod, Plonky2SignPod
- Main: MockMainPod, Plonky2Pod

book/src/statements.md
```diff
@@ -1 +1,45 @@
...
+| Code | Identifier  | Args                | Meaning                                                           |
+|------|-------------|---------------------|-------------------------------------------------------------------|
+| 0    | `None`      |                     | no statement (useful for padding)                                 |
+| 1    | `ValueOf`   | `ak`, `value`       | `value_of(ak) = value`                                            |
+| 2    | `Equal`     | `ak1`, `ak2`        | `value_of(ak1) = value_of(ak2)`                                   |
+| 3    | `NotEqual`  | `ak1`, `ak2`        | `value_of(ak1) != value_of(ak2)`                                  |
+| 4    | `Gt`        | `ak1`, `ak2`        | `value_of(ak1) > value_of(ak2)`                                   |
+| 5    | `LEq`       | `ak1`, `ak2`        | `value_of(ak1) <= value_of(ak2)`                                  |
+| 6    | `Contains`  | `ak1`, `ak2`        | `(key_of(ak2), value_of(ak2)) ∈ value_of(ak1)` (Merkle inclusion) |
+| 7    | `Sintains`  | `ak1`, `ak2`        | `(key_of(ak2), value_of(ak2)) ∉ value_of(ak1)` (Merkle exclusion) |
```
ed255, line 40: Here we use the AnchorKey Key as the Dictionary key (Merkle Tree path).
The AnchorKey Key is a Hash coming from a String.
But in the circuit there's no check that the Hash comes from a String.
It could be useful to have Dictionaries where the key doesn't come from a String.  It could be an integer for example, so then I could use the Dictionary as an array.
Or it could be the value itself, then I could use the Dictionary as a Set.
To use the Dictionary as an array I would need to create an AnchorKey where the Key is an integer.
To use the Dictionary as a Set I would need to create an AnchorKey where the Key is a value, and show that `key_of(ak) = value_of(ak)`.

I wonder if we can get the functionality of Dictionary, Array and Set (inclusion and non-inclusion) with less than 6 statements.

ed255 (approved):
LGTM! Left some comments for further discussion but we could merge this already.

PR #22, opened 2025-02-04
docs: add native operation table

ax0:
This PR documents the native operations as discussed in #7 .


book/src/operations.md
```diff
@@ -1 +1,29 @@
 # Operations
+The mechanism by which statements are derived is furnished by *operations*. Roughly speaking, with few exceptions, an operation deduces a statement from one or more existing statements according to some relation that must be satisfied between these statements. For example, if `Equal(ak1, ak2)` holds true, then the operation `SymmetricEq` applied to this statement yields `Equal(ak2, ak1)`.
+
+More precisely, an operation is a code (or, in the frontend, string identifier) followed by 0 or more arguments. These arguments may consist of up to three statements, up to one key-value pair and up to one Merkle proof.
```
ed255, line 4: > These arguments may consist of up to three statements, up to one key-value pair and up to one Merkle proof.

Do you think we need to specify these limits?  Will they make the circuit simpler / have less constraints?  Or is it just an observation?

book/src/operations.md
```diff
@@ -1 +1,29 @@
...
+
+The following table summarises the natively-supported operations:
+
+| Code | Identifier         | Args                | Condition                                                                                                             | Output                                                         |
+|------|--------------------|---------------------|-----------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------|
+| 0    | `None`             |                     |                                                                                                                       | `None`                                                         |
+| 1    | `NewEntry`         | `(key, value)`      |                                                                                                                       | `ValueOf(ak, value)`, where `ak` has key `key` and origin ID 1 |
```
ed255, line 11: This one is interesting.  The operation includes the `(key, value)` in its arguments so that it can generate a statement with that `key` and `value`.
If we think about the circuit, what is public is the statement and not the operation.  So we need the `key, value` be part of the `ValueOf` statement.  Then, in the circuit, do we need the `key, value` in the operation?  I don't think so.  The `key, value` in the `ValueOf` is just unconstrained witness data, it doesn't come from anywhere.

So it seems to me that in the language level a `NewEntry` has arguments `(key, value)` but in the circuit `NewEntry` doesn't have any argument.

book/src/operations.md
```diff
@@ -1 +1,29 @@
...
+| 3    | `SymmetricEq`      | `s`                 | `s = Equal(ak1, ak2)`                                                                                                 | `Equal(ak2, ak1)`                                              |
+| 4    | `TransitiveEq`     | `s1`, `s2`          | `s1 = Equal(ak1, ak2)`, `s2 = Equal(ak3, ak4)`, `ak2 = ak3`                                                           | `Equal(ak1, ak4)`                                              |
+| 5    | `EntryNEq`         | `s1`, `s2`          | `s1 = ValueOf(ak1, value1)`, `s2 = ValueOf(ak2, value2)`, `value1 != value2`                                          | `NotEqual(ak1, ak2)`                                           |
+| 6    | `SymmetricNEq`     | `s`                 | `s = NotEqual(ak1, ak2)`                                                                                              | `NotEqual(ak2, ak1)`                                           |
+| 7    | `GtToNonequality`  | `s`                 | `s = Gt(ak1, ak2)`                                                                                                    | `NotEqual(ak1, ak2)`                                           |
```
ed255, line 17: I observe 3 different ways to write the equality/non-equality part of a statement/operation:
- Eq, Equal
- NEq, NotEqual, Nonequality
Can we unify them and pick 1 for equality and 1 for nonequality and use it everywhere?

book/src/operations.md
ax0: More an observation than anything else.

ed255 (approved):
LGTM!

PR #23, opened 2025-02-04
feat: integrate mt in mock_sign

ed255:
Just moved the logic of calculating the id of a signed pod into the MockSignedPod.
Also updated the SignedPod, MainPod traits to return Result on the sign/prove operations.

### DIFF
--- a/src/frontend.rs
+++ b/src/frontend.rs
@@ -347,7 +347,7 @@ impl MainPod {
         self.pod.id()
     }
     pub fn origin(&self) -> Origin {
-        Origin(PodClass::Signed, self.id())
+        Origin(PodClass::Main, self.id())
     }
 }
 
--- a/src/frontend.rs
+++ b/src/frontend.rs
@@ -315,8 +315,7 @@ impl MainPodBuilder {
             statements: &self.statements,
             operations: &self.operations,
         };
-        // TODO: Error handling
-        let (statements, operations) = compiler.compile(inputs).expect("TODO");
+        let (statements, operations) = compiler.compile(inputs)?;
 
         let inputs = MainPodInputs {
             signed_pods: &self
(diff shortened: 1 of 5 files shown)


src/merkletree.rs
```diff
@@ -127,6 +127,10 @@ impl MerkleTree {
     pub fn iter(&self) -> std::collections::hash_map::Iter<Hash, Value> {
         self.kvs.iter()
     }
+
+    pub fn kvs(&self) -> &HashMap<Hash, Value> {
```
ed255, line 131: To verify the Pod ID of a MockSignedPod we need to recalculate the root.  To recalculate the root we need to create a new MerkleTree object.  The MerkleTree constructor takes a HashMap, so I found it easier to add this method to easily recalculate the root of an existing MerkleTree object like this:
```rust
        let new_mt = MerkleTree::new(mt.kvs());
        let new_id = new_mt.root();
```

In the future we can have a generic MerkleTree constructor that takes an iterator, and then we can do
```rust
        let new_mt = MerkleTree::new(mt.iter());
        let new_id = new_mt.root();
```

src/frontend.rs
```diff
@@ -307,17 +307,18 @@ impl MainPodBuilder {
...
+    pub fn prove<P: PodProver>(&self, prover: &mut P) -> Result<MainPod> {
         let compiler = MainPodCompiler::new(&self.params);
         let inputs = MainPodCompilerInputs {
             // signed_pods: &self.input_signed_pods,
             // main_pods: &self.input_main_pods,
             statements: &self.statements,
             operations: &self.operations,
         };
+        // TODO: Error handling
         let (statements, operations) = compiler.compile(inputs).expect("TODO");
```
arnaucube, line 319:
```suggestion
        let (statements, operations) = compiler.compile(inputs)?;
```

arnaucube (approved):
LGTM!

src/frontend.rs
ed255: well spotted!  fixed

PR #30, opened 2025-02-05
Aardocs

tideofwords:
Some updates to docs:

Replace "MerkleTree" with "Dictionary" as a front-end type
New doc to clarify how a dictionary is converted to a Merkle tree
Draft doc on Statements.  The doc is incomplete: the Merkle tree statements need to take depth as an entry, and I haven't thought through the best design for this.


book/src/statements.md
```diff
@@ -1 +1,127 @@
 # Statements
```
ed255, line 1: @ax0 already wrote a statements doc entry here https://github.com/0xPARC/pod2/blob/main/book/src/statements.md
So this PR has a conflict with that file.
Could you try to merge the contents of this file with the already existing ones?

tideofwords: OK, will do :)

Issue #26, opened 2025-02-04
Specify array, set, dictionary


arnaucube:
Specify array, set, dictionary and how they use the merkletree under the hood.

tideofwords:
I added some docs based on the idea that there is only one non-atomic type -- dictionary.  Do we want to add array and set as well?

arnaucube:
From what I understood, the system uses the three similar types (array, set, dictionary). While all of them can use a merkletree under the hood, each of them uses it in a specific way:
- **dictionary**: the user original keys and values are hashed to be used in the leaf.
    - `leaf.key=hash(original_key)`
    - `leaf.value=hash(original_value)`
- **array**: the elements are placed at the value field of each leaf, and the key field is just the array index (integer)
    - `leaf.value=original_value` 
    - `leaf.key=i` 
- **set**: the value field of the leaf is unused, and the key contains the hash of the element
    -  `leaf.key=hash(original_value)`
    - `leaf.value=0`

In the three types, the merkletree under the hood allows to prove inclusion & non-inclusion of the particular entry of the {dictionary/array/set} element.

What I don't know yet, is POD use cases that require each of them.

arnaucube:
Also, @ed255 , in the current code structure, should the `dictionary`,`array`,`set` types (and their logic) belong to the frontend or backend?

ed255:
> Also, [@ed255](https://github.com/ed255) , in the current code structure, should the `dictionary`,`array`,`set` types (and their logic) belong to the frontend or backend?

Probably in the frontend.  But there's an open question about the statements that deal with inclusion and non-inclusion.  If we can find a way to make those statements generic for all `dictionary`, `array` and `set`, then the backend doesn't need to know about them.  But if we make different statements for each container type then the backend needs to be a aware of those container types.

For example let's say that we only have two statements:
`contains(root, key, value)`
`sintains(root, key)`
And then we support each container type with some extra logic:
```
# dictionary
- contains(root, key, value)
- sintains(root, key value)
# array
- contains(root, index, value)
# set
- hash(value, value_hash)
- contains(root, value_hash, 0)
- sintains(root, value_hash)
```
In this case the backend doesn't need to now about array, dict or set distinction.

But if we have these statements:
```
# dictionary
- contains(root, key, value)
- sintains(root, key value)
# array
- contains(root, index, value)
# set
- set_contains(root, value)
- set_sintains(root, value)
```
[...]

Issue #2, opened 2025-01-30
Specify how SignaturePod.id is calculated


ed255:
SignaturePod contains a key-value map where keys are strings.
How is the PodId calculated?
Is the PodId deterministic given the key-values?
Is the signer an entry in the key-values?

Ideas:
- Make pairs or all key-values, sort by key, encode and hash
- sort key-values by key, encode each pair, and merkelize them with sequential indexes as paths
- build a merkle tree where each leaf is an encoded key-value and the path is hash(key)

gubsheep:
What is the alternative for referencing and operating on signer if it's not a KV entry?

The current Devcon implementation uses the hardcoded string "_signer" as a special reserved keyname for a KV entry that stores the signer public key. This enables the _signer key to be operated on using logical deductions just as any other entry; for example, proving that a signer is contained in a list, or that it's equal / not equal to another field that stores a public key, etc. This is the simplest solution I can currently think of but open to any others.

ax0:
> SignaturePod contains a key-value map where keys are strings. How is the PodId calculated? Is the PodId deterministic given the key-values? Is the signer an entry in the key-values?
> 
> Ideas:
> 
> * Make pairs or all key-values, sort by key, encode and hash
> * sort key-values by key, encode each pair, and merkelize them with sequential indexes as paths
> * build a merkle tree where each leaf is an encoded key-value and the path is hash(key)

The Devcon implementation essentially lexicographically sorts the statements according to their underlying keys and forms the ID as a hash of the concatenation of them (with appropriate padding). It seems natural to replace this linear approach with a sparse Merkle tree. As for including the signer key, another reason to do so is that it would naturally tie the ID to the signer of the POD.

arnaucube:
+1, from what I understand it would be:
- with the merkletree that we have specified ([pod2/merkletree.html](https://0xparc.github.io/pod2/merkletree.html)), the root is deterministic from the key-values (since leaf position is determined by the `hash(key)` binary representation), so a set of key-values always land into the same root
- then we can use the root as `POD.id` of the POD
- and in a SignedPOD, the `_signerKey` key-value is always stored in one of the leaves of the tree

arnaucube:
Closing as it was done at #17 and #31

PR #31, opened 2025-02-05
Extend merkletree spec, init SignedPod section, add typos checker in CI

arnaucube:
- extend merkletree spec, converting old hand-drawn diagrams to drawio diagrams, and adding new diagrams (related: #6)
- init SignedPod section (related: #2)
- initial draft of the types dictionary, set, array (related: #26)
- add typos checker in CI (and correct the ones that were detected)

Note on drawio diagrams: each image file contains the metadata to edit the diagram in the draw.io website.


.github/workflows/typos.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  typos:
+    if: github.event.pull_request.draft == false
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Use typos with config file
+      uses: crate-ci/typos@master
```
ed255, line 16: :smiley:

book/src/datatypes.md
```diff
@@ -1 +1,17 @@
 # Data types
+
+
+## Dictionary, array, set
```
ed255, line 4: Maybe this can be merged into https://github.com/0xPARC/pod2/blob/main/book/src/values.md ?
And replace the `MerkleTree` with these three?

book/src/merkletree.md
```diff
@@ -16,30 +29,63 @@ So for example, suppose we have the following data in a POD:
...
 }
 ```
 
 The merkletree will contain the key values from the `kvs` field.
 
-Suppose that the binary representation of the key `userPk` is `1011...`. This uniquely defines the leaf position that contains the public key of the authenticated user.
+Suppose that the binary representation of the key `userPk` is `1011...`. This uniquely defines the leaf position that contains the public key of the authenticated user. Similarly for the other key-values:
 
-![](img/merkletree1.png)
+![](img/merkletree-example-2-b.png)
```
ed255, line 41: In the diagrams I was a bit confused by the binary representation of the keys.
For example key=4 (in base 10), should be key=100 (in base 2), and then the binary encoding in little endian is [0, 0, 1].
The reason for that is that we always write numbers with the higher digits on the left and lower digits on the right.

book/src/datatypes.md
arnaucube: Right makes sense, will move it there.

book/src/merkletree.md
arnaucube: oh right. I was trying to put directly the little-endian encoding to make it easier to see in the diagram why that key goes to through that path, but the `_b` suffix mixes types incorrectly; will update it to make it correct.

arnaucube: done: https://github.com/0xPARC/pod2/pull/31/commits/7f3a769e07f7a21fc807c41d1b9c3435a32f07bc

ed255 (approved):
LGTM!

PR #35, opened 2025-02-05
compute MainPod.id from pub_statements; and introduce the trait `ToFields`

arnaucube:
Compute MainPod.id from pub_statements hash; for it, introduce the trait `ToFields` at the middleware and implement it for the Statement related types.

### DIFF
--- a/src/backends/mock_main.rs
+++ b/src/backends/mock_main.rs
@@ -1,9 +1,11 @@
 use crate::middleware::{
-    self, MainPod, MainPodInputs, NativeOperation, NativeStatement, NoneMainPod, NoneSignedPod,
-    Params, PodId, PodProver, SignedPod, Statement, StatementArg,
+    self, Hash, MainPod, MainPodInputs, NativeOperation, NativeStatement, NoneMainPod,
+    NoneSignedPod, Params, PodId, PodProver, SignedPod, Statement, StatementArg, ToFields,
 };
 use anyhow::Result;
 use itertools::Itertools;
+use plonky2::hash::poseidon::PoseidonHash;
+use plonky2::plonk::config::Hasher;
 use std::any::Any;
 use std::io::{self, Write};
 
@@ -156,7 +158,7 @@ impl MockMainPod {
         }
     }
 
-    fn process_priavte_statements_operations(
+    fn process_private_statements_operations(
         params: &Params,
         statements: &[Statement],
         input_operations: &[middleware::Operation],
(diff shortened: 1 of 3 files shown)


src/backends/mock_main.rs
```diff
@@ -265,6 +267,18 @@ impl MockMainPod {
...
 
+pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middleware::Hash> {
+    let (field_elems, _) = statements
+        .into_iter()
+        .map(|statement| statement.clone().to_fields())
+        .fold((Vec::new(), 0), |mut acc, (f, l)| {
+            acc.0.extend(f);
+            acc.1 += l;
+            acc
+        });
```
ax0, line 278:
```suggestion
    let field_elems = statements
        .into_iter()
        .flat_map(|statement| statement.clone().to_fields().0)
        .collect::<Vec<_>>();
```

ax0 (approved):
LGTM!

src/backends/mock_main.rs
arnaucube: oh right, more clean! thanks! ^^

src/backends/mock_main.rs
```diff
@@ -265,6 +267,14 @@ impl MockMainPod {
     }
 }
 
+pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middleware::Hash> {
+    let field_elems = statements
+        .into_iter()
+        .flat_map(|statement| statement.clone().to_fields().0)
+        .collect::<Vec<_>>();
+    Ok(Hash(PoseidonHash::hash_no_pad(&field_elems).elements))
```
ed255, line 275: Should we use a padding here?  Maybe we can revisit this in the future (for all hashes, not just this one)

src/middleware.rs
```diff
@@ -280,6 +299,36 @@ impl StatementArg {
     }
 }
 
+impl ToFields for StatementArg {
+    fn to_fields(self) -> (Vec<F>, usize) {
+        // NOTE: current version returns always the same amount of field elements in the returned
```
ed255, line 304: Actually I think always returning the same amount of field elements is much safer, even in the future, because if we use a dynamic length encoding then we may encounter more instances of collisions (different lists of StatementArg that are translated to the same vector of fields)

src/middleware.rs
```diff
@@ -367,3 +432,9 @@ pub struct MainPodInputs<'a> {
 pub trait PodProver {
     fn prove(&mut self, params: &Params, inputs: MainPodInputs) -> Result<Box<dyn MainPod>>;
 }
+
+pub trait ToFields {
+    /// returns Vec<F> representation of the type, and a usize indicating how many field elements
+    /// does the vector contain
+    fn to_fields(self) -> (Vec<F>, usize);
```
ed255, line 439: @arnaucube what's the reason to include the length as a value when the Vec already contains the same value with `Vec::len`?

PR #43, opened 2025-02-07
feat: Basic verification logic for mock MainPOD

ax0:
This PR fills in some of the gaps in the verification of mock MainPODs.


src/backends/plonky2.rs
```diff
@@ -0,0 +1 @@
+
```
ed255, line 1: :eyes:

src/backends/mock_main.rs
```diff
@@ -277,26 +300,92 @@ pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middlewar
...
-        // - define input_statements as `statements.[self.offset_input_statements()..]`
-        // - Calculate the id from a subset of the statements.  Check it's equal to self.id
-        // - Find a ValueOf statement from the public statements with key=KEY_TYPE and check that
-        // the value is PodType::MockMainPod
-        // - Check that all `input_statements` of type `ValueOf` with origin=SELF have unique keys
+        let input_statement_offset = self.offset_input_statements();
+        // get the input_statements from the self.statements
+        let input_statements = &self.statements[input_statement_offset..];
+        // get the id out of the public statements, and ensure it is equal to self.id
+        let ids_match = self.id == PodId(hash_statements(&self.public_statements).unwrap());
```
ed255, line 307: Note for myself: the MockMainPod has the field `public_statements` which is built from `statements[statements.len() - params.max_public_statements..]`, so we could get rid of it.

src/backends/mock_main.rs
```diff
@@ -277,26 +300,92 @@ pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middlewar
...
+        let has_type_statement = self
+            .public_statements
+            .iter()
+            .find(|s| {
+                s.0 == NativeStatement::ValueOf
+                    && s.1.len() > 0
+                    && if let StatementArg::Key(AnchoredKey(pod_id, key_hash)) = s.1[0] {
+                        pod_id == SELF && key_hash == hash_str(KEY_TYPE)
+                    } else {
+                        false
```
ed255, line 319: This would be a malformed invalid POD right?

src/backends/mock_main.rs
```diff
@@ -277,26 +300,92 @@ pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middlewar
...
+                        },
+                        s,
+                    )
+                })
+                .filter(|(i, s)| s.0 == NativeStatement::ValueOf)
+                .flat_map(|(i, s)| {
+                    if let StatementArg::Key(ak) = &s.1[0] {
+                        vec![(i, ak.1, ak.0)]
+                    } else {
+                        vec![]
```
ed255, line 345: This would be a malformed POD right?

src/backends/mock_main.rs
```diff
@@ -277,26 +300,92 @@ pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middlewar
...
         // (no duplicates)
-        // - Verify that all `input_statements` are correctly generated
+        let value_ofs_unique = {
+            let key_id_pairs = input_statements
+                .into_iter()
+                .enumerate()
+                .map(|(i, s)| {
+                    (
+                        // Separate private from public statements.
+                        if i < self.params.max_priv_statements() {
```
ed255, line 332: Does this mean that I can have a ValueOf for the same key with different values as long as one is in priv_statements and the other is in pub_statements?

src/backends/mock_main.rs
```diff
@@ -277,26 +300,92 @@ pub fn hash_statements(statements: &[middleware::Statement]) -> Result<middlewar
...
+        let statement_check = input_statements
+            .iter()
+            .enumerate()
+            .map(|(i, s)| {
+                self.operations[i]
+                    .deref(&self.statements[..input_statement_offset + i])
+                    .check(s.clone())
+            })
+            .collect::<Result<Vec<_>>>()
+            .unwrap();
```
ed255, line 362: This would panic if any `check` returns `Err` right?

src/middleware.rs
```diff
@@ -384,9 +428,128 @@ pub enum OperationArg {
...
+impl Operation {
+    pub fn code(&self) -> NativeOperation {
+        self.0
+    }
+    pub fn args(&self) -> &[OperationArg] {
+        &self.1
+    }
+    // TODO: Argument checking.
+    /// Checks the given operation against a statement.
+    pub fn check(&self, output_statement: Statement) -> Result<bool> {
```
ed255, line 461: About the return type:
Is the `Err` case reserved for malformed operation/statements, and the `Ok(false)` for properly formed data but invalid deduction?

src/middleware.rs
```diff
@@ -384,9 +428,128 @@ pub enum OperationArg {
...
+            // Nothing to check.
+            None => Ok(output_statement.code() == NativeStatement::None),
+            // Check that the resulting statement is of type `ValueOf`
+            // and its origin is `SELF`.
+            NewEntry =>
+                Ok(output_statement.code() == NativeStatement::ValueOf && output_statement.args()[0].key()?.origin() == SELF)
+            ,
+            // Check that the operation acts on a statement *and* the
+            // output is equal to this statement.
+            CopyStatement => Ok(output_statement == self.args()[0].statement()?)
```
ed255, line 473: This is a design question: In the Devcon implementation Operation and Statement were more strongly typed:
https://github.com/0xPARC/parcnet/blob/f7ef5e2fedea10f349c6343fb24d84374edbbad2/pod2/src/pod/operation.rs#L17-L33
https://github.com/0xPARC/parcnet/blob/f7ef5e2fedea10f349c6343fb24d84374edbbad2/pod2/src/pod/statement.rs#L79-L90

Here they are very generic, where the arguments are not strongly typed to be correct.
This generic code leads to a more verbose `check` implementation, but I also think it has benefits in other parts of the code.

Do you have any thoughts on this approach VS the previous one?  Do you like this implementation of `check`?

ed255 (approved):
LGTM!

arnaucube (approved):

PR #44, opened 2025-02-07
Migrate fmt to use Display trait

arnaucube:
Migrate fmt to use Display trait to simplify the usage of it.

Also with this, in the same snipped of code we can print types from backend, middleware and frontend; which before needed to import the two different `Printer` structs (frontend::Printer and mock_main::Printer).

Before:
```
let printer = mock_main::Printer { skip_none: false };
let mut w = io::stdout();
printer.fmt_mock_main_pod(&mut w, &pod).unwrap();

let printer = frontend::Printer { skip_none: false };
printer.fmt_main_pod_builder(&mut w, &pod_builder).unwrap();
```

now:
```
println!("{:#}", pod);
println!("{:#}", pod_builder);
```

And the equivalent to the old `skip_none: true`, now is done by just using `println!("{}", pod);


src/backends/mock_main.rs
```diff
@@ -46,6 +65,75 @@ pub struct MockMainPod {
...
+            if i == offset_input_statements {
+                writeln!(f, "  private statements:")?;
+            }
+            if i == offset_public_statements {
+                writeln!(f, "  public statements:")?;
+            }
+
+            let op = (i >= offset_input_statements)
+                .then(|| &self.operations[i - offset_input_statements]);
+            if !(!f.alternate() && st.is_none()) {
```
ed255, line 103: Nitpick: The check `!(!f.alternate() && st.is_none())` is also done in `fmt_statement_index`

ed255 (approved):
Very nice!

src/backends/mock_main.rs
```diff
@@ -32,6 +33,24 @@ impl OperationArg {
 #[derive(Clone, Debug, PartialEq, Eq)]
 struct Operation(pub NativeOperation, pub Vec<OperationArg>);
 
+impl fmt::Display for Operation {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?} ", self.0)?;
+        for (i, arg) in self.1.iter().enumerate() {
+            if !(!f.alternate() && arg.is_none()) {
```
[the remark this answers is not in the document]
arnaucube: I learned about it also for this PR ^^

src/backends/mock_main.rs
arnaucube: right, it was inherited from the [previous version](https://github.com/0xPARC/pod2/blob/f23a78dcb824b21a978b65e738aaca27f8ac9122/src/backends/mock_main.rs#L404), updated it now [at this amend](https://github.com/0xPARC/pod2/compare/40e7051c9f5cf61cbf4e3d2f264c08d4b425f666..36aa5f0d9f054a4f84882b77a23637dffa7cb6f9)

PR #49, opened 2025-02-10
Aard custom

tideofwords:
initial PR for custom deduction docs


book/src/custom1b.md
```diff
@@ -0,0 +1,41 @@
...
+- Directly, as either the AND or OR of two pre-existing predicates, or
+- Recursively, in a _batch_ of up to 10 custom predicates.  Each custom predicate in the batch is either the AND or OR of five predicates which are either pre-existing, or defined in the same batch.
+
+[Note: We could potentially allow the AND or OR of, say, two predicates instead of five.  To make this work, we might need to have access to pod ID as a virtual key, see github issue #60]
+
+## Arguments of custom predicates
+
+The definition of a custom predicate might also be called an _operation_ or _deduction rule_.  It includes two (or, potentially, say, five) input statements.  The arguments to the input statements are decomposed as (origin, key) pairs: if statements are allowed to have arity at most 4, then the input statements in a deduction rule will have at most 8 arguments (4 origins and 4 keys).  The same holds for the output statement.
+
+Each argument (origin or key) to an input statement is either a wildcard or a literal.  In the backend, the wildcard arguments will be identified as *1, *2, *3, ....
```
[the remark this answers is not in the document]
ed255: But I think I like the term "wildcard" better than "tag".

book/src/custom2.md
```diff
@@ -0,0 +1,130 @@
...
+ValueFromPodKey(local_key: KeyOrLiteral::String, origin_id: KeyOrLiteral::OriginID, key: KeyOrLiteral::String).
+```
+
+```
+Reference(reference_key, key)
+```
+means that the *value* of `reference key` is the *key name* of `key`.
+
+```
+ValueFromPodKey
```
ed255, line 89:
```suggestion
ValueFromPodKey(local_key, origin_id, key)
```

book/src/custom2.md
```diff
@@ -0,0 +1,130 @@
...
+To make a statement about an inherited value (a value introduced in an ancestor POD), the value must be copied to a local value:
+
+The statements below assert that "name" on pod1 and "friend" on pod2 are assigned the same value.
+```
+ValueFromPodKey(name_from_pod1, pod1, "name")
+ValueFromPodKey(friend_from_pod2, pod2, "friend")
+Equal(name_from_pod1, friend_from_pod2)
+```
+
+## How to inherit local variables from a previous POD
```
ed255, line 104: I would appreciate an example use-case to be added here https://github.com/0xPARC/pod2/blob/main/book/src/examples.md
which requires propagating keys from PODs, because I believe there are different ways to approach this propagation / inheritance of keys but I find it hard to reason if a particular way is good or not because I don't have a use-case to validate it.

book/src/notes
```diff
@@ -0,0 +1,6 @@
+Ethdos(x, y, d) = OR ( EthdosBase(x, y, d), EthdosInd(x, y, d) )
```
ed255, line 1: Maybe this file was committed by mistake?

book/src/customexample.md
```diff
@@ -0,0 +1,79 @@
...
+    
+    // the attestation pod is signed by (src_or, src_key)
+    Equal((attestation_pod, "signer"), (src_or, src_key))  
+
+    // that same attestation pod has an "attestation"
+    Equal((attestation_pod, "attestation"), (dst_or, dst_key))
+>
+
+On the backend, this is converted to a rule where all the arguments are either wildcards or literals:
+eth_friend(*1, *2, *3, *4) = and<
```
ed255, line 18: This is a very interesting example to discuss the topic of arities in custom predicates.
This custom predicate has:
- The predicate has 4 arguments (that's what I was calling public  in https://github.com/0xPARC/pod2/issues/54)
- The predicate has 5 wildcards
- The predicate has 1 wildcard that doesn't come from a predicate argument (that's what I was calling private in https://github.com/0xPARC/pod2/issues/54)

book/src/custom1.md
```diff
@@ -0,0 +1,52 @@
...
+A custom operation [^operation] is a rule that allows one to deduce a custom statement from one or more existing statements according to a logical rule, described below.
+
+> Note: Unlike built-in operations, it is not possible to perform arbitrary calculations inside a custom operation.
+
+The syntax of a custom operation is best explained with an example.
+
+Original example with anchored keys, origins, and keys.
+| Args | Condition            | Output                      |
+|------------|-----------------------------------------|----|
+| pod: Origin, <br> good_boy_issuers: AnchoredKey::MerkleRoot, <br> receiver: AnchoredKey | ValueOf(AK(pod, "_type"), SIGNATURE), <br> Contains(good_boy_issuers, AK(pod,"_signer")), <br> Equals(AK(pod, "friend"), receiver) | GoodBoy(receiver, good_boy_issuers) |
```
ed255, line 14: I would like to discuss how to achieve the following functionality:
Imagine you want a Condition that is something like `Lt(18, (pod, "age"))`.
How would you implement that?
In the current spec we don't support constants in the `Lt` predicate.  So we would translate it to
```
ValueOf(AK(SELF, "eighteen"), 18)
Lt(18, (pod, "age"))
```

But that doesn't work very well here.
Unless you use the "CUSTOM 2 DO NOT USE" way, which would allow us to do
```
ValueOf(local_varname_eighteen, 18)
Lt(local_varname_eigtheen, 18)
```

And every time we create a custom statement from it we use a different variable name for `local_varname_eighteen`.

book/src/custom.md
```diff
@@ -1 +1,18 @@
 # Custom statements and custom operations
+
+Users of the POD system can introduce _custom predicates_ (previously called _custom statements_) to express complex logical relations not available in the built-in predicates.  Every custom predicate is defined as the conjunction (AND) or disjunction (OR) of a small number of other statements.
+
+When a custom predicate is introduced in a MainPod, it becomes available for use in that POD and all PODs that inherit[^inherit] from it.
+
+A custom predicate can be defined either _nonrecursively_ or _recursively_ (as part of a "batch"). A nonrecursive custom predicate is defined in terms of previously defined predicates (whether custom or native).  A "batch" of custom predicates can be defined _recursively_: the definition of any custom predicate in the batch can use both previously defined predicates and all the predicates in the batch.
```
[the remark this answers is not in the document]
tideofwords: Changing to "group".

In general, will the frontend allow combinations of conjunction and disjunction in a single custom predicate?  As in
pred1:
  AND(OR(pred2, pred3), pred4)?
This looks like a good idea.

In this case, of course we would also allow a recursive definition, like
pred1:
  AND(OR(pred2, pred3), pred1).

If we require a single top-level predicate for each group, the user won't be able to make multiple predicates depend on each other recursively.  I can't think of a natural use case where you would want that, but if the backend allows it, why disable it on the frontend?

book/src/custom.md
```diff
@@ -1 +1,18 @@
...
+
+## Custom predicates and their IDs
+
+A custom predicate, like a built-in predicate, is identified by a _name_ on the front end and an _identifier_ on the back end.  In the non-recursive case, the back-end identifier is defined as a hash of the definition of the custom predicate.
```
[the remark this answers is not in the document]
tideofwords: Sounds good to me

book/src/custom1.md
```diff
@@ -0,0 +1,52 @@
...
+
+| Args | Condition            | Output                      |
+|------------|-----------------------------------------|----|
+| *1 (pod), <br> *2 (good_boy_issuers_origin), <br> *3 (good_boy_issuers_key), <br> *4 (receiver_origin), <br> *5 (receiver_key) | ValueOf(AK(*1, "_type"), SIGNATURE), <br> Contains(AK(*2, *3), AK(*1,"_signer")), <br> Equals(AK(*1, "friend"), AK(*4, *5)) | GoodBoy(AK(*4, *5), AK(*2, *3)) |
+
+If you want to apply this deduction rule to prove a `GoodBoy` statement,
+you have to provide the following witnesses in-circuit.
+
+- Copy of the deduction rule
+- Values for *1, *2, *3, *4, *5.  (*1, *2, *4 should be OriginIDs for the three origins; *3 and *5 should be key names.)
```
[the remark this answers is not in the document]
tideofwords: That's right, you don't have to enforce type checking in-circuit (I think you can't anyway?), because the conditions can only be satisfied if the types are right.

book/src/custom1b.md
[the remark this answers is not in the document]
tideofwords: "Statement templates" is good

book/src/custom1.md
tideofwords: >> But that doesn't work very well here.

Why not?
In the condition, you'd simply have
ValueOf(*3, *4, 18)
Lt(*3, *4, *1, *2).

Outside the condition, you'd first introduce a new key SELF.local_eighteen.  Then to apply the custom predicate, use *3 = SELF, *4 = local_eighteen.

book/src/custom2.md
tideofwords: Cool, I'm adding "attested great boy pod", which reveals the names of the friends who have good boy pods.

book/src/notes
tideofwords: Yep, removed it

arnaucube:
Since @ed255 is off for the next few weeks, is it ok @tideofwords if we merge this PR as a consolidation of the current spec that we're already implementing?
And when @ed255 comes back can check your answers and we can iterate if needed.

book/src/customexample.md
```diff
@@ -0,0 +1,80 @@
+
+# Ethdos custom predicate, using binary AND and OR: example of a direct definition and a recursive group
+
+## Direct definition
+(DEPRECATED: All predicates must be defined as part of a group.)
```
arnaucube, line 5: does this mean that this part of the doc needs to be updated?

book/src/custom.md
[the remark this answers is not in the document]
ed255: > Changing to "group".
> 
> In general, will the frontend allow combinations of conjunction and disjunction in a single custom predicate? As in pred1: AND(OR(pred2, pred3), pred4)? This looks like a good idea.
> 
> In this case, of course we would also allow a recursive definition, like pred1: AND(OR(pred2, pred3), pred1).

I also think this is a good idea for the frontend.  Nevertheless I also think it's OK to start with a simpler version where we don't implement it yet and leave it as a TODO for later. 

> If we require a single top-level predicate for each group, the user won't be able to make multiple predicates depend on each other recursively. I can't think of a natural use case where you would want that, but if the backend allows it, why disable it on the frontend?

Why having a single top-level predicate in a group doesn't allow the user to have multiple predicates depend on each other recursively?  Is it because those internal predicates don't have names that the user can refer to?  If that's the case, I think the user could still give names to internal predicates during the build process to reference them, and still end up with a single top-level predicate in the group.  Or is there any other reason?

I thought about assuming a single top-level predicate for each group because then we could say the top-level (entry-point) predicate always takes the same index in the group (let's say: 0).  This would be a convention that could facilitate debugging.

book/src/custom1.md
ed255: yes, you're right, I got confused for a second.
The solution is passing a string via wildcard that will be used as a key (which could be part of an AnchoredKey in CUSTOM 1, or a local key in CUSTOM 2)

book/src/customexample.md
```diff
@@ -0,0 +1,80 @@
...
+```
+eth_friend(src_or, src_key, dst_or, dst_key) = and<
+    // there is an attestation pod that's a SIGNATURE POD
+    ValueOf(attestation_pod, "type", SIGNATURE)
```
arnaucube, line 9: question, is the `attestation_pod` a 'private' argument? if so, should it need to be defined in some way at the predicate definition, something like
`eth_friend(src_or, src_key, dst_or, dst_key, priv(attestation_pod) = and<...`?
(probably not with this syntax, but specifying somehow that the `attestation_pod` is an input to the template)

book/src/customexample.md
tideofwords: Yes.  Fixed.

book/src/customexample.md
tideofwords: attestation_pod is an argument to the operation, but not an argument to the predicate.

In other words:
When you do the operation in-circuit, you need to provide attestation_pod as a witness.
But in the list of statements, when you have an eth_friend statement, you don't show the attestation_pod there.

arnaucube (approved):

PR #55, opened 2025-02-11
Implement Containers (Dictionary,Set,Array) on top of MerkleTree. And restructure the code.

arnaucube:
- Reorganize the code grouping backends, middleware, frontend, (crypto) primitives.
- Add containers (Dictionary,Set,Array) at the middleware layer, so that it can be used both by the backend and frontend. The Dictionary, Set, Array use the merkletree differently under the hood as specified at https://github.com/0xPARC/pod2/blob/f2575d1524c9e15a5b688eb444ee3c39df242c81/book/src/values.md#dictionary-array-set
	- The containers introduce the trait Container, which has the method 'cm()'. At the current version this uses a merkletree under the hood, and the method 'cm' returns the merkle root.
- Ideally neither frontend nor backend use the MerkleTree type, and they use the wrappers {Dictionary,Set,Array}. Note that at the current implementation the MerkleTree is used at the mock-backend to check internal values, but not at the struct types.
- updated the spec's merkletree section updating the defined interface
- traitify the MerkleTree to make it extendable by Containers, so that the tree methods are exposed at Dictionary,Set,Array types

update: [added](https://github.com/0xPARC/pod2/pull/55/commits/7338cd263a868ccf0560fa45cec8710b34fd11b1) also a github ci to run the tests at each new commit at `main` branch and new PR to `main` branch. Reasoning: initially we were skipping running the tests so that we could move fast, but now the code has more complexity, so better that we ensure that we don't break anything with new commits/PRs.


book/src/merkletree.md
```diff
@@ -149,12 +149,15 @@ impl MerkleTree {
...
     /// returns the root of the tree
-    fn root(&self) -> Result<Hash>;
+    fn root(&self) -> Hash;
+    
+    /// returns the value at the given key
+    pub fn get(&self, key: &Value) -> Value;
     
     /// returns a proof of existence, which proves that the given key exists in
     /// the tree. It returns the `value` of the leaf at the given `key`, and
     /// the `MerkleProof`.
```
ax0, line 159:
```suggestion
    /// the tree. It returns the `MerkleProof`.
```

src/middleware/containers.rs
```diff
@@ -0,0 +1,120 @@
+/// This file implements the types defined at
+/// https://0xparc.github.io/pod2/values.html#dictionary-array-set .
+use plonky2::hash::poseidon::PoseidonHash;
+use plonky2::plonk::config::Hasher;
+use std::collections::HashMap;
+
+use super::{Hash, Value, EMPTY};
+use crate::primitives::merkletree::MerkleTree;
+
+/// Contaiter is a wrapper of a MerkleTree, used to achieve Dictionary, Set, Array frontend types.
```
ax0, line 10:
```suggestion
/// Container is a wrapper of a MerkleTree, used to achieve Dictionary, Set, Array frontend types.
```

ax0 (approved):
LGTM. I wonder whether it would make sense to add additional methods to the container types, e.g. `get`, rather than referring to the underlying Merkle tree explicitly.

book/src/merkletree.md
arnaucube: right, thanks!

arnaucube:
> LGTM. I wonder whether it would make sense to add additional methods to the container types, e.g. `get`, rather than referring to the underlying Merkle tree explicitly.

Yes! agree, I've updated it at the commit https://github.com/0xPARC/pod2/pull/55/commits/44db29d04947996dfda371f1c2e527a9512e7728 , now the `MerkleTree` is defined by a type, and the `Container` is a trait that extends it, so that `Dictionary`,`Set`,`Array` expose also the methods of the MerkleTree.

More concretely, the `Container` trait is a thin wrapper on top of the `MerkleTreeTrait` which adds the `new` method, which allows to define the type used to create a new tree/container. For example, while the `MerkleTree` is internally built from `HashMap<Value,Value>`, the types `Dictionary`,`Set`,`Array` are built from different types and approaches:
- `Dictionary` is built from `HashMap<Hash,Value>`
- `Set` is built from `Vec<Value>` (where the tree leafs are `key=hash(Value)`, `value=0`)
- `Array` is built from `Vec<Value>` (where the tree leafs are `key=i`, `value=Value`)

src/middleware/containers.rs
```diff
@@ -0,0 +1,191 @@
...
+
+use super::{Hash, Value, EMPTY};
+use crate::primitives::merkletree::{MerkleProof, MerkleTree, MerkleTreeTrait};
+
+/// Container is a wrapper of a MerkleTree, used to achieve Dictionary, Set, Array frontend types.
+/// It offers all the methods of the trait `MerkleTreeTrait`, with an additional constructor `new`
+/// that allows each specific type (ie. Dictionary, Set, Array) to define how each type is
+/// constructed (for example a Dictionary is built from HashMap<Hash,Value>, whereas a set is built
+/// from Vec<Value>).
+pub trait Container: MerkleTreeTrait {
```
ed255, line 16: I don't understand the goal of this trait.

Following this pattern you have:
- Each of Dictionary, Set and Array have a method `root`.  But to me that's leaking the underlying abstraction.  Shouldn't it be `commit` instead?
- Each of Dictionary, Set and Array have `get` method that uses `key: &Value`.  But in an Array you want `get(index: usize)`, and in a Set, the method `get` doesn't make much sense.
- In Array you have `prove_nonexistence` but again that doesn't make much sense.

Why not create 3 generic types called `Dictionary, Set, Array` that don't implement any trait, and instead offer a Dictionary, Set and Array API?  And internally each one is a wrapper to a generic MerkleTree?

Dicionary would have:
- new(raw: &HashMap<Hash, Value>) -> Self
- commit(&self) -> Hash
- get(&self, key: &Value) -> Result<Value>
- prove(&self, key: &Value) -> Result<MerkleProof>
- prove_nonexistence(&self, key: &Value) -> Result<MerkleProof>
- verify(root: Hash, proof: &MerkleProof, key: &Value, value: &Value)
- verify_nonexistence(root: Hash, proof: &MerkleProof, key: &Value)
- iter(&self) -> std::collections::hash_map::Iter<Value, Value>

Set would have:
- new(raw: Vec<Value>) -> Self
- commit(&self) -> Hash
- contains(&self, key: &Value) -> bool
- prove(&self, value: &Value) -> Result<MerkleProof>
- prove_nonexistence(&self, value: &Value) -> Result<MerkleProof>
- verify(root: Hash, proof: &MerkleProof, value: &Value)
- verify_nonexistence(root: Hash, proof: &MerkleProof, value: &Value)
- iter(&self) -> std::collections::vec::Iter<Value>

Array would have:
- new(raw: Vec<Value>) -> Self
- commit(&self) -> Hash
- get(&self, index: usize) -> Result<Value>
- prove(&self, index: usize) -> Result<MerkleProof>
- verify(root: Hash, proof: &MerkleProof, index: usize, value: &Value)
- iter(&self) -> std::collections::vec::Iter<Value>

ed255 (commented):
I have some questions / proposal about the container types APIs.

src/middleware/containers.rs
arnaucube: (for context, regarding the method `commit/commitment` vs `root`, initially [at the first commit](https://github.com/0xPARC/pod2/pull/55/commits/0782da0a9009c8cd94ae75e37ed53bcd90d8bf41#diff-673845316285abdee93534a618bdc1c189f3c52059dce23ed39f68b98f004bfaR17) the `Container` trait had the method `.cm()` (for commitment), but since it was just a repetition of the already existing method `.root()` of the tree, I've ended up removing it at the [3rd commit](https://github.com/0xPARC/pod2/pull/55/commits/b3f8aa4978b2bfd7ae66b4a474cc10a992c94f05#diff-673845316285abdee93534a618bdc1c189f3c52059dce23ed39f68b98f004bfaL17). But this is not a problem once getting rid of the traits as you suggest, commented the traits in the next paragraph.)

Regarding the traits and structs, I was trying to reuse as much code as possible (having dictionary,set,array being variations of the same wrapper), but I agree that it makes sense to have 3 different struct implementations for the different types (dictionary,set,array) without traits.

Updated at https://github.com/0xPARC/pod2/pull/55/commits/ecede2f0ba2b45cd14404a5485bba2d4d818acd5 removing the traits and adding dedicated impl for the Dictionary,Set,Array struct types.

src/middleware/containers.rs
```diff
@@ -0,0 +1,159 @@
...
+    type IntoIter = std::collections::hash_map::Iter<'a, Value, Value>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.mt.iter()
+    }
+}
+
+impl PartialEq for Dictionary {
+    fn eq(&self, other: &Self) -> bool {
+        self.mt.root() == other.mt.root() && self.mt.root() == other.mt.root()
```
ed255, line 60: is this the same comparison twice?

src/middleware/containers.rs
```diff
@@ -0,0 +1,159 @@
...
+            })
+            .collect();
+        Self {
+            mt: MerkleTree::new(&kvs),
+        }
+    }
+    pub fn commitment(&self) -> Hash {
+        self.mt.root()
+    }
+    pub fn contains(&self, key: &Value) -> bool {
```
ed255, line 89:
```suggestion
    pub fn contains(&self, value: &Value) -> bool {
```

A set just has values, no keys.

src/middleware/containers.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.mt.contains(key)
+    }
+    pub fn prove(&self, key: &Value) -> Result<MerkleProof> {
```
ed255, line 92:
```suggestion
    pub fn prove(&self, value: &Value) -> Result<MerkleProof> {
```
Prove that a value is in the set

src/middleware/containers.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.mt.prove(key)
+    }
+    pub fn prove_nonexistence(&self, key: &Value) -> Result<MerkleProof> {
```
ed255, line 95:
```suggestion
    pub fn prove_nonexistence(&self, value: &Value) -> Result<MerkleProof> {
```

src/middleware/containers.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.mt.prove_nonexistence(key)
+    }
+    pub fn verify(root: Hash, proof: &MerkleProof, key: &Value, value: &Value) -> Result<()> {
+        MerkleTree::verify(root, proof, key, value)
+    }
```
ed255, line 100:
```suggestion
    pub fn verify(root: Hash, proof: &MerkleProof, value: &Value) -> Result<()> {
        MerkleTree::verify(root, proof, value, EMPTY)
    }
```

src/middleware/containers.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn verify_nonexistence(root: Hash, proof: &MerkleProof, key: &Value) -> Result<()> {
```
ed255, line 101:
```suggestion
    pub fn verify_nonexistence(root: Hash, proof: &MerkleProof, value: &Value) -> Result<()> {
```

src/middleware/containers.rs
arnaucube: righ! my bad, thanks!, applied together with the other suggestions at https://github.com/0xPARC/pod2/pull/55/commits/cd7d968d150e44a92f46b92675611b0dd237c3ca

ed255 (approved):
LGTM!

src/middleware/containers.rs
arnaucube: right, updated at https://github.com/0xPARC/pod2/pull/55/commits/0bcac22492cf274beed6014f55868f369259c4fb

PR #62, opened 2025-02-14
small fixes for various todos

rrrliu:


src/backends/mock_main.rs
```diff
@@ -278,7 +287,7 @@ impl MockMainPod {
             Self::pad_operation_args(params, &mut mid_args);
             let mut args = Vec::with_capacity(mid_args.len());
             for mid_arg in &mid_args {
-                args.push(Self::find_op_arg(statements, mid_arg));
+                args.push(Self::find_op_arg(statements, mid_arg).expect("Failed to find operation argument"));
```
arnaucube, line 290: since you added at the `find_op_arg` method to return an error (line 248), maybe this current line could return that error to the upper level instead of panicking by the `.expect`?

(same for the `expect` at line 312)

src/frontend.rs
```diff
@@ -599,7 +601,12 @@ pub mod tests {
         let kyc = zu_kyc_pod_builder(&params, &gov_id, &pay_stub);
         println!("{}", kyc);
 
-        // TODO: prove kyc with MockProver and print it
+        let mut prover = MockProver {};
```
ed255, line 604: This TODO was meant to mean to print the type `MainPod` which internally has been created with the MockProver, but that shouldn't matter in the frontend.
Something like this;
```rust
let mut prover = MockProver {};
let kyc_pod = kyc.prove(&mut prover).unwrap();
println!("{}", kyc_pod);
```

But the `Display` trait is not implemented for `frontend::MainPod` so that's what's actually missing to complete this.

rrrliu: makes sense -- i've just reverted the file now since we don't have a clear picture of what the Display should show!

arnaucube (approved):
LGTM!

PR #63, opened 2025-02-15
feat: add new bool frontend type

rrrliu:
- after brief discussion with aard, we decided adding at least some of the POD1 types to POD2 would make sense
- starting with a `bool` atomic type to POD2 since it is a valid JSON type
- added a brief test with a 'ticket' signed pod and a main pod that proves some things about the ticket

### DIFF
--- a/src/examples.rs
+++ b/src/examples.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use crate::backends::mock_signed::MockSigner;
 use crate::frontend::{MainPodBuilder, SignedPod, SignedPodBuilder, Value};
 use crate::middleware::{containers::Dictionary, Params, PodType, KEY_SIGNER, KEY_TYPE};
 use crate::op;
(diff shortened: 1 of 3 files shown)


src/examples.rs
```diff
@@ -194,3 +193,35 @@ pub fn great_boy_pod_full_flow() -> MainPodBuilder {
...
+
+pub fn tickets_sign_pod_builder(params: &Params) -> SignedPodBuilder {
+    // Create a signed pod with all atomic types (string, int, bool)
+    let mut builder = SignedPodBuilder::new(params);
+    builder.insert("eventId", 123);
+    builder.insert("productId", 456);
+    builder.insert("attendeeName", "John Doe");
+    builder.insert("attendeeEmail", "john.doe@example.com");
+    builder.insert("isConsumed", true);
+    builder.insert("isRevoked", false);
```
arnaucube, line 207: question from my ignorance on the POD system:
how is the POD revoked? since the `isRevoked` field is part of the key-values of the SignedPOD, in that case, the POD.id will change so it's no longer the original POD but a new one, right?
(same applies for the field `isConsumed`)

rrrliu: good question -- this is a bit of a Zupass-specific example, but we have a protocol called 'feeds' that continually polls PODs from a server every minute, and the server re-issues certain PODs to a given user (there is an external stable "PCD ID" to map original PODs -> new PODs)

ed255 (approved):
LGTM!

arnaucube (approved):

Issue #65, opened 2025-02-17
Explore refactoring Statement and Operation to be more strongly typed


ed255:
Currently Statement and Operation are defined with a Vector of arguments that can take different types, but each operation and statement as specified takes a fixed number of arguments of fixed type each.

The devcon implementation had this typed like this:
https://github.com/0xPARC/parcnet/blob/f7ef5e2fedea10f349c6343fb24d84374edbbad2/pod2/src/pod/statement.rs#L79-L90

Explore if we can do the same in this implementation, what are the tradeoffs of each implementation, etc.

PR #69, opened 2025-02-18
chore: enums for statement and op types

ax0:
Towards #65.

### DIFF
--- a/src/backends/mock_main.rs
+++ b/src/backends/mock_main.rs
@@ -2,13 +2,15 @@ use crate::middleware::{
     self, hash_str, AnchoredKey, Hash, MainPodInputs, NativeOperation, NativeStatement, NonePod,
     Params, Pod, PodId, PodProver, Statement, StatementArg, ToFields, KEY_TYPE, SELF,
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use itertools::Itertools;
 use plonky2::hash::poseidon::PoseidonHash;
 use plonky2::plonk::config::Hasher;
 use std::any::Any;
 use std::error::Error;
 use std::fmt;
+
+pub const VALUE_TYPE: &str = "MockMainPOD";
 
 pub struct MockProver {}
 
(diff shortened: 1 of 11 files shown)


src/middleware.rs
```diff
@@ -447,133 +480,172 @@ pub enum NativeOperation {
...
-        &self.1
+    /// Forms operation from op-code and arguments.
+    pub fn op(op_code: NativeOperation, args: &[Statement]) -> Result<Self> {
+        type NO = NativeOperation;
+        let arg_tup = (
+            args.get(0).cloned(),
+            args.get(1).cloned(),
+            args.get(2).cloned(),
+        );
+        Ok(match (op_code, arg_tup) {
```
ed255, line 580: Maybe we also need to include `args.len()` in the match to validate the arguments length of each operation

src/backends/mock_main.rs
```diff
@@ -35,18 +37,103 @@ impl OperationArg {
...
+    type Error = anyhow::Error;
+    fn try_into(self) -> Result<middleware::Statement> {
+        type S = middleware::Statement;
+        type NS = NativeStatement;
+        type SA = StatementArg;
+        let args = (
+            self.1.get(0).cloned(),
+            self.1.get(1).cloned(),
+            self.1.get(2).cloned(),
+        );
```
ed255, line 93: Maybe we also need to include `args.len()` in the match to validate the arguments length of each statement

src/middleware.rs
```diff
@@ -447,133 +480,172 @@ pub enum NativeOperation {
...
-                let s2_value:i64 = s2.args()[1].literal()?.try_into()?;
-                let s3 = self.args()[2].statement()?;
-                let s3_key = s3.args()[0].key()?;
-                let s3_value: i64 = s3.args()[1].literal()?.try_into()?;
-                let sum_holds = s1.code() == NativeStatement::ValueOf && s2.code() == NativeStatement::ValueOf && s3.code() == NativeStatement::ValueOf && s1_value == s2_value + s3_value;
-                Ok(sum_holds && output_statement.code() == NativeStatement::SumOf && output_statement.args()[0].key()? == s1_key && output_statement.args()[1].key()? == s2_key && output_statement.args()[2].key()? == s3_key)
-            },
-            // TODO: Remaining ops.
-            _ => Ok(true)
+    pub fn check(&self, output_statement: &Statement) -> Result<bool> {
```
ed255, line 608: This function looks much cleaner with this refactor.

src/frontend.rs
```diff
@@ -550,17 +739,13 @@ pub mod tests {
...
-    }
-
     macro_rules! op {
-        (eq, $($arg:expr),+) => { Operation(NativeOperation::EqualFromEntries, args!($($arg),*)) };
-        (ne, $($arg:expr),+) => { Operation(NativeOperation::NotEqualFromEntries, args!($($arg),*)) };
-        (gt, $($arg:expr),+) => { Operation(NativeOperation::GtFromEntries, args!($($arg),*)) };
-        (lt, $($arg:expr),+) => { Operation(NativeOperation::LtFromEntries, args!($($arg),*)) };
-        (contains, $($arg:expr),+) => { Operation(NativeOperation::ContainsFromEntries, args!($($arg),*)) };
-        (not_contains, $($arg:expr),+) => { Operation(NativeOperation::NotContainsFromEntries, args!($($arg),*)) };
+        (eq, $($arg:expr),+) => { Operation::EqualFromEntries($(OperationArg::try_from($arg).unwrap()),*) };
```
ed255, line 743: This is cool, now the compiler will complain when using the macro with the wrong number of arguments because it's building an enum.

src/frontend.rs
```diff
@@ -150,6 +155,12 @@ impl SignedPod {
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct AnchoredKey(pub Origin, pub String);
 
+impl Into<middleware::AnchoredKey> for AnchoredKey {
```
ed255, line 158: It's preferable to implement the `From` trait so that we get both `From` and `Into`.
Rust allows implementing a `From` trait on a foreign type without issues.

src/frontend.rs
```diff
@@ -165,13 +176,80 @@ impl fmt::Display for StatementArg {
...
+            Self::Contains(ak1, ak2) => vec![Key(ak1), Key(ak2)],
+            Self::NotContains(ak1, ak2) => vec![Key(ak1), Key(ak2)],
+            Self::SumOf(ak1, ak2, ak3) => vec![Key(ak1), Key(ak2), Key(ak3)],
+            Self::ProductOf(ak1, ak2, ak3) => vec![Key(ak1), Key(ak2), Key(ak3)],
+            Self::MaxOf(ak1, ak2, ak3) => vec![Key(ak1), Key(ak2), Key(ak3)],
+        }
+    }
+}
+
+impl Into<middleware::Statement> for Statement {
```
ed255, line 230: It's preferable to implement the `From` trait so that we get both `From` and `Into`.
Rust allows implementing a `From` trait on a foreign type without issues.

src/backends/mock_main.rs
```diff
@@ -35,18 +37,103 @@ impl OperationArg {
...
+                acc
+            });
+        (
+            [native_statement_f, vec_statementarg_f].concat(),
+            native_statement_f_len + vec_statementarg_f_len,
+        )
+    }
+}
+
+impl TryInto<middleware::Statement> for Statement {
```
ed255, line 83: It's preferable to implement the `TryFrom` trait so that we get both `TryFrom` and `TryInto`.
Rust allows implementing a `TryFrom` trait on a foreign type without issues.

src/middleware.rs
```diff
@@ -447,133 +480,172 @@ pub enum NativeOperation {
...
+            Self::RenameContainedBy(s1, s2) => vec![s1, s2],
+            Self::SumOf(s1, s2, s3) => vec![s1, s2, s3],
+            Self::ProductOf(s1, s2, s3) => vec![s1, s2, s3],
+            Self::MaxOf(s1, s2, s3) => vec![s1, s2, s3],
+        }
     }
-    pub fn args(&self) -> &[OperationArg] {
-        &self.1
+    /// Forms operation from op-code and arguments.
+    pub fn op(op_code: NativeOperation, args: &[Statement]) -> Result<Self> {
```
ed255, line 573: Is this method used anywhere?

ed255:
Here's my summary of the new types for Statement and Operation introduced in this PR:

```
enum frontend::Statement
- fn code(&self) -> middleware::NativeStatement
- fn args(&self) -> Vec<frontend::StatementArg>

frontend::Statement -> middleware::Statement

enum frontend::Operation
- fn code(&self) -> middleware::NativeOperation
- fn args(&self) -> Vec<frontend::StatementArg>

enum middleware::Statement
- fn code(&self) -> middleware::NativeStatement
- fn args(&self) -> Vec<middleware::StatementArg>

enum middleware::Operation
- fn code(&self) -> middleware::NativeOperation
- fn args(&self) -> Vec<middleware::Statement>

// Runtime type-checking 
(middleware::NativeOperation, &[middleware::Statement]) -> middleware::Operation

struct mock_main::Statement(pub middleware::NativeStatement, pub Vec<middleware::StatementArg>)

// Runtime type-checking
mock_main::Statement -> middleware::Statement

middleware::Statement -> mock_main::Statement
```

And now some thoughts:
- I think having Statement and Operation defined in the middleware in a strongly-typed is very useful because the type already documents itself on what number of arguments and of which type it uses.
- I think having runtime type-checkers in the middleware to go from `(code, args) -> enum_type` is very useful both to be used by frontend and backend.  Currently this runtime type-check is implemented for `Operation` in the middleware and for `Statement` in mock_main (I would move the later one to the middleware)
- I'm less convinced about having the Statement and Operation enum types in the frontend, I think they add significant boilerplate and it's not clear to me the advantage of this.

Overall I think there are two important points of discussion for this refactor:
- 1. Runtime VS Compile-time Statement/Operation argument type-checking
- 2. Ergonomics
- 3. Future refactors speed

For (1) I don't see any strong advantage on any direction.  Compile-time checking is nice, but I don't think it's super useful for us.  I think we can get a nice UX with runtime-checks and proper error reporting.  Moreover for custom predicates we can't have compile-time checks; and if we want to unify the API for native predicates and custom predicates we may end up having an API that can't do compile-time checks for native predicates.

For (2) I think having the enums in the middleware has benefits because it serves as spec and also allows us to keep the runtime checks in a single place with clean code.  On the other hand, I think having the enums in the frontend may worsen the ergonomics.

For (3) I think this refactor as is would make it more costly to change operations / statements and how their arguments work in the future.  For example if we decide to get rid of anchored keys, the current code would make it simpler than the code in this PR.  If we decide to allow constants in any statement argument the same reasoning applies.  This is another reason why I'm a bit hesitant to have the enum types in the frontend.

src/examples.rs
```diff
@@ -208,20 +208,36 @@ pub fn tickets_sign_pod_builder(params: &Params) -> SignedPodBuilder {
     builder
 }
 
-pub fn tickets_pod_builder(params: &Params, signed_pod: &SignedPod, expected_event_id: i64, expect_consumed: bool, blacklisted_emails: &Value) -> MainPodBuilder {
+pub fn tickets_pod_builder(
```
ed255, line 211: This is just a formatting change right?  Maybe we should include a rustfmt.toml in the repo to have a single formatting style. @arnaucube

ed255:
looking good!

ed255 (approved):
LGTM!

PR #74, opened 2025-02-18
Prototype custom predicates

### LINKED ISSUE #66: Prototype Custom Statements
Implement a prototype of how a custom statement would look like at the frontend or middleware.

ed255:
This is an initial prototype of custom predicates.
This code doesn't need to be merged straight forward, it is here to discuss the implementation (for this reason I marked it as Draft).

I added a lot of logic in a test module inside the middleware, but that logic doesn't belong there.  I did it that way for prototyping purposes, and because playing with the middleware types is simpler; but that logic should belong to the frontend and use a mix of middleware and frontend types.

~~A missing piece from this prototype is recursion.~~ I've found a clean way to add recursion which makes this prototype more complete.

In this implementation you can see how having AnchoredKeys complicates some parts, whereas having flat arguments would make things more straight forward at the cost of introducing extra statements.

I implemented the examples from #49  in the test, here's the output:

```
a.0. eth_friend = and<
  ValueOf(*4, ceb8ae2d…), 3),
  Equal(*4, 65a0f8c9…), (*0, *1)),
  Equal(*4, 53418ede…), (*2, *3)),
>(*0, *1, *2, *3)

b.0. eth_dos_distance_base = and<
  Equal(*0, 9784bc82…), (*2, *3)),
  ValueOf(*4, *5), 0),
>(*0, *1, *2, *3, *4, *5)

b.1. eth_dos_distance_ind = and<
  self.3(*0, *1), (*10, *11), (*8, *9)),
  ValueOf(*6, *7), 1),
  SumOf(*4, *5), (*8, *9), (*6, *7)),
  366abc4e….0(*10, *11), (*2, *3)),
>(*0, *1, *2, *3, *4, *5)

b.2. eth_dos_distance = or<
  self.0(*0, *1), (*2, *3), (*4, *5)),
  self.1(*0, *1), (*2, *3), (*4, *5)),
>(*0, *1, *2, *3, *4, *5)
```

Resolve #66


arnaucube (approved):

ax0 (approved):
LGTM

Issue #80, opened 2025-02-19
Frontend custom predicates


arnaucube:
- [x] how to declare custom predicate in the frontend
- [x] move & adapt `CustomPredicateBatchBuilder` to frontend (currently lives in the middleware tests, and needs to be simplified its usage for frontend)
- [x] how to operate with `CustomPredicate` in the frontend: how do you call an operation to create a `Statement` that uses a `CustomPredicate`

arnaucube:
Closing as done across #83, #97, #101 .

PR #83, opened 2025-02-24
Frontend: simplify custom predicates interfaces

arnaucube:
- migrate middleware::custom::tests types and structs to frontend::custom, adapt it and add comments detailing logic
- simplify custom predicate's frontend interfaces, making it less verbose to define Statement Template arguments

The main idea is that when defining the arguments at a statement template, it can be done from 3 different inputs:
i. `(&str, literal)`: this is to set a POD and a field, ie. `(POD, literal("field")`)
ii. `(&str, &str)`: this is to define a origin-key wildcard pair, ie. `(src_origin, src_dest)`
iii. `Value`: this is to define a literal value, ie. `0`
(this is [documented in the code](https://github.com/0xPARC/pod2/blob/22a309b285e324585375adcb2556e8930fef2cf8/src/frontend/custom.rs#L33), and it's usage can be seen at [the frontend::custom tests](https://github.com/0xPARC/pod2/blob/22a309b285e324585375adcb2556e8930fef2cf8/src/frontend/custom.rs#L191))

This PR is in the direction of #80 .

### DIFF
--- a/src/backends/mock_main/mod.rs
+++ b/src/backends/mock_main/mod.rs
@@ -115,6 +115,11 @@ fn fill_pad<T: Clone>(v: &mut Vec<T>, pad_value: T, len: usize) {
     }
 }
 
+/// Inputs are sorted as:
+/// - SignedPods
+/// - MainPods
+/// - private Statements
+/// - public Statements
 impl MockMainPod {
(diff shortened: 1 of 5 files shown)


ax0 (approved):
LGTM!

PR #90, opened 2025-02-25
Serialize and hash custom predicates

### LINKED ISSUE #75: Hashing of `CustomPredicate` & `CustomPredicateBatch`

tideofwords:
This resolves #75 

Please take a look at my code and give me pointers for coding style.
- I include `params` as an input to `to_fields` so that `to_fields` can pad everything to a fixed length.  I think we have to do this because eventually when we do everything in-circuit, we're going to need to be able to find everything at a constant offset.  Is there a better way to deal with this issue?
- All my `to_fields()` implementations chain together a bunch of vectors, my code is ugly, please suggest more idiomatic Rust.

### DIFF
--- a/.github/workflows/typos.toml
+++ b/.github/workflows/typos.toml
@@ -1,3 +1,4 @@
 [default.extend-words]
 groth = "groth" # to avoid it dectecting it as 'growth'
 BA = "BA"
+ANDed = "ANDed"
--- a/.github/workflows/typos.toml
+++ b/.github/workflows/typos.toml
@@ -1,4 +1,4 @@
 [default.extend-words]
 groth = "groth" # to avoid it dectecting it as 'growth'
 BA = "BA"
-ANDed = "ANDed"
+Ded = "Ded" # "ANDed", it thought "Ded" should be "Dead"
--- a/src/backends/mock_main/mod.rs
+++ b/src/backends/mock_main/mod.rs
@@ -327,7 +327,7 @@ impl MockMainPod {
             statements[statements.len() - params.max_public_statements..].to_vec();
 
         // get the id out of the public statements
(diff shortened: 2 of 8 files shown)


arnaucube (commented):
Nice!
A concern I have is that current code allows to define a `Params.hash_size=4` but on the same time having the method that returns a size different than 4.

This is, each type (Statement, Hash, Value, etc) self-defines which is their amount of fields returned in their `.to_field` method, which is not affected by the `Params.hash_size` value.

For example, the `PodId.to_fields` gets a `hash_size` in the given
`Params.hash_size` passed as input at
[src/middleware/mod.rs#L192](https://github.com/0xPARC/pod2/tree/b12c5dfc4c6b5dfe579cb8be345468ad548c4b71/src/middleware/mod.rs#L192) but then internally this is going to return the amount of fields defined at the method `.to_fields` from the `Hash` type, which depends only on the definition of the type itself, not on the given parameters `Params`; which currently is `4`, but it could be that a `Hash` has `8` field elements (in the case of for example using ~32 bit fields as in Plonky3), so that that method would be returning a vector of length `8` despite the `Params.hash_size=4`. We then should fix it and make the code values match, but ideally we find a way in which the code enforces this match without us having to pay attention (for example not compiling if those values don't match).

<br><br>

One thing I'm thinking is that maybe we should define a parameter/constant that defines the "minimal unit length" (in plonky2 this is `4`, in plonky3 this is `8`, in circom this is `1` (or maybe `2` if we want to support values between 253 and 256 bits too)). And that this parameter should be defined by the backend, so then if we swap backends, the middleware and frontend know this parameter from the backend being used; and this parameter is what would be used for the length when returning at the methods `.to_fields()`.

arnaucube (approved):
LGTM!

PR #91, opened 2025-02-26
Better formatting for SignedPods on frontend

tideofwords:
Show keys as strings as well as hashes

### DIFF
--- a/.github/workflows/typos.toml
+++ b/.github/workflows/typos.toml
@@ -1,3 +1,4 @@
 [default.extend-words]
 groth = "groth" # to avoid it dectecting it as 'growth'
 BA = "BA"
+ANDed = "ANDed"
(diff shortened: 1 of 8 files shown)


arnaucube:
I'm checking this PR, seems that it contains the changes (commits) from the PR #90 , with the extra commit https://github.com/0xPARC/pod2/pull/91/commits/d10dbb7181cfac8d67152ef022acec572ee59c84 . Is this correct?
If so, maybe it can be updated to contain only the change from that commit, to avoid undoing changes that have been done between the #90 and the latest `main` branch.

tideofwords:
> I'm checking this PR, seems that it contains the changes (commits) from the PR #90 , with the extra commit [d10dbb7](https://github.com/0xPARC/pod2/commit/d10dbb7181cfac8d67152ef022acec572ee59c84) . Is this correct? If so, maybe it can be updated to contain only the change from that commit, to avoid undoing changes that have been done between the #90 and the latest `main` branch.

Yes, it's just that one commit.  Can you give me a quick hint how to do this?

tideofwords:
closing this PR in favor of #95 -- it does the same thing but the git history is cleaned up

PR #101, opened 2025-02-28
limit the number of StatementTmpl in CustomPredicate:

### LINKED ISSUE #79: Limit the number of `StatementTmpl` in a `CustomPredicate` via a parameter in `Params`, or limit it to always 2 (hardcoded)

arnaucube:
- add constructor method for CustomPredicate
- make size checks at the CustomPredicate creation, so that once instantiated we can assume that contains valid data

This resolves #79


ax0 (approved):
LGTM!

PR #117, opened 2025-03-06
chore: add statement and KV metadata to frontend PODs

ax0:
Resolves #113.

PR #120, opened 2025-03-06
Aard cleanpod

tideofwords:
Clean up frontend SignedPod format.  Just have frontend key-value pairs, no hashes (because the hashes can be computed from them anyway).  MainPod still to be done, this is just the SignedPod for now.


tideofwords:
Oh no.  Looks like I duplicated #117 ...

PR #127, opened 2025-03-10
chore(frontend): expose literal method in MainPod builder + more op macros

ax0:
This PR exposes the frontend constant-introducing mechanism as a MainPod builder method (`pub_literal`/`priv_literal`) and rounds out the ops exposed via the `op!` macro. These make the ETHDoS example less unwieldy and have been incorporated into it.


arnaucube (approved):
Nice!

Issue #118, opened 2025-03-06
Estimate amount of field ops, hashes, array indexing, and other operations


arnaucube:
From @gubsheep :
Get as best of an estimate as we can on the # of field elements, hashes to merkelize, indexing into arrays, etc. necessary to run a MainPOD with various different parameter sizes, before we get started with the circuit. the work that we have on counting the number of field elements in a statement is valuable, let's have that sort of "benchmarking" on the most important performance dimensions that we can already start to anticipate before writing the circuit.

arnaucube:
Closing this issue as with #130 we can now count how many hashes and tree ops are involved in operations. Counting the amount of field operations and array indexing will be hard because the mock backend does not do those operations. And estimating by hand might end up being a too long task and not adjusting to the actual values of the circuits. The best way to get numbers will be to implement the actual circuits.

PR #130, opened 2025-03-11
add initial counter setup

arnaucube:
Currently counts hashes, tree insertions and tree proof generations. We can extend it to also count the POD operations or other kind of logic that we might want to count.

On the direction of #118 .


src/backends/plonky2/primitives/merkletree.rs
```diff
@@ -319,7 +315,7 @@ impl Node {
         }
     }
 
-    /// Goes down from the current node until it encounters a terminal node,
+    /// Goes down from the current node until it encounter a terminal node,
```
ax0, line 318:
```suggestion
    /// Goes down from the current node until it encounters a terminal node,
```

ax0 (approved):
Nice!

PR #138, opened 2025-03-13
A real yolo-y vibez pod viewer built by claude code that gubsheep and…

alexallain:
… I were playing with, pushing to a separate branch for him to look at


robknight:
Since this was written, I've changed the serialization format a bit. However, I've also added the ability to output a JSON Schema for Signed Pods and Main Pods. This will be helpful to both human and AI developers who need to know what these things can contain, how to format them, and how to handle them in API requests and responses.

PR #141, opened 2025-03-13
Print debugging info if a pod does not verify

tideofwords:


tideofwords:
> This looks very useful to figure out why a check is not passing, but printing can easily add a lot of noise. I have two possible suggestions:
> 
>     * a. Build a nice API to report these failure information via the return type.  Right now I can't think of a clean API for this, because we have `Result<bool>` where `Error` means malformed (and includes error details) and `Ok` means not-malformed and just has a boolean (with no details when `false).
> 
>     * b. Use a logging library.  This would not require an API change.  Logging libraries allow different levels of logging and runtime configuration over what to log and what to skip.  @tideofwords I could update this branch to use logging if you'd like, just let me know!

Good point.  I switched it to logging.

ed255 (approved):
LGTM!

arnaucube:
should we merge this PR?

PR #142, opened 2025-03-14
Playground web UI wip

robknight:


src/server/handlers.rs
```diff
@@ -0,0 +1,240 @@
+use axum::Json;
```
arnaucube, line 1: How would you see placing the `server` outside of the `src`? The reasoning would be to aim for not having any http server related dependencies in the pod2 library.

arnaucube (commented):
How would you see the web-app using pod2 lib through WASM compilation instead through a remote http server?
(in this way it would be a stand-alone web-app, and we would not require adding http server related code in the repo)

src/server/handlers.rs
arnaucube: In case we go with the http server approach, I think it makes sense to have it in a separate crate, so that when an external project imports the pod2 library does not import the server dependencies.

On the wasm vs httpserver approach, I'm not sure yet; I'm slightly inclined in favor of the wasm approach so that the webapp is a standalone app without requiring a remote server, but open to other reasonings.

arnaucube: *separate crate, still in the same repo

ed255:
I've tried the Playground and here's some of my feedback:

I tried creating a MainPod that proves I'm over 18 based on a "passport" signed pod.  But I didn't manage to get the `Gt` statement working:
image

I also noticed that in the previous view, first argument shows "age = 22".  I think it would be nice to show the pod nickname as well: "passport.age = 22".

Things I liked:
- The autosuggestion for entries in the FirstArgument of a MainPod
- The preview of the value behind an anchored key in FirstArgument

If I ask myself: how is this useful to me as a developer of POD2?
- I think the visualization of Pods could be interesting to debug use cases.  This is visualization of SignedPods, MainPods, custom predicates.
- Building a SignedPod or MainPod via a UI doesn't seem super useful.  So for printing a serialized Pod via a rust cli and pasting it into a playground to visualize it would be interesting, but doing further manipulation / creation in a visual manner is less interesting.
- Having a DSL to define custom predicates would be useful, and maybe visualizing them with nice math symbols would be nice.  I think this should not be in the scope of the playground but it's been discussed in the playground context.

If I ask myself: how is this useful as a user of the POD2 system, then I think I can find more satisfying answers:
- Seeing a collection of my PODs, or a history of generated PODs, where I can inspect each POD.
- Creating a POD manually to send it to a friend.  For example I want to prove I have 1000 frog cards to brag.
- Visualizing a query I received and interactively deciding what which statements/pods to use (with the help of a solver)

Issue #133, opened 2025-03-12
SignedPOD verification approach


arnaucube:
We have two main identified options for verifying the [SignedPOD](https://0xparc.github.io/pod2/signedpod.html) inside the MainPOD:

- A) Verify the SignedPOD in the MainPOD circuit, that is `k * merkleproof.verify() + 1 * signature.verify()`
(where `signature.verify() == plonky_proof.verify()` in our current version).

- B) Verify (A) in a independent circuit, and the proof of this circuit is what gets verified in the MainPOD circuit, that is `1 * plonky_proof.verify()`.
And this option (B) takes as 'public inputs' for the plonky_proof verification the key-values that the SignedPOD exposes, and the plonky_proof ensures that their respective merkle_proofs have been verified correctly previously.

gubsheep:
> option (B) takes as 'public inputs' for the plonky_proof verification the key-values that the SignedPOD exposes, and the plonky_proof ensures that their respective merkle_proofs have been verified correctly previously.

I see, this makes sense for why the calculation is the way it is.

One thing I'm still unsure of is to what extent we need to care right now that this intermediate step to outputs something in a consistent format with future intro gadgets or what that format should even be. My first reaction is to do whatever is easiest from an implementation standpoint and see what we learn.

PR #143, opened 2025-03-14
Merkleproof verify circuit

### LINKED ISSUE #124: Implement (standalone) in-circuit merkletree verification
This is a standalone circuit that verifies the entire merkle-proof (2nd approach described at #122 ).

arnaucube:
This PR implements the plonky2 circuit that verifies the merkleproof of existance and merkleproof of non-existence described at #124 and in [docs/merkletree](https://0xparc.github.io/pod2/merkletree.html#proofs-of-inclusion-and-non-inclusion).

This PR resolves #124 and unblocks starting to work on #133.

### DIFF
--- a/src/backends/plonky2/primitives/merkletree_circuit.rs
+++ b/src/backends/plonky2/primitives/merkletree_circuit.rs
@@ -1,4 +1,13 @@
-//! Circuits compatible with the merkletree.rs implementation.
+//! Circuits compatible with the merkletree.rs implementation. This module
+//! offers two different circuits:
+//!
+//! - `MerkleProofCircuit`: allows to verify both proofs of existence and proofs
+//! non-existence with the same circuit.
+//! - `MerkleProofExistenceCircuit`: allows to verify proofs of existence only.
+//!
+//! If only proofs of existence are needed, use `MerkleProofExistenceCircuit`,
+//! which requires less amount of constraints than `MerkleProofCircuit`.
+//!
 use anyhow::Result;
 use plonky2::{
     field::types::Field,
@@ -16,6 +25,10 @@ use plonky2::{
 use crate::backends::plonky2::basetypes::{Hash, Value, D, EMPTY_HASH, EMPTY_VALUE, F, VALUE_SIZE};
 use crate::backends::plonky2::primitives::merkletree::MerkleProof;
 
+/// `MerkleProofCircuit` allows to verify both proofs of existence and proofs
+/// non-existence with the same circuit.
+/// If only proofs of existence are needed, use `MerkleProofExistenceCircuit`,
+/// which requires less amount of constraints.
 pub struct MerkleProofCircuit<const MAX_DEPTH: usize> {
     root: HashOutTarget,
     key: Vec<Target>,
@@ -84,7 +97,7 @@ impl<const MAX_DEPTH: usize> MerkleProofCircuit<MAX_DEPTH> {
             .collect();
 
         // get leaf's hash for the selected k & v
-        let h = Self::kv_hash_target(builder, &k, &v);
+        let h = kv_hash_target(builder, &k, &v);
 
         // if we're in the case i), use leaf_hash=EMPTY_HASH, else use the
         // previously computed hash h.
@@ -96,14 +109,14 @@ impl<const MAX_DEPTH: usize> MerkleProofCircuit<MAX_DEPTH> {
         );
 
         // get key's path
-        let path = Self::keypath_target(builder, &key);
+        let path = keypath_target::<MAX_DEPTH>(builder, &key);
 
         // compute the root for the given siblings and the computed leaf_hash
         // (this is for the three cases (existence, non-existence case i, and
         // non-existence case ii).
         // This root will be assigned in the `set_targets` method, and it is a
         // public input.
-        let root = Self::compute_root_from_leaf(builder, &path, &leaf_hash, &siblings)?;
+        let root = compute_root_from_leaf::<MAX_DEPTH>(builder, &path, &leaf_hash, &siblings)?;
 
         Ok(Self {
             existence,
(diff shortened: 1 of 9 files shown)


ax0 (approved):
LGTM!

arnaucube:
Update: added the circuit for verifying non-existence proofs too.

src/backends/plonky2/primitives/merkletree_circuit.rs
```diff
@@ -0,0 +1,484 @@
...
+            //     input_2 = sibling * (1-s) + h * s = select(s, h, sibling)
+            //     new_h = hash([input_1, input_2])
+            // TODO explore if to group multiple muls in a single gate
+            let input_1: Vec<Target> = (0..4)
+                .map(|j| builder.select(path[i], sibling.elements[j], h.elements[j]))
+                .collect();
+            let input_2: Vec<Target> = (0..4)
+                .map(|j| builder.select(path[i], h.elements[j], sibling.elements[j]))
+                .collect();
+            let new_h = builder.hash_n_to_hash_no_pad::<PoseidonHash>([input_1, input_2].concat());
```
ed255, line 208: In the spec the intermediate node is calculated like this:
https://github.com/0xPARC/pod2/blob/abce0af675ba707b6f07b36b44a72c77036bf5a3/book/src/merkletree.md?plain=1#L43
This implementation is missing the `2` as third input.  The rust implementation has the same discrepancy.

The idea behind padding with `1` and `2` was to avoid collisions when calculating a leaf and intermediate node, so that an attacker can't reuse an intermediate node as a leaf or vice versa.  I think the current implementation also protects against this attack; but we should keep the implementation and spec in sync, so we need to choose one option.

src/backends/plonky2/primitives/merkletree_circuit.rs
```diff
@@ -0,0 +1,484 @@
...
+                .collect(),
+        );
+
+        // get key's path
+        let path = Self::keypath_target(builder, &key);
+
+        // compute the root for the given siblings and the computed leaf_hash
+        // (this is for the three cases (existence, non-existence case i, and
+        // non-existence case ii)
+        let computed_root = Self::compute_root_from_leaf(builder, &path, &leaf_hash, &siblings)?;
```
ed255, line 105: This is a very small optimization.
Since the `computed_root` is equal to `root`, we can just call this `root` and remove some duplicated wires:

```suggestion
        let root = Self::compute_root_from_leaf(builder, &path, &leaf_hash, &siblings)?;
```
And remove this line from the beginning of the function:
```rust
        let root = builder.add_virtual_hash();
```

src/backends/plonky2/primitives/merkletree_circuit.rs
```diff
@@ -0,0 +1,484 @@
...
+
+            // assign the input values to the targets
+            pw.set_target_arr(&key_targ, &key.0.to_vec())?;
+            for i in 0..MD {
+                pw.set_bool_target(expected_path_targ[i], expected_path[i])?;
+            }
+
+            // generate & verify proof
+            let data = builder.build::<C>();
+            let proof = data.prove(pw)?;
```
ed255, line 310: Is there any functionality in plonky2 to verify the constraints without computing the proof?
Some sort of circuit debugger that just goes through all constraints and verifies that they pass after the witness assignment?

I'm asking this because we could use that in tests to run them much faster.

src/backends/plonky2/primitives/merkletree_circuit.rs
```diff
@@ -0,0 +1,484 @@
...
+        Ok(())
+    }
+
+    fn compute_root_from_leaf(
+        builder: &mut CircuitBuilder<F, D>,
+        path: &Vec<BoolTarget>,
+        leaf_hash: &HashOutTarget,
+        siblings: &Vec<HashOutTarget>,
+    ) -> Result<HashOutTarget> {
+        assert!(siblings.len() <= MAX_DEPTH);
```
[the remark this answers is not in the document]
arnaucube: right! updated

src/backends/plonky2/primitives/merkletree_circuit.rs
arnaucube: Right, in this case the circuit is implemented to mimic the rust impl. But seeing now the spec, I think it makes sense to go with the spec approach, so that hashing a leaf and an intermediate node takes the same amount of inputs. This is not specifically relevant for when using plonky2, but might be relevant for other proving systems so that the same hash circuit/chip is reused both for leaf & intermediate node. Will update the rust & circuit impls of the tree with that.

src/backends/plonky2/primitives/merkletree_circuit.rs
arnaucube: Good point! updated

ed255: I've just learned that a virtual target is not mapped to an actual circuit wire.  From the plonky2 code:
```rust
    /// Adds a new "virtual" target. This is not an actual wire in the witness, but just a target
    /// that help facilitate witness generation. In particular, a generator can assign a values to a
    /// virtual target, which can then be copied to other (virtual or concrete) targets. When we
    /// generate the final witness (a grid of wire values), these virtual targets will go away.
    pub fn add_virtual_target(&mut self) -> Target {
```

So my suggestion doesn't bring any optimization.  Nevertheless I still think it's useful to keep the change because it simplifies the code.

src/backends/plonky2/primitives/merkletree_circuit.rs
ed255: Oh I see!  That's unfortunate then.

ed255 (approved):
LGTM!

ax0 (approved):

PR #145, opened 2025-03-17
Fe contains

tideofwords:
DRAFT

Add contains statements for frontend types: 
DictContains
DictNotContains
SetContains
SetNotContains
ArrayContains

Work in progress -- will need to create duplicate frontend::Operation and middleware::Operation, and force the compiler to do the conversion.


src/frontend/predicate.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum NativePredicate {
+    None = 0,
+    ValueOf = 1,
+    Equal = 2,
+    NotEqual = 3,
+    Gt = 4,
+    Lt = 5,
+    Contains = 6,
+    NotContains = 7,
```
robknight, line 17: Are these still needed?

tideofwords:
OK I think this is ready to merge.  It does:
- A Merkle proof is included as part of the "Operation" for a Contains or NotContains statement
- Frontend Contains and NotContains predicates are replaced with different predicates for different frontend compound types (DictContains, ...)

src/frontend/predicate.rs
tideofwords: Nope!  Removed them, thanks for pointing that out :)

src/middleware/operation.rs
```diff
@@ -104,72 +132,95 @@ impl Operation {
...
+            Self::NotContainsFromEntries(_, _, _) => OT::Native(NotContainsFromEntries),
             Self::SumOf(_, _, _) => OT::Native(SumOf),
             Self::ProductOf(_, _, _) => OT::Native(ProductOf),
             Self::MaxOf(_, _, _) => OT::Native(MaxOf),
             Self::Custom(cpr, _) => OT::Custom(cpr.clone()),
         }
     }
 
-    pub fn args(&self) -> Vec<Statement> {
+    pub fn args(&self) -> Vec<OperationArg> {
```
ed255, line 144: My suggestion would be to change this so that `args` still returns `Vec<Statement>`.
And instead of introducing an `OperationArg`, add a new method `Operation::aux(&self) -> OperationAux` which returns the auxiliary data of the operation.

```rust
enum OperationAux {
  None,
  MerkleProof(MerkleProof),
}
```

I think this would simplify things a bit.

Then on the frontend I would suggest doing something similar.

```rust
pub struct Operation(pub OperationType, pub Vec<OperationArg>, pub OperationAux);
```
And keep the `frontend::OperationArg` untouched.

And the same idea in mock_main

ed255 (approved):
I've taken a quick look to see the changes related to this suggestion https://github.com/0xPARC/pod2/pull/145#discussion_r2007160015
and this LGTM!

tideofwords:
(Responding to a question from a call this morning)

Why did I implement this new logic manual_compile_st_op() for DictContains (to introduce a ValueOf statement for EMPTY_VALUE), when we already have a method for converting literal statement arguments to ValueOf constants?

Because I want the literal to be introduced at a different time: I want it to be introduced only in the middleware pod statements, not in the frontend pod statements.

In other words: I want the frontend statements to contain a single statement
`DictContains(root, val)`
which the compiler converts to two middleware statements
```
ValueOf(empty_ak, EMPTY_VALUE)
Contains(root, val, empty_ak).
```

... Are you (@ax0 @ed255 @arnaucube) suggesting that I redesign it, so the frontend statements look like
```
ValueOf(empty_ak, EMPTY_VALUE)      // this now appears to be here for no reason at all
DictContains(root, val)
```
and the backend looks the same as before
```
ValueOf(empty_ak, EMPTY_VALUE)
Contains(root, val, empty_ak)?
```

I guess this would make the code simpler but the frontend pods would have an unexplained ValueOf statement...

ed255:
We discussed the possibility of having a canonical hardcoded ValueOf statement with key=EMPTY_VALUE (or some other reserved key) and value=0.  And then we can always assume this ValueOf statement exists and then turn `DictContains(root, val)` into `Contains(root, val, (SELF, EMTPY_VALUE))` knowing that the EMPTY_VALUE statement will exist.

We also discussed merging this PR as is so that we can advance and keep the discussion open and resolve this later.

And while writing this I thought of a new solution: add `frontend::NativePredicate::Contains` and `NotContains`.
So the user creates operations of `Dict, Set, Array` and passes them to the builder, which in turn translates them into frontend `Contains` and `NotContains` with literals if necessary, and then these go through the process of moving away the literal into a `ValueOf`, all in the builder.
What I'm saying is that the frontend type contains the syntax sugar but also the underlying thing.  And the user can use the syntax sugar at convenience, which is immedietaly converted to the underlying representation already with frontend types.

BTW shouldn't it be `SetContains(root, v) -> Contains(root, v, empty_value)`? and not `DictContains`?

tideofwords:
> BTW shouldn't it be `SetContains(root, v) -> Contains(root, v, empty_value)`? and not `DictContains`?

yes

tideofwords:
I won't be able to do much work on this for at least a couple days so... maybe just merge now and discuss later?

Issue #149, opened 2025-03-18
implement zuKYC circuit


arnaucube:
Includes defining the circuit layout that will be used for general circuits further than zuKYC circuit.

ed255:
This is a list of the pieces that the circuit needs to support ZuKYC.  I'll start implementing a skeleton for it.

## SignPod verification

- MerkleTree inclusion proof of each key-value
- Layout key-value into ValueOf
- Signature verification (mock for now)
- we need 2 of them (gov_id, paystub)

## Statements

- List of public/private statements
- Needed predicates:
    - Eq, Lt, Gt, SetDoesNotContain
- From public statements: take them as array and hash -> Id

## Operations

- Operations that reference past statements
- Needed operations:
    - EqFromEntries, LtFromEntries, GtFromEntries, SetDoesNotContainFromEntries

PR #151, opened 2025-03-18
MainPod circuit skeleton

ed255:
It includes a skeleton for the MainPod verification circuit in plonky2.
It's missing a few things, including tests.

Related to #149

### DIFF
--- a/src/backends/plonky2/common.rs
+++ b/src/backends/plonky2/common.rs
@@ -1,7 +1,7 @@
 //! Common functionality to build Pod circuits with plonky2
 
 use crate::middleware::STATEMENT_ARG_F_LEN;
-use crate::middleware::{Params, Value};
+use crate::middleware::{Params, Value, HASH_SIZE, VALUE_SIZE};
 use plonky2::field::extension::Extendable;
 use plonky2::field::types::PrimeField64;
 use plonky2::hash::hash_types::RichField;
@@ -10,12 +10,12 @@ use plonky2::plonk::circuit_builder::CircuitBuilder;
 
 #[derive(Copy, Clone)]
 pub struct ValueTarget {
-    pub elements: [Target; 4],
+    pub elements: [Target; VALUE_SIZE],
 }
 
 #[derive(Clone)]
 pub struct StatementTarget {
-    pub code: [Target; 6],
+    pub code: [Target; HASH_SIZE + 2],
     pub args: Vec<[Target; STATEMENT_ARG_F_LEN]>,
 }
 
@@ -62,7 +62,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilderPod<F, D>
 
     fn add_virtual_value(&mut self) -> ValueTarget {
         ValueTarget {
-            elements: self.add_virtual_target_arr::<4>(),
+            elements: self.add_virtual_target_arr(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/backends/plonky2/primitives/merkletree_circuit.rs
```diff
@@ -21,39 +21,43 @@ use plonky2::{
...
-    root: HashOutTarget,
-    key: Vec<Target>,
-    value: Vec<Target>,
-    existence: BoolTarget,
-    siblings: Vec<HashOutTarget>,
-    case_ii_selector: BoolTarget, // for case ii)
-    other_key: Vec<Target>,
-    other_value: Vec<Target>,
+    pub root: HashOutTarget,
+    pub key: ValueTarget,
```
arnaucube, line 38: Nice! we discussed this with @ax0 a couple of days ago, about adding this `ValueTarget` type, but seemed a bit extra just for the tree circuit, but makes a lot of sense to have it since it is used at the other parts aside from the tree circuit.

src/backends/plonky2/common.rs
```diff
@@ -0,0 +1,79 @@
...
+use crate::middleware::{Params, Value};
+use plonky2::field::extension::Extendable;
+use plonky2::field::types::PrimeField64;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::target::{BoolTarget, Target};
+use plonky2::plonk::circuit_builder::CircuitBuilder;
+
+#[derive(Copy, Clone)]
+pub struct ValueTarget {
+    pub elements: [Target; 4],
```
arnaucube, line 13:
```suggestion
    pub elements: [Target; VALUE_SIZE],
```

arnaucube: same for the other `4` occurrences (at lines 42 and 49)

src/backends/plonky2/common.rs
```diff
@@ -0,0 +1,95 @@
...
+            .chain(self.args.iter().flatten())
+            .cloned()
+            .collect()
+    }
+}
+
+// TODO: Implement Operation::to_field to determine the size of each element
+#[derive(Clone)]
+pub struct OperationTarget {
+    pub code: [Target; 6],
```
arnaucube, line 35: This `6` comes from `Params::predicate_size()`, which returns `HASH_SIZE + 2`. Maybe we could find a way to avoid having the hardcoded 6 in this file and use it from a basetypes constantant
(mentioning this not because it is an issue having the `6` hardcoded now, but thinking into having a single 'source of truth' of the value, so that if in the future somebody changes the 'operation code length', the other parts that depend on that definition already get the correct value).
Same for the StatementTarget's code.

src/backends/plonky2/main.rs
```diff
@@ -0,0 +1,278 @@
...
+    },
+    iop::{
+        target::{BoolTarget, Target},
+        witness::{PartialWitness, WitnessWrite},
+    },
+    plonk::circuit_builder::CircuitBuilder,
+};
+use std::collections::HashMap;
+
+const MD: usize = 32;
```
arnaucube, line 30: does this stand for "max depth" of the merkletree?

arnaucube (commented):
The layout and structure looks good!

src/backends/plonky2/main.rs
ed255: yes! I'll add a comment

arnaucube (approved):

ax0 (approved):
LGTM!

PR #154, opened 2025-03-19
chore: simplify ToFields trait

### LINKED ISSUE #153: Remove length from to_fields
Currently we have:
```rust
pub trait ToFields {
    /// returns Vec<F> representation of the type, and a usize indicating how many field elements
    /// does the vector contain
    fn to_fields(&self, params: &Params) -> (Vec<F>, usize);
}
```
But the usize is redundant because the vector already contains the length.  So we can remove the second return value.

ed255:
Resolve #153


arnaucube (approved):
LGTM!

PR #158, opened 2025-03-21
organize files & dirs (mostly backend structure)

arnaucube:
I was working on the SignedPod impl, and found files a bit scattered across different levels of directories, this PR aims to group files.

When reviewing it, the gitdiff is almost minimal, the main changes are regarding file placement in the directory tree and their names:
```python
src/
  # group plonky2 circuit related files:
  backends/plonky2/common.rs => backends/plonky2/circuits/common.rs
  backends/plonky2/main.rs => backends/plonky2/circuits/mainpod.rs
  # group mock related files:
  backends/plonky2/mock_main/mod.rs => backends/plonky2/mock/mainpod/mod.rs
  backends/plonky2/mock_main/operation.rs => backends/plonky2/mock/mainpod/operation.rs
  backends/plonky2/mock_main/statement.rs => backends/plonky2/mock/mainpod/statement.rs
  backends/plonky2/mock_signed.rs => backends/plonky2/mock/signedpod.rs
  # we already had the 'examples' directory containing `examples/custom.rs`,
  # move the `examples.rs` there too:
  examples.rs => examples/mod.rs
```

Issue #157, opened 2025-03-20
Implement SignedPod verification circuit


arnaucube:
depends on #147

PR #160, opened 2025-03-21
implement SignedPod (non-mock) using proof-based signatures

### LINKED ISSUE #147: Non-mock SignedPOD implementation
Non-mock SignedPOD implementation, internally using a dictionary and proof-based signatures.

arnaucube:
Resolves #147, unblocks #157.

~Depends on merging #158 first (the *'organize files...'* commit is from the #158); or if preferred we can merge this PR (#160) directly, which includes the changes of #158, but at least they are more clean to review separately.~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ anyhow = "1.0.56"
 dyn-clone = "1.0.18"
 log = "0.4"
 env_logger = "0.11"
+lazy_static = "1.5.0"
 # enabled by features:
 plonky2 = { git = "https://github.com/0xPolygonZero/plonky2", optional = true }
 serde = "1.0.219"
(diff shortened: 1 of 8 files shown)


src/backends/plonky2/mock/signedpod.rs
```diff
@@ -138,17 +141,17 @@ pub mod tests {
...
 
-        assert!(pod.verify());
+        pod.verify()?;
         println!("id: {}", pod.id());
         println!("kvs: {:?}", pod.kvs());
 
         let mut bad_pod = pod.clone();
         bad_pod.signature = "".into();
-        assert!(!bad_pod.verify());
+        bad_pod.verify().is_err();
```
ed255, line 150:
```suggestion
        assert!(bad_pod.verify().is_err());
```
And the same applies to the following cases

ed255 (commented):
Everything looks good to me except for the comment I made where the return value in tests is not being used (asserts are missing)

src/backends/plonky2/mock/signedpod.rs
arnaucube: oh right! my bad; thanks for catching it!, updated it, also used the opportunity to rebase this PR branch to latest `main` branch version.

ed255 (approved):
LGTM!

ax0 (approved):
LGTM!

Issue #156, opened 2025-03-20
Implement all the operations constraints needed for ZuKYC


ed255:
https://github.com/0xPARC/pod2/blob/main/book/src/examples.md#zukyc-simplified-for-p1

- [x] None
- [x] NewEntry
- [x] CopyStatement
- [x] EqualFromEntries
- [ ] ~~GtFromEntries~~
- [x] LtFromEntries
- [x] NotContainsFromEntries

To be implemented in here https://github.com/0xPARC/pod2/blob/22db6ce4c6c2ebbd4b291b779613eda2a16886a1/src/backends/plonky2/main.rs#L120

PR #165, opened 2025-03-25
chore(backend): implement some circuit op logic

ax0:
Towards #156.


src/backends/plonky2/circuits/common.rs
```diff
@@ -157,4 +261,80 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilderPod<F, D>
             self.and(ok, is_eq)
         })
     }
+
+    fn is_value(&mut self, xs: &[Target]) -> BoolTarget {
```
ed255, line 265: Maybe rename `statement_arg_is_value` for clarity?

src/backends/plonky2/circuits/common.rs
```diff
@@ -157,4 +261,80 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilderPod<F, D>
...
+
+        let big_limbs_eq = self.is_equal(x.elements[1], y.elements[1]);
+        let lhs = self.select(big_limbs_eq, x.elements[0], x.elements[1]);
+        let rhs = self.select(big_limbs_eq, y.elements[0], y.elements[1]);
+        assert_limb_lt(self, lhs, rhs);
+    }
+
+    // TODO: Revisit this when we need more than 64 statements.
+    fn vector_ref(&mut self, v: &[Target], i: Target) -> Target {
+        self.random_access(i, v.to_vec())
```
ed255, line 311: Since this will panic with a non-obvious error when the vector is more than 64, maybe add an assert here with a clear message?

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -137,91 +137,211 @@ impl OperationVerifyGate {
...
+        // TODO: Clean this up.
+        let resolved_op_args = if prev_statements.len() == 0 {
+            vec![]
+        } else {
+            op.args
+                .iter()
+                .flatten()
+                .map(|&i| {
+                    StatementTarget::from_flattened(
+                        builder.matrix_row_ref(
```
ed255, line 152: Idea: what if we define a trait to flatten and unflatten:
```rust
trait Flatable {
  fn flaten(&self) -> Vec<Target>;
  fn from_flattened(vs: &[Target]) -> Self;
}
```

And then define some generic methods lile:
```
fn select<T: Flatable>(b: BoolTarget, x: T, y: T) -> T;
fn vec_ref<T: Flatable>(i: usize, vs: &[T]) -> T;
```
These methods would be implemented to flatten the types, apply the logic, and reconstruct the `T`

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -137,91 +137,211 @@ impl OperationVerifyGate {
...
+        // as 'eval' restricted to the op of type X, where the
+        // returned target is `false` if the input targets lie outside
+        // of the domain.
+        let op_checks = vec![
+            vec![
+                self.eval_none(builder, st, op),
+                self.eval_new_entry(builder, st, op, prev_statements),
+            ],
+            // Skip these if there are no resolved op args
+            if resolved_op_args.len() == 0 {
```
ed255, line 178: This case only would happen if we configure the Pod to have 0 input signed pods and 0 input main pods right? From `prev_statements.len() == 0`
Do you think we'll ever have this configuration?

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -137,91 +137,211 @@ impl OperationVerifyGate {
...
-        let new_entry_ok = self.eval_new_entry(builder, st, op);
-        let ok = builder.select_bool(is_new_entry, new_entry_ok, ok);
+        let ok = builder.any(op_checks);
 
         builder.connect(ok.target, _true.target);
 
         Ok(OperationVerifyTarget {})
     }
 
+    fn eval_eq(
```
ed255, line 197: did you consider naming this `eval_eq_from_entries` so that it's clear this is checking an operation and not a statement?

ed255 (approved):
Overall looks good!  I've left some comments with discussions and suggestions.

PR #167, opened 2025-03-26
add circuit to verify signatures (SignatureVerifyGadget)

arnaucube:
This PR implements the `SignatureVerifyGadget`, the circuit that verifies the `Signature`; where since the signatures currently are proof-based signatures, the `SignatureVerifyGadget` internally is doing a 1-level recursive proof verification.

### DIFF
--- a/src/backends/plonky2/primitives/signature_circuit.rs
+++ b/src/backends/plonky2/primitives/signature_circuit.rs
@@ -20,7 +20,9 @@ use plonky2::{
 };
 
 use super::signature::{PublicKey, SecretKey, Signature, DUMMY_PUBLIC_INPUTS, DUMMY_SIGNATURE};
-use crate::backends::plonky2::basetypes::{Hash, Proof, Value, C, D, EMPTY_HASH, EMPTY_VALUE, F};
+use crate::backends::plonky2::basetypes::{
+    Hash, Proof, Value, C, D, EMPTY_HASH, EMPTY_VALUE, F, VALUE_SIZE,
+};
 use crate::backends::plonky2::circuits::common::{CircuitBuilderPod, ValueTarget};
 
 lazy_static! {
@@ -34,6 +36,7 @@ pub struct SignatureVerifyTarget {
     verifier_data_targ: VerifierCircuitTarget,
     selector: BoolTarget,
     pk: ValueTarget,
+    msg: ValueTarget,
     // proof of the SignatureInternalCircuit (=signature::Signature.0)
     proof: ProofWithPublicInputsTarget<D>,
     dummy_proof: ProofWithPublicInputsTarget<D>,
@@ -58,12 +61,30 @@ impl SignatureVerifyGadget {
 
         let common_data = super::signature::VP.0.common.clone();
 
+        // targets related to the 'public inputs' for the verification of the
+        // `SignatureInternalCircuit` proof.
         let pk_targ = builder.add_virtual_value();
+        let msg_targ = builder.add_virtual_value();
+        let inp: Vec<Target> = [pk_targ.elements.to_vec(), msg_targ.elements.to_vec()].concat();
+        let s_targ = builder.hash_n_to_hash_no_pad::<PoseidonHash>(inp);
 
         let verifier_data_targ =
             builder.add_virtual_verifier_data(common_data.config.fri_config.cap_height);
 
         let proof_targ = builder.add_virtual_proof_with_pis(&common_data);
+
+        // connect the {pk, msg, s} with the proof_targ.public_inputs
+        for i in 0..VALUE_SIZE {
+            builder.connect(pk_targ.elements[i], proof_targ.public_inputs[i]);
+            builder.connect(
+                msg_targ.elements[i],
+                proof_targ.public_inputs[VALUE_SIZE + i],
+            );
+            builder.connect(
+                s_targ.elements[i],
+                proof_targ.public_inputs[(2 * VALUE_SIZE) + i],
+            );
+        }
 
         // NOTE: we would use the `conditional_verify_proof_or_dummy` method,
         // but since we're using the `standard_recursion_zk_config` (with zk),
@@ -82,10 +103,11 @@ impl SignatureVerifyGadget {
         );
 
         Ok(SignatureVerifyTarget {
-            selector,
-            proof: proof_targ,
-            pk: pk_targ,
             verifier_data_targ,
+            selector,
+            pk: pk_targ,
+            msg: msg_targ,
+            proof: proof_targ,
             dummy_proof: dummy_proof_targ,
         })
     }
@@ -103,6 +125,7 @@ impl SignatureVerifyTarget {
     ) -> Result<()> {
         pw.set_bool_target(self.selector, selector)?;
         pw.set_target_arr(&self.pk.elements, &pk.0 .0)?;
+        pw.set_target_arr(&self.msg.elements, &msg.0)?;
 
         // note that this hash is checked again in-circuit at the `SignatureInternalCircuit`
         let s = Value(PoseidonHash::hash_no_pad(&[pk.0 .0, msg.0].concat()).elements);
--- a/src/backends/plonky2/primitives/signature_circuit.rs
+++ b/src/backends/plonky2/primitives/signature_circuit.rs
@@ -19,9 +19,9 @@ use plonky2::{
     plonk::proof::{ProofWithPublicInputs, ProofWithPublicInputsTarget},
 };
 
-use crate::backends::plonky2::basetypes::{Hash, Value, C, D, EMPTY_HASH, EMPTY_VALUE, F};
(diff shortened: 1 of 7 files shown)


src/backends/plonky2/primitives/signature_circuit.rs
```diff
@@ -0,0 +1,223 @@
...
+        let circuit = SignatureVerifyGadget {}.eval(&mut builder)?;
+
+        let circuit_data = builder.build::<C>();
+        Ok(circuit_data.verifier_data())
+    }
+}
+
+impl SignatureVerifyGadget {
+    /// creates the targets and defines the logic of the circuit
+    pub fn eval(&self, builder: &mut CircuitBuilder<F, D>) -> Result<SignatureTarget> {
```
arnaucube, line 56: I'm still not onboard with the naming 'eval', since it makes it think like the method is "evaluating", when actually it is setting the targets and the gadget logic.
Since the other method is `set_targets`, I'm fine with `add_targets` for this one, but maybe something else like `build_circuit` (although in some cases will not be a 'circuit' but a 'gadget'). Then maybe something like `build_logic` or `targets_logic` or similar.

src/backends/plonky2/primitives/signature_circuit.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl SignatureVerifyGadget {
+    /// creates the targets and defines the logic of the circuit
+    pub fn eval(&self, builder: &mut CircuitBuilder<F, D>) -> Result<SignatureVerifyTarget> {
+        let selector = builder.add_virtual_bool_target_safe();
+
+        let common_data = super::signature::VP.0.common.clone();
+
+        let pk_targ = builder.add_virtual_value();
```
ed255, line 61: This pk is not connected to anywhere.

You should connect it to the `proof.public_inputs[0]`

Also I think you should expose the message in the `SignatureVerifyTarget`.  That can be done by taking `proof.public_inputs[1]`

And then we're left with `proof.public_inputs[2]` which is the `s` value.  I don't know what should be done with that.  Does the verifier need to calculate `s == H(pk, msg)` and check against the `s` in the public inputs?

src/backends/plonky2/primitives/signature_circuit.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        let proof_targ = builder.add_virtual_proof_with_pis(&common_data);
+
+        // NOTE: we would use the `conditional_verify_proof_or_dummy` method,
+        // but since we're using the `standard_recursion_zk_config` (with zk),
+        // internally it fails to generate the `dummy_circuit`, which mentions
+        // that degree calculation could be off if zk is enabled. So we use
+        // `conditional_verify_proof` feeding in our own dummy_proof
+        // (signature::DUMMY_PROOF).
+        let dummy_proof_targ = builder.add_virtual_proof_with_pis(&common_data);
```
ed255, line 74: I don't understand why the circuit has 2 proofs.
How about this?
- Use only 1 proof
  - We always assign a valid proof.
- The selector is used to pick the exposed (pk, msg) or a dummy hardcoded (pk, msg)
- In witness assignment, if we want to do a dummy proof set the selector to false, and assign a dummy proof that can be verified with the dummy hardcoded pk and msg.

src/backends/plonky2/primitives/signature_circuit.rs
ed255: From the spec, the verification requires calculating `H(pk, msg)` which is missing in the circuit.
But I don't yet understand why that is needed, since the circuit already does that and the circuit proof verification already validates that.
So I don't know the role of `s` and why it is a public input.

arnaucube: Good point!

Regarding exposing the `msg` in the `SignatureVerifyTarget`, agree.

Regarding the 2nd msg on the calculation of `H(pk, msg)`, agree, did this in the `set_targets` section, but it should be done into the constraints definition (`eval` method).

Regarding `s`, it's like a 'challenge' that depends on the `msg` (kind of a challenge that comes from the transcript, where the transcript absorbs `msg` & `pk`); so that the prover (signer) can not generate a valid proof until they know the `msg`. The proof then, ensures that the prover knows `sk` and `msg` at the same time.
When implementing the signature, thought about making `s` non-public input, but then, inside the circuit (`SignatureInternalCircuit`), there is nothing to check the internally-computed `s` against something.

Pushed this new commit https://github.com/0xPARC/pod2/pull/167/commits/d2ebd4f54dd70968c16eb02afb650c40ec008aea , where the computation of `s` is done at the `SignatureVerifyGadget`, and also it connects the `pk`, `msg`, `s` with the `proof.public_inputs` in-circuit.

src/backends/plonky2/primitives/signature_circuit.rs
arnaucube: I'm checking the plonky2's `conditionally_verify_proof` method (
https://github.com/0xPolygonZero/plonky2/tree/6a2c1b47b7b00864c2b65a44c18099b0152f550e/plonky2/src/recursion/conditional_recursive_verifier.rs#L24 ), and it looks like it does the approach of selecting one of the two `proofs` & one of the two `verifier_data`, to then run a single `verify_proof` call.
In our case, we can reuse the same `verifier_data` (we use the same verifier_data target at line 99 & 101: [signature_circuit.rs#L99](https://github.com/0xPARC/pod2/blob/d2ebd4f54dd70968c16eb02afb650c40ec008aea/src/backends/plonky2/primitives/signature_circuit.rs#L99)).
I'm not sure I'm understanding the approach you propose, would it be different than this one?

ax0 (approved):
LGTM!

src/backends/plonky2/primitives/signature_circuit.rs
ed255: My proposal is different.  This is what I mean:
```rust
    pub fn eval(&self, builder: &mut CircuitBuilder<F, D>) -> Result<SignatureVerifyTarget> {
        let selector = builder.add_virtual_bool_target_safe();

        let common_data = super::signature::VP.0.common.clone();

        // targets related to the 'public inputs' for the verification of the
        // `SignatureInternalCircuit` proof.
        let pk_targ = builder.add_virtual_value();
        let msg_targ = builder.add_virtual_value();
        let inp: Vec<Target> = [pk_targ.elements.to_vec(), msg_targ.elements.to_vec()].concat();
        let s_targ = builder.hash_n_to_hash_no_pad::<PoseidonHash>(inp);

        let verifier_data_targ =
            builder.add_virtual_verifier_data(common_data.config.fri_config.cap_height);

        let proof_targ = builder.add_virtual_proof_with_pis(&common_data);

        let pk_targ_dummy = builder.constant_value(Value::from(PK_DUMMY));
        let msg_targ_dummy = builder.constant_value(Value::from(MSG_DUMMY));
        let s_targ_dummy = builder.constant_value(Value::from(S_DUMMY));

        // connect the {pk, msg, s} with the proof_targ.public_inputs conditionally
        let pk_targ_connect = builder.select_value(selector, pk_targ, pk_targ_dummy);
        let msg_targ_connect = builder.select_value(selector, msg_targ, msg_targ_dummy);
        let s_targ_connect = builder.select_value(selector, s_targ, s_targ_dummy);
        for i in 0..VALUE_SIZE {
            builder.connect(pk_targ_connect.elements[i], proof_targ.public_inputs[i]);
            builder.connect(
                msg_targ_connect.elements[i],
                proof_targ.public_inputs[VALUE_SIZE + i],
            );
            builder.connect(
                s_targ_connect.elements[i],
                proof_targ.public_inputs[(2 * VALUE_SIZE) + i],
            );
        }

        builder.verify_proof::<C>(&proof_targ, &verifier_data_targ, &common_data);

        Ok(SignatureVerifyTarget {
            verifier_data_targ,
            selector,
            pk: pk_targ,
            msg: msg_targ,
            proof: proof_targ,
        })
    }
```

It doesn't require twice the space for proof (which has lots of targets)

arnaucube: hmm I still don't understand the intended approach. From what I see, it does not use anywhere the dummy_proof (`DUMMY_SIGNATURE` from the lazy_static), right? I've tried running it and when generating proofs for the disabled verification (which uses the dummy values) the test does not pass. It can be tried in [this branch](https://github.com/0xPARC/pod2/tree/circuit-verify-signature_DUMMY_VERIFICATION)

ed255: Ah you were missing a change in the witness assignment.
This is the commit to get what I mean: https://github.com/0xPARC/pod2/commit/92a576ebac11f01a07fb4ea9b01f7176d39c3965

arnaucube: Oh! I see it now, thanks! I've cherry-picked your commit into the branch of this PR, so that this PR should be ready for merge.

ed255 (approved):
LGTM!

PR #172, opened 2025-04-01
complete full zukyc example doc

ed255:
Document the missing parts of the full ZuKYC example.

This requires defining the `Hash` native predicate and `HashOf` native operation.
The public key derivation is just a hash following the current spec of our signature scheme.  This will change to something more complex if we change the signature scheme, and in that case I think we should rely on an introductory gadget for the key derivation.

I've defined the Hash as taking 2 values so that we can create chains to absorb data of dynamic size.


arnaucube (approved):

ax0 (approved):
LGTM!

PR #191, opened 2025-04-08
Add clippy

ed255:
Also remove the unused hash counter.

I've also pinned the rust nightly version so that we can run the same rust version locally and in the CI.  We can update the rust version from time to time.

From unused variables warnings I fixed some operation processing that was not using all the anchored keys.


src/backends/plonky2/primitives/merkletree_circuit.rs
```diff
@@ -154,6 +154,8 @@ impl MerkleProofGadget {
 
 impl MerkleClaimAndProofTarget {
     /// assigns the given values to the targets
+    // @arnaucube, can we define a MerkleClaimAndProofInput for this?
```
ed255, line 157: @arnaucube this comment is for you

arnaucube: I'm in favor of not bloating the code with a custom struct type for each method that takes multiple inputs, bcs it's extra overhead to deal with, and also would mean having to account for types that we actually don't need. Therefore I'm in favor of using the `#[allow(clippy::too_many_arguments)]` to avoid being forced to bloat the code.

arnaucube (approved):

PR #193, opened 2025-04-10
Use the sanctions set passed in as arg to builder function

robknight:
Rather than using a hard-coded value, use the argument passed to the `zu_kyc_sign_pod_builders`, which was previously being ignored.


ed255 (approved):
LGTM!

PR #194, opened 2025-04-14
WIP Refactor

ed255:
Deduplicate frontend/middleware types for all types except for Operation where any operation can take literals as arguments in the frontend.


ed255:
I've refactored how custom predicates work slightly.
When I say "custom statement" I mean a statement built from a custom predicate.
- Previously custom statements took anchored keys as arguments.  Now they take WildcardValue
- Previously custom statements required arguments via pairs (from the previous point)
- New argument type for statements is WildcardValue.  It's very similar to a Value but can only be Key or PodId.  It still maps to 4 field elements
- Now a statement template can use wildcards in: the pod_id of an anchored key, the key of an anchored key, a wildcard value (new).

From these changes we need updated parameters: the eth_dos custom statements now require 6 arguments (before the default max was 5).  This will affect circuit performance from the random accesses.

ax0 (approved):

Issue #183, opened 2025-04-07
Refactor: preprocess the witness outside of the circuit `set_targets` methods


ed255:
For example padding is one kind of processing.

PR #196, opened 2025-04-16
move MainPod layouting to mainpod

### LINKED ISSUE #182: Refactor: reuse layouting and padding code between MockMainPod and MainPod

ed255:
Resolve #182 by
- Create a folder module for `mainpod`
- Move `src/backends/plonky2/mock/mainpod/operation.rs` and `src/backends/plonky2/mock/mainpod/statement.rs` to `src/backends/plonky2/mainpod/`
- Move the common functions for main pod layouting from `src/backends/plonky2/mock/mainpod/mod.rs` to `src/backends/plonky2/mainpod/mod.rs` 
- Now the mock main pod is a single file instead of a folder module.
  - Now the mock main pod imports the layouting functions from main pod instead of the other way around.

I've also
- moved the `MerkleClaimAndProof` to the `merkletree_circuit.rs` to keep things tidy
- changed `MerkleClaimAndProof` to wrap the already existing `MerkleProof` type 
- changed the `set_targets` in the merkle circuits to take `MerkleClaimAndProof` as input (and in the process noticed we had a duplicate field of `existence`)
  - with this change the padding of the siblings now happens outside of `set_targets`, in the creation of the `MerkleClaimAndProof`.  This follows https://github.com/0xPARC/pod2/issues/183


ax0 (approved):

src/backends/plonky2/primitives/merkletree.rs
```diff
@@ -260,6 +264,99 @@ impl MerkleProof {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct MerkleClaimAndProof {
+    /// `enabled` determines if the merkleproof verification is enabled
+    pub enabled: bool,
```
arnaucube, line 270: Initially I was not sure on having this boolean in the proof structure, but I understand that is in order to do stuff like: https://github.com/0xPARC/pod2/tree/cedd98dbe10433b0c030f1b0916485cef3773a90/src/backends/plonky2/circuits/mainpod.rs#L144 .
The concern was about devs missusing the merkletree module (like verifying a disabled proof as it were enabled), but I understand that this is not the target of the module, as the target usage of the merkletree module is not for external usage but only for internal usage by other modules of POD, right?

arnaucube (approved):
LGTM!

src/backends/plonky2/primitives/merkletree.rs
ed255: > The concern was about devs missusing the merkletree module (like verifying a disabled proof as it were enabled), but I understand that this is not the target of the module, as the target usage of the merkletree module is not for external usage but only for internal usage by other modules of POD, right?

That's right, the target usage is internal.  Are you thinking about removing the `enabled` from `MerkleClaimAndProof` and then having this API for `set_targets`?
```rust
impl MerkleClaimAndProofTarget {
    /// assigns the given values to the targets
    pub fn set_targets(
        &self,
        pw: &mut PartialWitness<F>,
        enabled: bool,
        mp: &MerkleClaimAndProof,
    ) -> Result<()> {
```

Perhaps that's more clear?

ed255: Doing the logic at https://github.com/0xPARC/pod2/tree/cedd98dbe10433b0c030f1b0916485cef3773a90/src/backends/plonky2/circuits/mainpod.rs#L144 with the `enabled` outside would be more cumbersome because currently we're including the `enabled` in the random access.
But we could just force that all merkle proofs have enabled=true and fill them with valid proofs and then we don't need to add `enabled` into the random access.

arnaucube: Thanks for the explanation ^^
I don't have any strong opinion, and agree with that removing the `enabled` field from the claimandproof struct would add a bit of extra overhead in the code. So, happy with the way it is.

PR #197, opened 2025-04-16
migrate from anyhow to thiserror

### LINKED ISSUE #190: Migrate from `anyhow` to `thiserror`
We discussed it [a couple months ago](https://github.com/0xPARC/pod2/issues/12), but back then we were fine with `anyhow`. Recently we've been dealing more with error conditions, so we just agreed (in an impromptu meeting) to migrate to `thiserror` so that we have error types (enums).

arnaucube:
This PR migrates from using `anyhow` to `thiserror` as agreed in #190 . Contains an initial commit that does the migration, but not as a final version to be merged.

@ax0 @ed255 feel free to update it with commits polishing the error messages. All the occurrences of the string `Error::Custom(format!`, are error messages that were not used more than once an it was not clear if a specific error type was needed for them. Maybe how we can approach it is that if the error message will not be used in the upper layers, it can be as a `Error::Custom(String)`, as it was before with anyhow.

Once ready, this will resolve #190.


src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+
+pub type Result<T, E = Error> = core::result::Result<T, E>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    // Wrappers on top of other errors
+    #[error("std::io::Error: {0}")]
+    IOError(#[from] std::io::Error),
+    #[error("anyhow::Error: {0}")]
+    Anyhow(#[from] anyhow::Error),
```
ed255, line 11: It would be interesting if we can distinguish plonky2 errors (which are `anyhow::Error`) from other `anyhow::Errors`).  But perhaps that would require a lot of boilerplate in the code :/

src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error(transparent)]
+    Middleware(#[from] crate::middleware::error::MiddlewareError),
+
+    // Comparators errors
+    #[error("not equal")]
+    NotEqual,
+    #[error("{0} {1} is over the limit {2}")]
+    MaxLength(String, usize, usize),
+    #[error("{0} amount of {1} should be {1} but it's {2}")]
+    DiffAmount(String, String, usize, usize),
```
ed255, line 27: Did you consider naming the fields so that the error creation is more clear?
For example:
```rust
DiffAmount{object: String, unit: String, expect: usize, found: usize}
```
The same applies to other error cases that have several fields

src/examples/custom.rs
```diff
@@ -9,10 +8,11 @@ use crate::{
         CustomPredicateBatch, CustomPredicateRef, NativePredicate as NP, Params, PodType,
         Predicate, KEY_SIGNER, KEY_TYPE,
     },
+    Error,
```
ed255, line 11: Can we use `Result` here?  Then we don't need to explicitly say `Result<..., Error>` and just do `Result<...>`

src/frontend/mod.rs
```diff
@@ -271,7 +273,7 @@ impl MainPodBuilder {
             // We are dealing with a copy here.
             match (args).first() {
                 Some(OperationArg::Statement(s)) if args.len() == 1 => Ok(s.predicate().clone()),
-                _ => Err(anyhow!("Invalid arguments to copy operation: {:?}", args)),
+                _ => Err(Error::OpInvalidArgs("copy".to_string())),
```
ed255, line 276: oops, the `args` is not captured by the error.  I see that we were not doing that in some of the following errors before.

src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error("repeated ValueOf")]
+    RepeatedValueOf,
+    #[error("Statement did not check")]
+    StatementNotCheck,
+    #[error("Key not found")]
+    KeyNotFound,
+
+    // Other
+    #[error("{0}")]
+    Custom(String),
```
ed255, line 49: Could we remove this `Custom` entry and use `anyhow` instead?
This way we can use the anyhow macro for convenience and it should get converted into `Error::Anyhow`.

src/middleware/custom.rs
```diff
@@ -9,6 +8,7 @@ use plonky2::field::types::Field;
 use crate::{
     middleware::HASH_SIZE,
     middleware::{hash_fields, Hash, Key, NativePredicate, Params, ToFields, Value, F},
+    Error,
```
ed255, line 11: Can we import `Result` here instead?

src/middleware/operation.rs
```diff
@@ -280,11 +279,11 @@ impl Operation {
                         StatementArg::Key(ak2.clone()),
                     ])
                 } else {
-                    return Err(anyhow!("Invalid operation"));
+                    return Err(Error::Middleware(MiddlewareError::InvalidOp));
```
ed255, line 282: is there a way to get automatic conversion from `MiddlewareError` to `Error` here?
So that we can write this:
```rust
return Err(MiddlewareError::InvalidOp);
```

src/backends/plonky2/mainpod.rs
```diff
@@ -99,11 +100,7 @@ impl Pod for MainPod {
         // 2. get the id out of the public statements
         let id: PodId = PodId(hash_statements(&self.public_statements, &self.params));
         if id != self.id {
-            return Err(anyhow!(
-                "id does not match, expected {}, computed {}",
-                self.id,
-                id
-            ));
+            return Err(Error::Middleware(MiddlewareError::IdNotEqual(self.id, id)));
```
ed255, line 103: Is it ok to create a middleware error in the backend?

src/backends/plonky2/primitives/signature.rs
```diff
@@ -121,6 +121,7 @@ impl Signature {
             proof: self.0.clone(),
             public_inputs,
         })
+        .map_err(|_| Error::Plonky2ProofFail)
```
ed255, line 124: is it worth encapsulating the verify error instead of discarding it?

ed255 (commented):
I've taken an initial look at the PR and left a lot of comments to discuss details :)

ed255:
One major thing that I find missing from the proposal is that with this change we're losing the backtrace in the errors.  `anyhow` includes backtraces in errors created with the `anywhow!` macro, so if there's an `anywhow::Error` displayed with the env var `RUST_BACKTRACE=1`, the backtrace attached to the error is also displayed.

I see in the documentation of `thiserror` mentions of backtraces https://docs.rs/thiserror/latest/thiserror/ but we would need to change the error type to include them and probably use a macro to create the errors so that the backtrace is generated.

arnaucube:
Agree with everything, for context (@ax0 ), we had a call with @ed255 , and he is going to try adding macros for defining & returning the errors (defined by thiserror), while also returning the backtrace.

Update on the PR: got rid of the git-conflicts while rebasing the PR branch to last `main` branch updates.

src/middleware/custom.rs
arnaucube: Agree, done at https://github.com/0xPARC/pod2/compare/d9453797e2f437aefa43712667dc2cc9aac3e20a..92fcdf35427c4f2f90264530b0a27032e2adc170

ed255:
I'm currently working on this PR, although I'm not finished yet.  Here are my goals:
- Include backtraces in the errors we generate.  To get this we can't just return a literal enum, because the backtrace requires a call.
- Related to the previous point: I'll create methods to create errors so we can include the backtrace conveniently without changing too much the syntax.  So instead of `Err(Error::KeyNotFound(key))` (literal enum) it will be `Err(Error::key_not_found(key))` (method call)
- I think each error should be local to its scope, and each scope should only return its own error.
  - The merkle tree should return `TreeError` and not Error
  - The middleware should return `MiddlewareError` and not Error
- With a global Error we can't easily include backend/frontend types in the error fields, so I'll declare a BackendError and a FrontendError and follow the pattern from the previous point
- The Pod traits should be able to return backend errors and will be used in the frontend; for that I'll change them to use trait object Error: `dyn std::error::Error`

ed255:
Topics to discuss:
- Do you prefer types `FrontendError, MiddlewareError, BackendError`? or `frontend::Error, middleware::Error, backend::Error`?
- Do you prefer type aliases `FrontendResult, MiddlewareResult, BackendResult`? or `frontend::Result, middleware::Result, backend::Result`?

ed255:
@arnaucube could you please review the new commits?

src/frontend/error.rs
```diff
@@ -0,0 +1,67 @@
+use std::{backtrace::Backtrace, fmt::Debug};
+
+use crate::middleware::{DynError, Statement, StatementTmpl};
+
+pub type FrontendResult<T, E = FrontendError> = core::result::Result<T, E>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum InnerError {
+    #[error(transparent)]
+    Tree(#[from] crate::backends::plonky2::primitives::merkletree::error::TreeError),
```
arnaucube, line 10: is this one used? I've tried removing it and seems to not be used. I was trying this bcs I was thinking if maybe we could avoid using the `TreeError` since it could be 'embedded' in the `backend::Error` already.

ax0 (approved):
LGTM!

arnaucube (commented):
LGTM!
I can not mark it as 'approve' in the GitHub UI since I opened the PR, but everything looks good from my side ^^

PR #199, opened 2025-04-17
chore(middleware): remove output statement logic

### LINKED ISSUE #189: Refactor: Fix or remove output statement logic in middleware
The logic [here](https://github.com/0xPARC/pod2/blob/1214cdfa1bfa32c4880318d000cb8e37fe1d22fb/src/middleware/operation.rs#L257) should be removed or fixed, viz.
- The output statements (e.g. SumOf) should have the right arguments.
- The error cases should be combined into one case with a useful error message.

ax0:
Resolves #189.


src/middleware/operation.rs
```diff
@@ -262,180 +262,6 @@ impl Operation {
             OperationType::Custom(cpr) => Self::Custom(cpr, args.to_vec()),
         })
     }
-    /// Gives the output statement of the given operation, where determined
```
ed255, line 265: this was easy :D

ed255 (approved):
LGTM!

PR #200, opened 2025-04-17
Padding in set target

ed255:
Resolve https://github.com/0xPARC/pod2/issues/183 with a twist

The issue was about processing the witness outside of the `set_target` method, and that processing includes padding.  Part of this change was already started in https://github.com/0xPARC/pod2/pull/196

My original arguments for this pattern were:
- Make the code around the circuit as simple and transparent as possible so that we focus on the constraint logic
- Allow writing tests that use `set_targets` and directly set all the witness values

But once I started applying this pattern in the MainCircuit I noticed some issues.  Here are my counter arguments:
- The code looks more clean by doing the padding inside `set_targets` because it allows reusing existing types as arguments to `set_targets` (and those types are not yet padded).  It also allows having the padding logic in the circuit code, which makes sense because the circuit logic defines what is considered padding. 
- We're not doing any test that requires direct manipulation of witness values that would not be supported by `set_targets` methods that do some processing
  - If we had such requirement we could resolve it with: a `test_set_targets` method that does what the test needs.  Or setting the targets in the test by making the `FooTarget` fields `pub(crate)`

So for this reason I'm reverting the merkle tree padding outside of `set_target` from  https://github.com/0xPARC/pod2/pull/196

In this PR i also remove the `enabled` from `MerkleClaimAndProof` following the discussion https://github.com/0xPARC/pod2/pull/196#discussion_r2048325507 because it makes things more clear and less error prone, and the `enabled: bool` is not actually used in the random access, it's the `BoolTarget` counterpart, which still exists inside of `MerkleClaimAndProofTarget`.

### DIFF
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -707,12 +707,11 @@ mod tests {
         );
 
         let merkle_proofs = vec![MerkleClaimAndProof::new(
-            params.max_depth_mt_gadget,
             Hash::from(root.raw()),
             key,
             None,
-            &no_key_pf,
-        )?];
+            no_key_pf,
+        )];
         let prev_statements = vec![root_st, key_st];
         operation_verify(st, op, prev_statements, merkle_proofs.clone())?;
 
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -492,11 +492,15 @@ impl MainPodVerifyTarget {
             self.statements[i].set_targets(pw, &self.params, st)?;
             self.operations[i].set_targets(pw, &self.params, op)?;
         }
-        assert_eq!(input.merkle_proofs.len(), self.params.max_merkle_proofs);
+        assert!(input.merkle_proofs.len() <= self.params.max_merkle_proofs);
         for (i, mp) in input.merkle_proofs.iter().enumerate() {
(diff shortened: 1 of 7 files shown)


arnaucube (approved):
LGTM, happy with moving the padding & witness preparation logic to the `set_targets` methods :smiley_cat:

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -492,11 +492,15 @@ impl MainPodVerifyTarget {
...
         for (i, mp) in input.merkle_proofs.iter().enumerate() {
-            assert_eq!(mp.proof.siblings.len(), self.params.max_depth_mt_gadget);
-            self.merkle_proofs[i].set_targets(pw, mp)?;
+            self.merkle_proofs[i].set_targets(pw, true, mp)?;
         }
+        let pad_mp = MerkleClaimAndProof::empty();
+        for i in input.merkle_proofs.len()..self.params.max_merkle_proofs {
+            self.merkle_proofs[i].set_targets(pw, false, &pad_mp)?;
+        }
+        // Padding
```
ax0, line 503:
```suggestion
        // Padding
        let pad_mp = MerkleClaimAndProof::empty();
        for i in input.merkle_proofs.len()..self.params.max_merkle_proofs {
            self.merkle_proofs[i].set_targets(pw, false, &pad_mp)?;
        }
```

PR #201, opened 2025-04-18
Re-implement serialization

### LINKED ISSUE #195: Reimplement serialization/deserialization
In https://github.com/0xPARC/pod2/pull/194 we merged the frontend and middleware types and at the same time commented out all the serialization/deserialization code to be more fluid with the refactor.
We need to reimplement the serialization/deserialization on the new middleware types.  Now all middleware types contain rich information so we shouldn't need frontend type serialization.

My suggestion to serialize and deserialize `Value` and `Key` is to:
- serialize only the `TypedValue` in `Value` and `String` in `Key`
- on deserialization, reconstruct `RawValue` in `Value` and `Hash` in `Key`.

robknight:
Closes #195 

Re-implements serialization based on the new refactored code.

Serialization format should be pretty similar to how it was before, with some small cosmetic changes.

I've had to implement a few more custom serialization functions, in order to avoid serializing middleware data. I've commented in the places where this was necessary, so that we can be aware that those custom serialization functions will not automatically track changes to the types in the way that Serde's attribute-based serialization does.

The new version has more tests, and in particular has an end-to-end test which:

- Serializes a SignedPod, and two quite different MainPods
- Generates JSON Schemas for SignedPod and MainPod
- Tests that the real SignedPod and MainPods can be validated at runtime using the schemas

### DIFF
--- a/src/backends/plonky2/mock/signedpod.rs
+++ b/src/backends/plonky2/mock/signedpod.rs
@@ -44,7 +44,7 @@ pub struct MockSignedPod {
 }
 
 impl MockSignedPod {
-    pub fn deserialize(id: PodId, signature: String, kvs: HashMap<Key, Value>) -> Self {
+    pub(crate) fn new(id: PodId, signature: String, kvs: HashMap<Key, Value>) -> Self {
         Self { id, signature, kvs }
     }
 }
--- a/src/backends/plonky2/mock/signedpod.rs
+++ b/src/backends/plonky2/mock/signedpod.rs
@@ -43,15 +43,11 @@ pub struct MockSignedPod {
     kvs: HashMap<Key, Value>,
 }
 
-// impl MockSignedPod {
-//     pub fn deserialize(id: PodId, signature: String, dict: Dictionary) -> Self {
-//         Self {
-//             id,
-//             signature,
-//             dict,
-//         }
-//     }
-// }
+impl MockSignedPod {
+    pub fn deserialize(id: PodId, signature: String, kvs: HashMap<Key, Value>) -> Self {
+        Self { id, signature, kvs }
+    }
+}
 
 impl Pod for MockSignedPod {
     fn verify(&self) -> Result<()> {
--- a/src/frontend/serialization.rs
+++ b/src/frontend/serialization.rs
@@ -1,19 +1,18 @@
-/*
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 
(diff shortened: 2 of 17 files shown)


src/backends/plonky2/mock/signedpod.rs
```diff
@@ -43,15 +43,11 @@ pub struct MockSignedPod {
...
-//     pub fn deserialize(id: PodId, signature: String, dict: Dictionary) -> Self {
-//         Self {
-//             id,
-//             signature,
-//             dict,
-//         }
-//     }
-// }
+impl MockSignedPod {
+    pub fn deserialize(id: PodId, signature: String, kvs: HashMap<Key, Value>) -> Self {
```
ed255, line 47: this looks like a constructor with a different name :P 
Will this method be used externally?  If not, can we make it `pub(crate)`?

src/frontend/serialization.rs
```diff
@@ -30,10 +30,9 @@ impl TryFrom<SignedPodHelper> for SignedPod {
             return Err(anyhow::anyhow!("pod_type is not Mock"));
```
ed255, line 30: What about deserializing a non-mock SignedPod?
Would this implementation have a switch here?

Same for the MainPod.

ed255 (approved):
LGTM!  I've left two comments.
One is a small suggestion, the other is more a question/discussion.

src/backends/plonky2/mock/signedpod.rs
robknight: It does have to be public if we want users of the library to be able to deserialize SignedPods.

src/frontend/serialization.rs
robknight: When I first wrote this, there were no non-mock SignedPods! Are the new non-mock Pods usable enough for me to implement serialization for them?

ed255: Both the non-mock SignedPod and MainPod should be ready enough to get serialization and deserialization :)
I would suggest doing that in another PR though.

src/backends/plonky2/mock/signedpod.rs
ed255: Won't users of the library interface the library through the frontend?  And deserialize `frontend::SignedPod` (which internally deserializes `MockSignedPod`)?

robknight: Oh, sorry, you're quite right. I'll make it `pub(crate)`.

PR #205, opened 2025-04-21
make Pod derive from Any

dgulotta:
This commit makes Pod inherit Any, which removes the need for the as_any and into_any functions.

It requires the `trait_upcasting` feature, which was considered experimental in the `nightly-2025-01-20` toolchain, but has since been stabilized: https://github.com/rust-lang/rust/issues/65991

### DIFF
--- a/src/backends/plonky2/mainpod/mod.rs
+++ b/src/backends/plonky2/mainpod/mod.rs
@@ -293,8 +293,7 @@ impl PodProver for Prover {
             .signed_pods
             .iter()
             .map(|p| {
-                let p = p
-                    .as_any()
+                let p = (*p as &dyn Any)
                     .downcast_ref::<SignedPod>()
                     .expect("type SignedPod");
                 p.clone()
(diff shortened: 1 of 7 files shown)


src/backends/plonky2/mainpod/mod.rs
```diff
@@ -293,9 +293,7 @@ impl PodProver for Prover {
             .signed_pods
             .iter()
             .map(|p| {
-                let p = p
-                    .as_any()
-                    .downcast_ref::<SignedPod>()
+                let p = (*p as &dyn Any).downcast_ref::<SignedPod>()
                     .expect("type SignedPod");
```
ax0, line 297:
```suggestion
                let p = (*p as &dyn Any)
                    .downcast_ref::<SignedPod>()
                    .expect("type SignedPod");
```
cf. `cargo fmt --check`.

ax0 (commented):
Nice! `cargo test --release` seems to be failing. Could you replace all instances of `into_any()` in the tests as well?

ed255 (approved):
Nice! LGTM!

ax0 (approved):

PR #207, opened 2025-04-23
apply new circuit naming convention

### LINKED ISSUE #181: Circuits naming convention
Currently we have (using the merkletree gadget as example):
```rust
struct MerkleProofGadget { // contains the config params
  max_depth: usize,
}

impl MerkleProofGadget {
  fn eval(&self, builder) -> Result<MerkleProofTarget> {
    // logic of the gadget
  }
}

struct MerkleProofTarget {
    max_depth: usize,
    enabled: BoolTarget,
    root: HashOutTarget,
    key: ValueTarget,
    value: ValueTarget,
    existence: BoolTarget,
    siblings: Vec<HashOutTarget>,
    case_ii_selector: BoolTarget,
    other_key: ValueTarget,
    other_value: ValueTarget,
}
impl MerkleProofTarget{
  fn set_targets(&self, partial_witness, inputs) -> Result<()> {
    // assign inputs to the targets
  }
}
```

Then to use it we have to do:
```rust
let targets = MerkleProofGadget { max_depth }.eval(&mut builder)?;
targets.set_targets( &mut pw, true, true, tree2.root(), proof.clone(), key, value)?;
```

I find the `eval` naming confusing, since it makes it look like the method is "evaluating", when actually it is setting the targets and the gadget logic (ie. building the gadget/circuit logic).
Since the other method is `set_targets`, maybe an option would be using `add_targets` instead of `eval`, but not convinced; maybe something else like `build_circuit` (although in some cases will not be a 'circuit' but a 'gadget'). Then maybe something like `build_logic` or `targets_logic` or similar. Or, since we're already in the `SomethingGadget` context, can be directly `SomethingGadget::build()`.

Also we can group the two structs to simplify a bit the interfaces, something like:

```rust
struct MerkleProofGadget {
    max_depth: usize,
    enabled: BoolTarget,
    root: HashOutTarget,
    key: ValueTarget,
    value: ValueTarget,
    existence: BoolTarget,
    siblings: Vec<HashOutTarget>,
    case_ii_selector: BoolTarget,
    other_key: ValueTarget,
    other_value: ValueTarget,
}

impl MerkleProofGadget {
  // NOTE: here the name 'build' could be 'build_logic', 'define_logic', 'define_gadget', etc
  fn build(builder, params) -> Result<Self> {
    // logic of the gadget
  }

  fn set_targets(&self, partial_witness, inputs) -> Result<()> {
    // assign inputs to the targets
  }
}
```

Then we would use it as
```rust
let targets = MerkleProofGadget::build(&mut builder, max_depth)?;
targets.set_targets( &mut pw, true, true, tree2.root(), proof.clone(), key, value)?;
```

Opening this issue to have a separate thread to discuss it, hoping to include it
in the refactor (#171).

ed255:
Resolve #181 
I'm setting this as a draft because some parts of this change don't feel right to me.

Things that don't feel right:
- In the MainPod circuit there were two methods that return a `MainPodVerifyTarget` one was for the `MainPodVerifyGadget` (internall unit) and the other one was for `MainPodVerifyCircuit` (external unit, which contains the gadget and makes some targets public).  But now we got rid of the Gadget and Circuit types.  I named the second method `build_circuit` but I don't know if it's very clear
- We had OperationVerifyGadget which added constraints but didn't return any target, just a `Result<()>`.  Now we have an empty `OperationVerifyTarget` which is weird because it doesn't contain any target.

### DIFF
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -378,7 +378,7 @@ impl MainPodVerifyTarget {
             for _statement in 0..params.max_public_statements {
                 statements.push(StatementTarget::new_native(
                     builder,
-                    &params,
+                    params,
                     NativePredicate::None,
                     &[],
                 ))
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -13,13 +13,11 @@ use crate::{
                 CircuitBuilderPod, Flattenable, MerkleClaimTarget, OperationTarget,
                 StatementTarget, ValueTarget,
             },
-            signedpod::{SignedPodVerifyGadget, SignedPodVerifyTarget},
+            signedpod::SignedPodVerifyTarget,
         },
         error::Result,
         mainpod,
-        primitives::merkletree::{
-            MerkleClaimAndProof, MerkleClaimAndProofTarget, MerkleProofGadget,
-        },
+        primitives::merkletree::{MerkleClaimAndProof, MerkleClaimAndProofTarget},
         signedpod::SignedPod,
     },
     middleware::{
@@ -32,14 +30,12 @@ use crate::{
 // MainPod verification
 //
 
-struct OperationVerifyGadget {
-    params: Params,
-}
+struct OperationVerifyTarget {}
 
-impl OperationVerifyGadget {
-    fn eval(
-        &self,
+impl OperationVerifyTarget {
+    fn build(
         builder: &mut CircuitBuilder<F, D>,
+        params: &Params,
         st: &StatementTarget,
         op: &OperationTarget,
         prev_statements: &[StatementTarget],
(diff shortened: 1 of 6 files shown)


src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -32,14 +30,12 @@ use crate::{
 // MainPod verification
 //
 
-struct OperationVerifyGadget {
-    params: Params,
-}
+struct OperationVerifyTarget {}
```
ax0, line 33: I have mixed feelings about this. On the one hand, no new targets come out of the op verification process, so the struct ought to be empty, but this could be better served with a label other than 'target'. On the other hand, an alternative approach would be to view `build` as analogous to the out-of-circuit `check` method and implement it on `OperationTarget`, doing away with this struct entirely. Perhaps the drawback to the latter approach is that we are sweeping the circuit builder mutating aspect under the rug.

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -452,26 +451,16 @@ impl MainPodVerifyGadget {
...
-    merkle_proofs: Vec<MerkleClaimAndProofTarget>,
-}
-
-pub struct MainPodVerifyInput {
-    pub signed_pods: Vec<SignedPod>,
-    pub statements: Vec<mainpod::Statement>,
-    pub operations: Vec<mainpod::Operation>,
-    pub merkle_proofs: Vec<MerkleClaimAndProof>,
-}
+    pub fn build_circuit(
```
ax0, line 455: This one is a bit subtle. I would be in favour of something more verbose like `build_with_public_id`.

PR #209, opened 2025-04-24
chore(backend): make Plonky2 config more explicit

ax0:
This PR makes the Plonky2 config in use more explicit, making it easier to swap out the extension degree and field for another in testing.


ed255 (approved):
LGTM!

PR #216, opened 2025-04-28
chore: implement Gt and GtEq as syntactic sugar

### LINKED ISSUE #214: Implement Gt and GtEq as syntax sugar
We will have Lt and LtEq implemented in the backend (both native predicate and corresponding operation from entries).

In the middleware we will also define `Gt` and `GtEq` as syntax sugar like the container operations we have right now:
https://github.com/0xPARC/pod2/blob/29545f03fcc45e0cf3604bf5f7f79e5310f186d5/src/middleware/operation.rs#L74-L80

Then the frontend is responsible of doing the following conversions:
- `Gt(a, b)` to `Lt(b, a)`
- `GtEq(a, b)` to `LtEq(b, a)`

ax0:
Resolves #214.

### DIFF
--- a/src/backends/plonky2/circuits/common.rs
+++ b/src/backends/plonky2/circuits/common.rs
@@ -310,6 +310,7 @@ pub trait CircuitBuilderPod<F: RichField + Extendable<D>, const D: usize> {
     // Convenience methods for checking values.
     /// Checks whether `xs` is right-padded with 0s so as to represent a `Value`.
     fn statement_arg_is_value(&mut self, arg: &StatementArgTarget) -> BoolTarget;
+
     /// Checks whether `x` is an i64, which involves checking that it
     /// consists of two `u32` limbs.
     fn assert_i64(&mut self, x: ValueTarget);
--- a/src/backends/plonky2/circuits/common.rs
+++ b/src/backends/plonky2/circuits/common.rs
@@ -64,6 +64,15 @@ impl ValueTarget {
     }
 }
 
+/// StatementArgTarget to ValueTarget coercion. Make sure to check
+/// that the arg is a value using the `statement_arg_is_value` method
+/// first!
+impl From<&StatementArgTarget> for ValueTarget {
+    fn from(st_arg: &StatementArgTarget) -> Self {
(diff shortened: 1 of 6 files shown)


src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -276,6 +274,62 @@ impl OperationVerifyGadget {
...
+
+        // Check for equality.
+        let args_equal = builder.is_equal_slice(
+            &arg1_value.elements[..VALUE_SIZE],
+            &arg2_value.elements[..VALUE_SIZE],
+        );
+        let lt_check_flag = [op_code_ok, builder.not(args_equal)]
+            .into_iter()
+            .reduce(|x, y| builder.and(x, y))
+            .expect("Empty array.");
```
ed255, line 312: why not this?
```suggestion
        let not_args_equal = builder.not(args_equal);
        let lt_check_flag = builder.and(op_code_ok, not_args_equal);
```

src/frontend/mod.rs
```diff
@@ -259,6 +259,15 @@ impl MainPodBuilder {
                 let [array, index, value] = op.1.try_into().unwrap(); // TODO: Error handling
                 Operation(Native(ContainsFromEntries), vec![array, index, value], op.2)
             }
+            Native(GtEqFromEntries) => {
```
ed255, line 262: Can you add a new line to the function documentation summarizing these transformations?

ed255 (approved):
LGTM!

src/backends/plonky2/circuits/common.rs
```diff
@@ -63,6 +64,15 @@ impl ValueTarget {
     }
 }
 
+/// StatementArgTarget to ValueTarget coercion. Make sure to check
+/// that the arg is a value using the `statement_arg_is_value` method
+/// first!
+impl From<&StatementArgTarget> for ValueTarget {
+    fn from(st_arg: &StatementArgTarget) -> Self {
```
ed255, line 71: This seems a bit dangerous.
What about adding a method to `StatementArgTarget` that does this instead of using the `From` trait?
I fear misusing this if we keep it like this.

ed255 (approved):
The new changes LGTM!

PR #217, opened 2025-05-01
feat: implement HashOf statement and operation

ax0:
Resolves #212.
Depends on #216.

Issue #213, opened 2025-04-25
Implement the gadgets for all the native operations


ed255:
In the MainPod Circuit implement the constraints to verify all the native operations and integrate them
- [x] None 0
- [x] NewEntry 1
- [x] Copy 2
- [x] EqFromEntries 3
- [x] LtFromEntries 5 (in the spec we have GtFromEntries)
- [x] NotContainsFromEntries 11
- [x] NotEqFromEntries 4
- [x] LtEqFromEntries 6
- [x] TransitiveEq 7
- [x] LtToNeq 9
- [x] ContainsFromEntries 10
- [ ] RenameContains? (Do we need this?)
- [ ] SumOf 13
- [ ] ProductOf 14
- [ ] MaxOf 15
- [x] HashOf 16

Issue #210, opened 2025-04-25
[meta] POD1 & GPC feature parity circuit


ed255:
This issue collects the list of tasks required to have a POD2 circuit that has feature parity with POD1 and GPC.
The target performance of this circuit is ~5 seconds on a phone via WASM.  We’d aim for a 2^13 circuit target to reach the 5 seconds.
More details:
- Doesn’t need recursion
- Needs to verify ~5 signatures per circuit.
- Doesn’t need custom predicates

ed255:
> One useful feature of POD1/GPC is "ownership", which allows the user to prove ownership of a keypair. If a POD has an entry whose value is a public key, the prover can prove ownership of the corresponding private key. For example, a group organizer issues a membership credential to a person, identified by their public key. This credential can only be used when accompanied by a proof that the bearer of the credential is also the "owner" of it. I don't think that POD2 has an equivalent feature yet (though perhaps `HashOf` could be used to prove knowledge of some secret, which might be close enough for the authentication use-case).

Could we use a SignedPod here?
If the user only needs to prove ownership of a secret key to a verifier, then the verifier can give a challenge to the user, the user builds a SignedPod with that challenge in a particular key (of the signed key-values) and sends that to the verifier.
If the user also needs to prove some extra statements, the user takes the previously defined SignedPod and uses it as input to a MainPod which copies the SignedPod statements and includes extra statements as necessary.

The credential can be bound to the SignedPod or the MainPod if necessary by adding an extra key-value with the credential.

The verifier challenge is very important here to make the SignedPod/MainPod not reusable.

PR #222, opened 2025-05-05
feat(backend): implement more ops

ax0:
Towards #210 and #213.
Depends on #217.

### DIFF
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -974,10 +974,7 @@ mod tests {
     }
 
     #[test]
-    fn test_operation_verify() -> Result<()> {
-        let params = Params::default();
-
-        // None
+    fn test_operation_verify_none() -> Result<()> {
         let st: mainpod::Statement = Statement::None.into();
         let op = mainpod::Operation(
             OperationType::Native(NativeOperation::None),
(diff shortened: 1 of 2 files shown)


src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -360,6 +473,41 @@ impl OperationVerifyGadget {
...
+        let arg2_key = resolved_op_args[0].args[1].clone();
+
+        let expected_st_args: Vec<_> = [arg1_key, arg2_key]
+            .into_iter()
+            .chain(std::iter::repeat_with(|| StatementArgTarget::none(builder)))
+            .take(self.params.max_statement_args)
+            .flat_map(|arg| arg.elements)
+            .collect();
+
+        let st_args_ok = builder.is_equal_slice(
```
ed255, line 500: This is a nitpick but since this check only considers 1 possible statement code, why not check the expected statement like this?
```rust
        let expected_statement = StatementTarget::new_native(
            builder,
            &self.params,
            NativePredicate::NotEqual,
            &[arg1_key1.clone(), arg2_key2.clone()],
        );
        let st_ok = builder.is_equal_flattenable(st, &expected_statement);
```
And get rid of the check 
```rust
let st_code_ok = st.has_native_type(builder, &self.params, NativePredicate::NotEqual);
```

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -133,6 +144,54 @@ impl OperationVerifyGadget {
...
+        resolved_merkle_claim: MerkleClaimTarget,
+        resolved_op_args: &[StatementTarget],
+    ) -> BoolTarget {
+        let op_code_ok = op.has_native_type(builder, NativeOperation::ContainsFromEntries);
+
+        let arg_types_ok = self.first_n_args_are_valueofs(builder, 3, resolved_op_args);
+
+        let merkle_root_value = resolved_op_args[0].args[1].as_value();
+        let key_value = resolved_op_args[1].args[1].as_value();
+        let value_value = resolved_op_args[2].args[1].as_value();
```
ed255, line 161: What do you think about merging these two operations into one?  Imagine this method
```rust
impl OperationVerifyGadget {
    fn first_n_args_as_valueofs<const N: usize>(
        &self,
        builder: &mut CircuitBuilder<F, D>,
        resolved_op_args: &[StatementTarget],
    ) -> (BoolTarget, [ValueTarget; N]) {
        // ...
    }
}
```
then you can call it like this:
```rust
        let (arg_types_ok, [merkle_root_value, key_value, value_value]) =
            self.first_n_args_as_valueofs::<3>(builder, resolved_op_args);
```

ed255 (approved):
LGTM! I've left a few suggestions for future readability improvements

Issue #180, opened 2025-04-04
Implement custom deductions circuits


arnaucube:
Depens on #171 .

ed255:
Subtasks:
- [ ] Implement all the targets for custom predicates
- [ ] Implement the custom predicate batch id calculation
- [ ] Load N batches in the MainPod circuit and calculate their id
- [ ] Verify the custom operation

PR #223, opened 2025-05-06
add target types for custom predicates

ed255:
Towards #180

### DIFF
--- a/src/backends/plonky2/circuits/common.rs
+++ b/src/backends/plonky2/circuits/common.rs
@@ -146,9 +146,8 @@ impl StatementTarget {
     ) -> Self {
         // if native_predicate is const then NativePredicate -> NativePredicateTarget
         // else just use as is
-        let native_predicate = native_predicate.build(builder, params);
         Self {
-            predicate: PredicateTarget::new_native(builder, &native_predicate),
+            predicate: PredicateTarget::new_native(builder, params, native_predicate),
             args: args
                 .iter()
                 .cloned()
@@ -183,8 +182,7 @@ impl StatementTarget {
         params: &Params,
         t: NativePredicate,
     ) -> BoolTarget {
-        let expected_t = NativePredicateTarget::constant(builder, params, t);
-        let expected_predicate = PredicateTarget::new_native(builder, &expected_t);
+        let expected_predicate = PredicateTarget::new_native(builder, params, &t);
         builder.is_equal_flattenable(&self.predicate, &expected_predicate)
     }
 }
@@ -265,10 +263,11 @@ pub struct PredicateTarget {
 impl PredicateTarget {
     pub fn new_native(
         builder: &mut CircuitBuilder<F, D>,
-        native_predicate: &NativePredicateTarget,
+        params: &Params,
+        native_predicate: impl Build<NativePredicateTarget>,
     ) -> Self {
         let prefix = builder.constant(F::from_canonical_usize(1));
-        let id = native_predicate.0;
+        let id = native_predicate.build(builder, params).0;
         let zero = builder.zero();
         Self {
             elements: [prefix, id, zero, zero, zero, zero],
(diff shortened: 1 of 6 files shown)


src/middleware/custom.rs
```diff
@@ -334,65 +336,6 @@ impl CustomPredicateRef {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, JsonSchema)]
```
ed255, line 337: I moved all these lines to `src/middleware/operation.rs` because they are not specific to custom predicate

src/backends/plonky2/circuits/common.rs
```diff
@@ -212,11 +229,159 @@ impl OperationTarget {
...
+}
+
+impl StatementTmplArgTarget {
+    pub fn as_none(&self, builder: &mut CircuitBuilder<F, D>) -> BoolTarget {
+        let none_prefix = builder.constant(F::from_canonical_usize(0));
+        builder.is_equal(self.elements[0], none_prefix)
+    }
+    pub fn as_literal(&self, builder: &mut CircuitBuilder<F, D>) -> (BoolTarget, ValueTarget) {
+        let none_prefix = builder.constant(F::from_canonical_usize(1));
+        let case_ok = builder.is_equal(self.elements[0], none_prefix);
```
ax0, line 332: I assume this should be:
```suggestion
        let literal_prefix = builder.constant(F::from_canonical_usize(1));
        let case_ok = builder.is_equal(self.elements[0], literal_prefix);
```

src/backends/plonky2/circuits/common.rs
```diff
@@ -212,11 +229,159 @@ impl OperationTarget {
...
+        let value = ValueTarget::from_slice(&self.elements[1..5]);
+        (case_ok, value)
+    }
+    pub fn as_key(
+        &self,
+        builder: &mut CircuitBuilder<F, D>,
+    ) -> (BoolTarget, Target, KeyOrWildcardTarget) {
+        let none_prefix = builder.constant(F::from_canonical_usize(2));
+        let case_ok = builder.is_equal(self.elements[0], none_prefix);
```
ax0, line 341: Similarly here.

src/backends/plonky2/circuits/common.rs
```diff
@@ -212,11 +229,159 @@ impl OperationTarget {
...
+        let id_wildcard_index = self.elements[1];
+        let value_key_or_wildcard = KeyOrWildcardTarget::from_slice(&self.elements[5..9]);
+        (case_ok, id_wildcard_index, value_key_or_wildcard)
+    }
+    pub fn as_wildcard_literal(&self, builder: &mut CircuitBuilder<F, D>) -> (BoolTarget, Target) {
+        let none_prefix = builder.constant(F::from_canonical_usize(3));
+        let case_ok = builder.is_equal(self.elements[0], none_prefix);
```
ax0, line 348: Similarly here.

src/backends/plonky2/circuits/common.rs
```diff
@@ -212,11 +229,159 @@ impl OperationTarget {
...
+    elements: [Target; Params::predicate_size()],
+}
+
+impl PredicateTarget {
+    pub fn new_native(
+        builder: &mut CircuitBuilder<F, D>,
+        params: &Params,
+        native_predicate: impl Build<NativePredicateTarget>,
+    ) -> Self {
+        let prefix = builder.constant(F::from_canonical_usize(1));
```
ax0, line 269: Thoughts on having these (and the template arg) prefixes as named constants?

ax0 (approved):
Nice!

src/backends/plonky2/circuits/common.rs
ed255: thanks for the suggestion! I've introduced enums for the prefixes :)

src/backends/plonky2/circuits/common.rs
ed255: I ended up using enums for the prefixes (like constants).  And chose to name the variables just `prefix` because from the enum case it's already clear which prefix it is.

Issue #224, opened 2025-05-06
Custom predicates aren't aware of syntactic sugar


robknight:
In https://github.com/0xPARC/pod2/pull/216 (issue #214) the Gt and GtEq statements were eliminated, and the Gt and GtEq native predicates became syntactic sugar. It's still possible to use the predicate in an operation, but the result will be a Lt or LtEq statement, with the arguments transposed. This makes logical sense.

However, this change breaks the use of `Gt` and `GtEq` predicates in statement templates. Since `Gt` and `GtEq` statements no longer appear, it is not possible for a `Gt` or `GtEq` statement template to match on anything.

I've added a test demonstrating this in a [new branch](https://github.com/0xPARC/pod2/tree/statement_template_sugar). This is the relevant code:

https://github.com/0xPARC/pod2/blob/06bd9d5f45795cb0316e5161babe0c46142d4b53/src/frontend/custom.rs#L243-L305

The first test defines a custom predicate which matches on an `Lt` statement, and then builds a MainPod containing an `Lt` statement and the custom statement. This builds successfully.

The second test does exactly the same thing, but with a `Gt` statement. This does _not_ build successfully. The `Gt` statement is valid, but the custom predicate can't match on it.

My immediate thought is that either:

1) Statement templates should only be able to use native predicates which are supported in `middleware::Statement`, or
2) Statement template matching should be "sugar-aware", meaning that a statement template for `Gt(A, B)` should match on the statement `Lt(B, A)`

But these might be naive solutions, and there may be a better option.

ed255:
What about desugaring in the `CustomPredicateBatchBuilder`?
So you'd build a CustomPredicateBatch with it.  You'd pass `StatementTmplBuilder` that use `Gt(A, B)` but the builder internally will change the statement template to become `Lt(B, A)`.

This means following this proposal:
> 1) Statement templates should only be able to use native predicates which are supported in `middleware::Statement`, or

But with the nice thing that we can still use syntax sugar when bulding custom predicates via the builder.  So the `StatementTmplBuilder` is allowed to use `Gt` but the resulting `StatementTmpl`s in the custom predicates will not have syntax sugar.

Do you think this solves the issue?  Do you see any inconvenience from this approach?

robknight:
In the playground code I parse a string like `Gt(?A, ?B)` directly into a StatementTmpl, without using StatementTmplBuilder, but the same principle could work there. I think this is probably the best solution.

[...]

PR #225, opened 2025-05-07
Podlog language v1

robknight:
This adds a simple DSL for specifying custom predicates and proof requests. It is similar to the language we have been using informally in our docs. The code includes tests, and should be capable of supporting all of our current predicates, with the exception of `SetContains`/`SetNotContains`. A language reference is included below.

## Example
A proof request for ZuKYC:

```podlog
REQUEST(
    NotContains(?sanctions["sanctionList"], ?gov["idNumber"]) 
    Lt(?gov["dateOfBirth"], ?SELF_HOLDER_18Y["const_18y"])      
    Equal(?pay["startDate"], ?SELF_HOLDER_1Y["const_1y"])         
    Equal(?gov["socialSecurityNumber"], ?pay["socialSecurityNumber"]) 
    ValueOf(?SELF_HOLDER_18Y["const_18y"], 1169909388)               
    ValueOf(?SELF_HOLDER_1Y["const_1y"], 1706367566)                  
)
```

Ethdos custom predicates:
```podlog
eth_friend(src_ori, src_key, dst_ori, dst_key, private: attestation_pod) = AND(
    ValueOf(?attestation_pod["_type"], 1)
    Equal(?attestation_pod["_signer"], ?src_ori[?src_key])
    Equal(?attestation_pod["attestation"], ?dst_ori[?dst_key])
)

eth_dos_distance_base(src_ori, src_key, dst_ori, dst_key, distance_ori, distance_key) = AND(
    Equal(?src_ori[?src_key], ?dst_ori[?dst_key])
    ValueOf(?distance_ori[?distance_key], 0)
)

eth_dos_distance_ind(src_ori, src_key, dst_ori, dst_key, distance_ori, distance_key, private: one_ori, one_key, shorter_distance_ori, shorter_distance_key, intermed_ori, intermed_key) = AND(
    eth_dos_distance(?src_ori, ?src_key, ?intermed_ori, ?intermed_key, ?shorter_distance_ori, ?shorter_distance_key)
    ValueOf(?one_ori[?one_key], 1)
    SumOf(?distance_ori[?distance_key], ?shorter_distance_ori[?shorter_distance_key], ?one_ori[?one_key])
    eth_friend(?intermed_ori, ?intermed_key, ?dst_ori, ?dst_key)
)

eth_dos_distance(src_ori, src_key, dst_ori, dst_key, distance_ori, distance_key) = OR(
    eth_dos_distance_base(?src_ori, ?src_key, ?dst_ori, ?dst_key, ?distance_ori, ?distance_key)
    eth_dos_distance_ind(?src_ori, ?src_key, ?dst_ori, ?dst_key, ?distance_ori, ?distance_key)
)
```
  
## Code structure

The grammar is implemented using [Pest](https://docs.rs/pest/latest/pest/), which is able to derive a basic parser implementation from a grammar file. We have tests for this in `parser.rs`. The parsed input is then turned into an `AST` by the code in `ast.rs`. This closely mirrors the internal POD2 data structures, but may diverge in some ways (for example, by allowing certain pieces of syntactic sugar). Finally the AST is turned into POD2 internal data structures in `processor.rs`. There are end-to-end tests in `mod.rs`.

## Outstanding issues

Currently `SetContains` and `SetNotContains` are unsupported, as de-sugaring these to lower-level templates might require some changes in `MainPodBuilder` or elsewhere. See https://github.com/0xPARC/pod2/issues/224

## Next steps

If we're happy with the grammar here, I will update the code examples in the book to use this syntax.

## Possible future features

A few features have been considered and are currently out-of-scope, but could be included in future:

### Multiple custom predicate batches

For a given Podlog document, all custom predicates will be included in the same batch. We might want to give the user more control by allowing them to specify separate batches, e.g.:

```podlog
batch "eth_friend" {
  eth_friend(A, B) = AND ( ... )
}

batch "eth_dos" {
  eth_dos_distance(X, Y) = AND ( eth_friend(X, Y) )
}
```

In this case, the call to `eth_friend` would be made via a `CustomPredicateRef` rather than a `BatchSelf`.

### Import syntax

It might be helpful to be able to import custom predicates from different files. For example:

```podlog
eth_friend = import("path/to/eth_friend.podlog").eth_friend
```

Given that each custom predicate has a unique hash, we could also imagine a content-addressable store of custom predicates:

```podlog
eth_friend = import (0x123456abcdef)
```

# Language reference

<details>

<summary>Language reference</summary>

# Podlog Language Syntax

Podlog is a declarative language used for defining custom proof logic and requests within the POD2 ecosystem. It allows for the specification of statement templates that can be evaluated by a prover.

## Table of Contents

1.  [Overview](#overview)
2.  [Comments](#comments)
3.  [Identifiers](#identifiers)
4.  [Variables](#variables)
5.  [Literals](#literals)
    *   [Integer](#integer)
    *   [String](#string)
    *   [Raw (Bytes)](#raw-bytes)
    *   [Boolean](#boolean)
    *   [Array](#array)
    *   [Set](#set)
    *   [Dictionary](#dictionary)
6.  [Anchored Keys](#anchored-keys)
7.  [Top-Level Definitions](#top-level-definitions)
    *   [Custom Predicate Definition](#custom-predicate-definition)
    *   [Request Definition](#request-definition)
8.  [Statements](#statements)
    *   [Native Predicate Calls](#native-predicate-calls)
    *   [Custom Predicate Calls](#custom-predicate-calls)

## Overview

A Podlog document consists of zero or more top-level definitions, which can be either custom predicate definitions or a single request definition. The language is designed to be human-readable while providing precise instructions for the POD2 prover.

## Comments

Single-line comments begin with `//`. All text from `//` to the end of the line is ignored.

```podlog
// This is a comment.
is_valid(User) = AND( // This is an inline comment
    Equal(?User["status"], "active")
)
```

## Identifiers

Identifiers are used for naming custom predicates and for the non-prefixed part of variable names.
*   Must start with an uppercase or lowercase letter (A-Z, a-z) or an underscore (`_`).
*   Subsequent characters can be letters, numbers (0-9), or underscores.
*   Keywords like `private`, `REQUEST`, `AND`, `OR`, and native predicate names (e.g., `Equal`, `ValueOf`) cannot be used as predicate identifiers.
*   The keyword `private` cannot be used as an identifier (e.g., for variable names).

Examples:
`my_predicate`
`_internal_check`
`isValidUser1`

Invalid:
`123starts_with_digit`
`has-hyphen`
`private` (as a predicate or variable name component)
`Equal` (as a predicate name)

## Variables

Variables represent placeholders for values that the prover will attempt to bind.
*   Variable names always start with a question mark (`?`).
*   The part following the `?` must adhere to the rules for [Identifiers](#identifiers).

Examples:
`?UserPod`
`?publicKey`
`?_tempValue`

## Literals

Literals represent fixed values.

### Integer

A sequence of digits, optionally prefixed with a `-` for negative numbers.
Maps to `middleware::Value::Int`.

Examples: `123`, `-45`, `0`

### String

Characters enclosed in double quotes (`"`).
Standard JSON-like escaping rules apply for special characters:
*   `\"` for double quote
*   `\\` for backslash
*   `/` for forward slash (optional, `/` can be unescaped)
*   `\b` for backspace
*   `\f` for form feed
*   `\n` for newline
*   `\r` for carriage return
*   `\t` for tab
*   `\uXXXX` for a Unicode character specified by 4 hexadecimal digits.
Maps to `middleware::Value::String`.

Examples: `"hello"`, `"user:name"`, `"escaped \\\" quote"`, `"\u0041"`

### Raw (Bytes)

A hexadecimal string prefixed with `0x`. Must contain an even number of hex digits (0-9, a-f, A-F) and at least two digits.
Maps to `middleware::Value::Raw`.

Examples: `0xdeadbeef`, `0x00`, `0x0102030405060708090a0b0c0d0e0f10`

Invalid: `0xabc` (odd length), `0x` (empty), `0xGG` (invalid hex chars)

### Boolean

The keywords `true` or `false`.
Maps to `middleware::Value::Bool`.

Examples: `true`, `false`

### Array

A comma-separated list of literals enclosed in square brackets (`[]`). Elements can be of mixed types.
Maps to `middleware::Value::Array`.

Examples: `[]`, `[1, "two", true]`, `[ [10, 20], #["a", "b"] ]`

### Set

A comma-separated list of literals enclosed in square brackets prefixed with a hash (`#[]`). Elements can be of mixed types. The processor will validate uniqueness if required by the underlying middleware.
Maps to `middleware::Value::Set`.

Examples: `#[]`, `#[1, 2, "apple"]`, `#[ 0x01, 0x02 ]`

### Dictionary

A comma-separated list of `string_key: literal_value` pairs enclosed in curly braces (`{}`). Keys must be string literals.
Maps to `middleware::Value::Dictionary`.

Examples: `{}`, `{ "name": "Alice", "age": 30 }`, `{ "data": [1,2,3], "metadata": { "version": "1.0" } }`

## Anchored Keys

Anchored keys are used to refer to values within a POD (represented by a variable).
Syntax: `?PodVariable["key_literal"]` or `?PodVariable[?KeyVariable]`

*   `?PodVariable`: A [Variable](#variables) representing a POD.
*   `"key_literal"`: A [String Literal](#string) representing a fixed key.
*   `?KeyVariable`: A [Variable](#variables) representing a key to be resolved at runtime.

Examples:
`?UserPod["name"]`
`?DataPod[?fieldName]`

## Top-Level Definitions

A Podlog document can contain multiple custom predicate definitions and at most one request definition.

### Custom Predicate Definition

Defines a reusable piece of logic.

Syntax:
`predicate_name(public_arg1, public_arg2, ..., private: private_arg1, private_arg2, ...) = CONJUNCTION_TYPE( statement1 statement2 ... )`

*   `predicate_name`: An [Identifier](#identifiers) for the predicate.
*   **Arguments**:
    *   A comma-separated list of public argument names (which are identifiers).
    *   Optionally, followed by `, private:` and then a comma-separated list of private argument names (identifiers).
    *   A predicate can have only public arguments, only private arguments (e.g., `pred(private: PrivA) = ...`), or no arguments (e.g., `pred() = ...`).
    *   Argument names become [Variables](#variables) (e.g., `?public_arg1`) within the predicate's scope.
*   `CONJUNCTION_TYPE`: Either `AND` or `OR`.
    *   `AND`: All statements in the body must be true.
    *   `OR`: At least one statement in the body must be true.
*   `statement_list`: One or more [Statements](#statements) enclosed in parentheses.

Examples:
```podlog
// Predicate with only public arguments
is_equal_on_key(PodA, PodB, KeyName) = AND(
    Equal(?PodA[?KeyName], ?PodB[?KeyName])
)

// Predicate with public and private arguments
check_attestation(SrcOrigin, DstOrigin, private: AttestationPod) = AND(
    ValueOf(?AttestationPod["_type"], "MockSigned")
    Equal(?AttestationPod["_signer"], ?SrcOrigin["publicKey"])
    Equal(?AttestationPod["attestation"], ?DstOrigin["id"])
)
```

### Request Definition

Defines the main set of statements to be proven by the prover. There can be at most one `REQUEST` block in a document.

Syntax:
`REQUEST( statement1 statement2 ... )`

*   `statement_list`: Optional. If present, one or more [Statements](#statements) enclosed in parentheses. Variables used here are implicitly defined and their bindings are sought by the prover.

Example:
```podlog
REQUEST(
    is_adult(?User) // Calling a custom predicate
    Equal(?User["countryCode"], ?X["country"])
    NotContains(?SanctionList["names"], ?User["fullName"])
)
```

## Statements

Statements are the core assertions or operations within custom predicate definitions and request blocks.

### Native Predicate Calls

These calls invoke built-in operations provided by the POD2 middleware. Arguments to native predicates are typically [Anchored Keys](#anchored-keys), except for `ValueOf` which takes an Anchored Key and a [Literal](#literals).

**Core Predicates (Directly map to middleware operations):**

*   `ValueOf(?PodVar["key"], LiteralValue)`
    *   Asserts that the value at `?PodVar["key"]` is equal to `LiteralValue`.
    *   Example: `ValueOf(?Config["port"], 8080)`

*   `Equal(?PodVar1["key1"], ?PodVar2["key2"])`
    *   Asserts that the values at the two anchored keys are equal.
    *   Example: `Equal(?User["id"], ?Order["userId"])`

*   `NotEqual(?PodVar1["key1"], ?PodVar2["key2"])`
    *   Asserts that the values at the two anchored keys are not equal.
    *   Example: `NotEqual(?Account["status"], "suspended")`

*   `Lt(?PodVar1["key1"], ?PodVar2["key2"])`
    *   Asserts `value(?PodVar1["key1"]) < value(?PodVar2["key2"])`.
    *   Example: `Lt(?Event["timestamp"], ?CutoffTime["value"])`

*   `LtEq(?PodVar1["key1"], ?PodVar2["key2"])`
    *   Asserts `value(?PodVar1["key1"]) <= value(?PodVar2["key2"])`.
    *   Example: `LtEq(?User["age"], ?MaxAge["limit"])`

*   `Contains(?PodVar[?Container], ?PodVar2[?Key], ?PodVar2[?Value])`
    *   A general predicate asserting that a container (e.g., dictionary, array, set) contains `?Value` at a location specified by `?Key`. In most cases, users should use `DictContains`, `SetContains`, or `ArrayContains` based on the expected type of the container.
    *   Example: `Contains(?Data["record"], ?User["nameKey"], ?User["expectedName"])`

*   `NotContains(?PodVar[?Container], ?PodVar2[?Key])`
    *   A general predicate asserting that `?Key` is not present in `?Container`.
    *   Example: `NotContains(?Banlist["users"], ?User["id"])`

*   `SumOf(?Target["keyT"], ?Source1["keyS1"], ?Source2["keyS2"])`
    *   Asserts that the value at `?Target["keyT"]` equals the sum of the values at `?Source1["keyS1"]` and `?Source2["keyS2"]`.
    *   Example: `SumOf(?Totals["all"], ?SubTotals["A"], ?SubTotals["B"])`

*   `ProductOf(?Target["keyT"], ?Source1["keyS1"], ?Source2["keyS2"])`
    *   Asserts that the value at `?Target["keyT"]` equals the product of the values at `?Source1["keyS1"]` and `?Source2["keyS2"]`.
    *   Example: `ProductOf(?Area["rect"], ?Dimensions["width"], ?Dimensions["height"])`

*   `MaxOf(?Target["keyT"], ?Source1["keyS1"], ?Source2["keyS2"])`
    *   Asserts that the value at `?Target["keyT"]` is equal to either of the values at `?Source1["keyS1"]` or `?Source2["keyS2"]`, whichever is higher.
    *   Example: `MaxOf(?Score["high"], ?Player1["score"], ?Player2["score"])`

*   `HashOf(?Target["keyT"], ?Source1["keyS1"], ?Source2["keyS2"])`
    *   Asserts `value(?Target["keyT"])` is the hash of `value(?Source1["keyS1"])` and `value(?Source2["keyS2"])`.
    *   Example: `HashOf(?User["computedId"], ?User["email"], ?User["salt"])`

**Syntactic Sugar Predicates (Translated by the processor):**

These provide more convenient syntax for common operations and are translated into the core middleware predicates.

*   `Gt(?PodVar1["key1"], ?PodVar2["key2"])`
    *   Asserts `value(?PodVar1["key1"]) > value(?PodVar2["key2"])`.
    *   Translated to: `Lt(?PodVar2["key2"], ?PodVar1["key1"])` (arguments are swapped).
    *   Example: `Gt(?User["score"], 1000)`

*   `GtEq(?PodVar1["key1"], ?PodVar2["key2"])`
    *   Asserts `value(?PodVar1["key1"]) >= value(?PodVar2["key2"])`.
    *   Translated to: `LtEq(?PodVar2["key2"], ?PodVar1["key1"])` (arguments are swapped).
    *   Example: `GtEq(?Balance["amount"], ?Minimum["required"])`

*   `DictContains(?DictVar["dict_key"], ?KeyToFind["key_name"], ?ExpectedValue["val_name"])`
    *   Asserts `value(?DictVar["dict_key"][?KeyToFind["key_name"]]) == value(?ExpectedValue["val_name"])`.
    *   Translated to: `Contains(?DictVar["dict_key"], ?KeyToFind["key_name"], ?ExpectedValue["val_name"])` (or similar, depending on exact mapping to generic Contains).
    *   Example: `DictContains(?UserPod["attributes"], ?_KeyNamePod["name"], ?_ExpectedValuePod["admin"])`
        *   Where `?_KeyNamePod["name"]` might resolve to a literal key like "role" and `?_ExpectedValuePod["admin"]` to "admin".

*   `DictNotContains(?DictVar["dict_key"], ?KeyToFind["key_name"])`
    *   Asserts that `?KeyToFind["key_name"]` is not a key in `value(?DictVar["dict_key"])`.
    *   Translated to: `NotContains(?DictVar["dict_key"], ?KeyToFind["key_name"])`.
    *   Example: `DictNotContains(?UserPod["flags"], ?_FlagNamePod["is_banned"])`

*   `ArrayContains(?ArrayVar["array_key"], ?IndexVar["idx_name"], ?ExpectedValue["val_name"])`
    *   Asserts `value(?ArrayVar["array_key"][?IndexVar["idx_name"]]) == value(?ExpectedValue["val_name"])`.
    *   Translated to: `Contains(?ArrayVar["array_key"], ?IndexVar["idx_name"], ?ExpectedValue["val_name"])`.
    *   Example: `ArrayContains(?LogPod["entries"], ?_IndexPod["target_idx"], ?_ValuePod["error_code_X"])`

*(Note: `SetContains` and `SetNotContains` are currently missing, pending an implementation for de-sugaring.)*

### Custom Predicate Calls

Calls a previously defined custom predicate.

Syntax:
`predicate_name(arg1, arg2, ...)`

*   `predicate_name`: The [Identifier](#identifiers) of the custom predicate to call.
*   `arg_list`: A comma-separated list of arguments. Arguments can be [Variables](#variables) or [Literals](#literals). The number and type of arguments must match the public argument definition of the called predicate.

Example:
```podlog
has_permission(User, Permission, private: Roles) = AND(
    // ... logic using ?User, ?Permission, ?Roles
)

REQUEST(
    has_permission(?CurrentUser, "edit_document")
    ValueOf(?SystemConfig["audit_level"], 3)
)
```
</details>


ed255 (changes requested):
I've finished reviewing this PR and I'm marking request changes because there are significant changes to the approach of this PR that I want to discuss.

src/lang/grammar.pest
```diff
@@ -0,0 +1,158 @@
...
+// -- End New String Literal Parsing Strategy --
+
+// Container Literals (recursive definition using literal_value)
+literal_array = { "[" ~ (literal_value ~ ("," ~ literal_value)*)? ~ "]" }
+literal_set = { "#[" ~ (literal_value ~ ("," ~ literal_value)*)? ~ "]" }
+literal_dict = { "{" ~ (dict_pair ~ ("," ~ dict_pair)*)? ~ "}" }
+dict_pair = { literal_string ~ ":" ~ literal_value }
+
+
+// Rules for specific native predicate calls
```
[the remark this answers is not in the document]
robknight: Since we need to build the AST anyway, I agree that it makes more sense to have this checking done when building the AST than when building the grammar. It will be annoying to have to change the grammar every time we want to add a new statement type.

src/lang/processor.rs
```diff
@@ -0,0 +1,1347 @@
...
+                    .ok_or_else(|| ProcessorError::Semantic {
+                        message: format!("Invalid wildcard format for BuilderArg: {}", full_name),
+                        span: Some(get_span(arg_content_pair)),
+                    })?;
+            if name_only.is_empty() {
+                return Err(ProcessorError::Semantic {
+                    message: "Wildcard name cannot be empty after '?' for BuilderArg".to_string(),
+                    span: Some(get_span(arg_content_pair)),
+                });
+            }
```
ed255, line 248: Here you're doing two checks:
Make sure there's a `?` prefix and making sure the identifier is not empty.  But the grammar already enforces this!
So by the definition of the grammar, these two errors will never occur. 
More over, if they were to happen, it would be a bug in the code because either pest is not returning a parsing error, or the grammar is not well defined.  For those cases I strongly suggest that we panic instead, and we can do this cleanly with `expect`.

Please take a look at this commit where I've made some changes to simplify this following this idea:
https://github.com/0xPARC/pod2/commit/b8063224a46fb2dfe91608b6588efa52c0a7bef2

I think there are a few more places where this changes can be applied.  The only place where we need to do some checking and do proper error reporting is when handling native predicates because their argument types are not validated by the grammar.

But any time we find a rule defined in the grammar, we know that the tokens defined by the grammar will be there.  For example, every time we find the `Rule::anchored_key` we know for sure that we'll encounter 2 Pairs.  The first one will either be `self_keyword` or `wildcard`; and the second one will either be `wildcard` or `literal_string`.  Any other case is not expected to happen on any kind of user input.

src/lang/processor.rs
```diff
@@ -0,0 +1,1347 @@
...
+                }
+            }
+        } else {
+            result.push(c);
+        }
+    }
+    Ok(result)
+}
+
+fn parse_hex_str_to_raw_value(hex_str: &str) -> Result<middleware::RawValue, ProcessorError> {
```
ed255, line 964: I just realized we already had this functionality here: https://github.com/0xPARC/pod2/blob/d3fef8392e110121e0b4284995d696cc873a64f9/src/middleware/basetypes.rs#L209-L218

It's used to deserialize a `Hash` from a hex string, and internally it creates the 4 field elements, which is the same we need to get a RawValue.  Then you can do `RawValue::from(h: Hash)`.

The encoding is different though, and it doesn't use a `0x` prefix.
The encoding uses little endian, and taking a little endian hex encoded value and adding a `0x` prefix is confusing from my point of view.

src/lang/processor.rs
```diff
@@ -0,0 +1,1181 @@
...
+                        first_span: first_request_span,
+                        second_span: Some(get_span(&pair)),
+                    });
+                }
+                first_request_span = Some(get_span(&pair));
+                ctx.request_pair = Some(pair);
+            }
+            Rule::EOI => break,
+            Rule::COMMENT | Rule::WHITESPACE => {}
+            _ => {
```
ed255, line 156: We call `first_pass` after checking we're in a `document`, and by definition there are no more rules in `document`, so this should be `unreachable` instead of error

src/lang/processor.rs
```diff
@@ -0,0 +1,1181 @@
...
+            let mut inner_ak_pairs = arg_content_pair.clone().into_inner();
+            let pod_id_pair = inner_ak_pairs.next().unwrap();
+
+            let pod_self_or_wc_str = match pod_id_pair.as_rule() {
+                Rule::wildcard => {
+                    let name = pod_id_pair.as_str().strip_prefix("?").unwrap();
+                    SelfOrWildcardStr::Wildcard(name.to_string())
+                }
+                Rule::self_keyword => SelfOrWildcardStr::SELF,
+                _ => {
```
ed255, line 232: the `anchored_key` rule by grammar definition can only `wildcard` or `self_keyword` for the first pair, so this should be `unreachable`.

src/lang/processor.rs
```diff
@@ -0,0 +1,1181 @@
...
+                            name: arg_name,
+                            span: Some(get_span(&arg_ident_pair)),
+                        });
+                    }
+                    private_arg_strings.push(arg_name);
+                }
+            }
+            Rule::private_kw | Rule::COMMENT | Rule::WHITESPACE => {}
+            _ if arg_part_pair.as_str() == "," => {}
+            _ => {
```
ed255, line 456: The `arg_section` rule by grammar definition doesn't have more cases, so this should be `unreachable`.

src/lang/processor.rs
```diff
@@ -0,0 +1,1181 @@
...
+                    let val = process_literal_value(&val_pair, context_stmt_name)?;
+                    Ok((Key::new(key_str), val))
+                })
+                .collect();
+            let middleware_dict = middleware::containers::Dictionary::new(pairs?).map_err(|e| {
+                ProcessorError::Internal(format!("Failed to create Dictionary: {}", e))
+            })?;
+            Ok(Value::from(middleware_dict))
+        }
+        _ => Err(ProcessorError::RuleMismatch {
```
ed255, line 746: We call `process_literal_value` after checking that the pair rule is `literal_value` which doesn't have more cases by definition, so this should be `unreachable`.

ed255 (approved):
LGTM! Thanks for addressing all the requests!

PR #226, opened 2025-05-07
Desugar statement templates

### LINKED ISSUE #224: Custom predicates aren't aware of syntactic sugar

robknight:
Closes #224 

This adds desugaring of statement templates during the building of a custom predicate. `Gt` and `GtEq` are replaced by `Lt` and `LtEq` respectively, `DictContains`/`ArrayContains` are replaced by `Contains`, and `DictNotContains`/`SetNotContains` are replaced by `NotContains`.

~~`SetContains` is TODO because it's not immediately obvious how the third argument to `Contains` should be generated.~~ EDIT: `SetContains` is now implemented, including a change in the Set implementation: instead of building the Set Merkle tree from pairs of `value` and `EMPTY_VALUE`, we now duplicate the `value`. This makes the `Contains` statement derived from `SetContains` deterministic.

### DIFF
--- a/book/src/values.md
+++ b/book/src/values.md
@@ -40,9 +40,9 @@ The array, set and dictionary types are similar types. While all of them use [a 
 - **array**: the elements are placed at the value field of each leaf, and the key field is just the array index (integer)
     - `leaf.key=i` 
     - `leaf.value=original_value` 
-- **set**: the value field of the leaf is unused, and the key contains the hash of the element
-    -  `leaf.key=hash(original_value)`
-    - `leaf.value=0`
+- **set**: both the key and the value are set to the hash of the value.
+    - `leaf.key=hash(original_value)`
+    - `leaf.value=hash(original_value)`
 
 In the three types, the merkletree under the hood allows to prove inclusion & non-inclusion of the particular entry of the {dictionary/array/set} element.
 
--- a/src/frontend/custom.rs
+++ b/src/frontend/custom.rs
@@ -24,6 +24,7 @@ pub fn key(s: &str) -> KeyOrWildcardStr {
 }
 
 /// Builder Argument for the StatementTmplBuilder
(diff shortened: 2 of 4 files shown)


ed255 (approved):
LGTM!

ed255 (approved):
Could you update the set spec from this documentation file https://github.com/0xPARC/pod2/blob/main/book/src/values.md#dictionary-array-set
with the new construction?

Otherwise LGTM!

PR #233, opened 2025-05-13
Use id as public inputs in MainPod verify

### LINKED ISSUE #232: In MainPod store Proof instead of ProofWithPublicInputs
Currently `MainPod` stores the proof as `ProofWithPublicInputs`.
When we're verifying the proof we're missing the check that the public inputs are the correct ones!

Instead we should store a `Proof` and then generate the public inputs from data in the `MainPod` (basically the `id`).
Otherwise we're not checking that the id calculated from the public statements is the same one passed as public input to verify the proof!

ed255:
Resolve #232


ax0 (approved):
LGTM!

PR #234, opened 2025-05-13
Serialization for Plonky2 Signed and Main PODs

robknight:
This PR adds serialization for the plonky2 variants of Signed and Main PODs.

Both mock and plonky2 PODs have identical signed representations. In Signed PODs, the `signature` field is an opaque string, and in Main PODs the `proof` field is an opaque string. External applications should not rely on this string having any particular contents or structure.

See the `SerializedSignedPod` and `SerializedMainPod` types - these are intermediate types used to standardize the representation before serialization.

### DIFF
--- a/src/backends/plonky2/mainpod/mod.rs
+++ b/src/backends/plonky2/mainpod/mod.rs
@@ -446,12 +446,14 @@ impl PodProver for Prover {
     }
 }
 
+pub type MainPodProof = Proof<F, C, D>;
+
 #[derive(Clone, Debug)]
 pub struct MainPod {
     params: Params,
     id: PodId,
     public_statements: Vec<Statement>,
-    proof: Proof<F, C, D>,
+    proof: MainPodProof,
 }
 
 /// Convert a Statement into middleware::Statement and replace references to SELF by `self_id`.
@@ -497,6 +499,28 @@ impl MainPod {
         })
         .map_err(|e| Error::custom(format!("MainPod proof verification failure: {:?}", e)))
     }
+
+    pub fn proof(&self) -> MainPodProof {
+        self.proof.clone()
+    }
+
+    pub fn params(&self) -> &Params {
+        &self.params
+    }
+
+    pub fn new(
+        proof: MainPodProof,
+        public_statements: Vec<Statement>,
+        id: PodId,
+        params: Params,
+    ) -> Self {
+        Self {
+            params,
+            id,
+            public_statements,
+            proof,
+        }
+    }
 }
 
 impl Pod for MainPod {
@@ -515,10 +539,6 @@ impl Pod for MainPod {
             .cloned()
             .map(|statement| normalize_statement(&statement, self.id()))
             .collect()
-    }
-
-    fn serialized_proof(&self) -> String {
-        todo!()
     }
 }
 
--- a/src/backends/plonky2/mainpod/mod.rs
+++ b/src/backends/plonky2/mainpod/mod.rs
@@ -2,6 +2,7 @@ pub mod operation;
 pub mod statement;
 use std::{any::Any, sync::Arc};
 
+use base64::{prelude::BASE64_STANDARD, Engine};
 use itertools::Itertools;
 pub use operation::*;
 use plonky2::{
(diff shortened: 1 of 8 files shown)


src/backends/plonky2/mainpod/mod.rs
```diff
@@ -347,12 +347,14 @@ impl PodProver for Prover {
...
 
+pub type MainPodProof = ProofWithPublicInputs<F, C, D>;
+
 #[derive(Clone, Debug)]
 pub struct MainPod {
     params: Params,
     id: PodId,
     public_statements: Vec<Statement>,
-    proof: ProofWithPublicInputs<F, C, D>,
+    proof: MainPodProof,
```
robknight, line 357: This change is mostly so that it's possible to import `MainPodProof` from the front-end without having to import `F`, `C` and `D`, which feel like they are implementation details that the front-end should not know about. Happy to be corrected on this! Also I see that there's a merge conflict for this line, so possibly this type needs to change.

ed255: The alias looks good to me!
As for the conflict, it should be easy to resolve, you should update the alias to be:
```rust
pub type MainPodProof = Proof<F, C, D>;
```

src/frontend/serialization.rs
```diff
@@ -1,109 +1,276 @@
...
             entries: pod.kvs,
             proof: pod.pod.serialized_proof(),
-            pod_class: "Signed".to_string(),
-            pod_type: "Mock".to_string(),
+            pod_type: if (&*pod.pod as &dyn Any)
+                .downcast_ref::<Plonky2SignedPod>()
+                .is_some()
+            {
+                SignedPodType::Signed
+            } else {
```
ed255, line 77: I would suggest doing a downcast to `MockSignedPod` here in an `else if`, and have a final `else` with an `unreachable!()`, so that this code is robust against future new signed pod types.  Otherwise the MockSigned would be the fallback.
The same applies to `impl From<MainPod> for SerializedMainPod`

src/frontend/serialization.rs
```diff
@@ -1,109 +1,276 @@
...
+        params: params.clone(),
+    }
+    .eval(&mut builder)
+    .map_err(|e| Error::custom(format!("Failed to evaluate MainPodVerifyCircuit: {}", e)))?;
+
+    let data = builder.build::<C>();
+    Ok(data.common)
+}
+
+fn decode_proof(proof: &str, params: &Params) -> Result<MainPodProof, Error> {
```
ed255, line 166: currently the MainPod in the backend has a method called `serialized_proof`.  I think it would be natural for the deserialization of the proof to also live there, because it's the same logic but in reverse.

Currently `serialized_proof` is part of the `Pod` trait and the signature of the method works for all pod types; whereas the deserialization is a bit different between a signed pod and a main pod, so it would not be straight forward find a generic method for deserialization.  But actually when deserializing the proof we already know the pod type, so this method doesn't need to be under the trait.

What do you think about just moving the functions `get_common_data` and `decode_proof` to `src/backends/plonky2/mainpod/mod.rs` (perhaps as methods of `MainPod`)
And the function `decode_signature` to `src/backends/plonky2/signedpod.rs` (perhaps as method of `SignedPod`)?

ed255 (commented):
Overall looks good to me! I've left a few suggestions.  In particular I think changing encodings from hex to base64 will reduce the serialization sizes.  I only mentioned the MainPod proof in the comments but this would also apply to the SignedPod signature.  I just checked and the MockPod serialization is already using base64.

src/frontend/serialization.rs
robknight: Done

src/frontend/serialization.rs
robknight: Done

src/backends/plonky2/mainpod/mod.rs
```diff
@@ -395,6 +397,28 @@ impl MainPod {
...
+
+    pub fn proof(&self) -> MainPodProof {
+        self.proof.clone()
+    }
+
+    pub fn params(&self) -> &Params {
+        &self.params
+    }
+
+    pub fn new(
```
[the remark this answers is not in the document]
robknight: Done

ed255 (approved):
Thanks for applying the requests!
LGTM!

PR #235, opened 2025-05-14
Fix custom predicate circuits and add tests for them

ed255:
- Add a test of the EthDos example using the MainPod (passing)
- Add a small test using a simple custom predicate
- Update the examples for eth_dos so that we can select using mock types or non-mock types
- Update docs with new `NativePredicate::False`
- Fixes
  - When the circuit verifies a custom predicate with recursion, the statement template may use BatchSelf, but the operation argument will not.  So we need to replace instances of BatchSelf by the batch_id and index.  This change can be seen in the constraints and the witness generation.
  - The index from the custom predicate verification to the custom predicate batch table was calculated wrongly
  - In an operation with a custom predicate the number of arguments is dynamic.  By using the 0 index as padding we're resolving the first statement for padding operation arguments!  To resolve this I've changed the design such that the statement at index 0 is a constant Statement None, so that resolving index 0 gives a well-known None which matches the values used in padding operation arguments.
  - When finding the index of the custom predicate verification I was only checking the batch_id and index.  But we can prove several instances of the same custom predicate!  So I added the missing op arguments check.
  - The predicate for padding statement templates was swapped between conjunction and disjunction.
- Extra goodies: I've added a debugger to print any target value of a plonky2 circuit :smiley:


src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -826,6 +826,7 @@ impl CustomOperationVerifyGadget {
             all_st_eq.target,
             some_st_eq.target,
         ));
+        // let is_op_args_ok = builder._true();
```
ax0, line 829:
```suggestion
```

src/backends/plonky2/circuits/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    fn run_once(
+        &self,
+        witness: &PartitionWitness<F>,
+        _out_buffer: &mut GeneratedValues<F>,
+    ) -> anyhow::Result<()> {
+        let xs = witness.get_targets(&self.xs);
+
+        println!("debug: values of {}", self.name);
```
ax0, line 52: Nice!

ax0 (approved):
LGTM!

PR #237, opened 2025-05-19
add custom gate for extension field multiplication

dgulotta:
This reduces the number of gates required for a signature check from about 12000 to about 3400.

I have not tested recursive proofs with this gate.


dgulotta:
The new commit adds a gate for elliptic curve addition, which reduces the number of gates for a signature check to about 1500.  It also adds tests to verify that recursion is implemented correctly for the new gates.

ax0 (approved):

PR #240, opened 2025-05-21
allow SELF in st_tmpl

### LINKED ISSUE #239: Allow hardcoded SELF in anchored keys in statement templates
Currently anchored keys in statement templates are defined as pod_id being always a wildcard and the key being a literal (hardcoded) value or a wildcard.

If we allow the pod_id to be SELF or a wildcard we get the following benefits:
- The statement template will clearly distinguish anchored keys from external pods from anchored keys used to store local values.  For example when a signed pod with some particular keys is required, the pod_id would be wildcard, but to store intermediate results a SELF anchored key would be used.
  - This will simplify the work of the solver, because now it clearly knows which anchored keys it is supposed to create and which ones it is supposed to find in a data base.
- The number of arguments to custom predicates will get reduced, because for local variables now we only need to pass the key and not the pod_id.

Downsides:
- The circuit logic needs to handle an extra case when resolving statement templates.  This should be quite small

Other thoughts:
- This design change is significantly smaller than #229 and is orthogonal to it (both can be implemented)

## Example with EthDOS

Before:
```
eth_friend(src_ori, src_key, dst_ori, dst_key, private: attestation_pod) = AND(
    ValueOf(?attestation_pod["_type"], 1)
    Equal(?attestation_pod["_signer"], ?src_ori[?src_key])
    Equal(?attestation_pod["attestation"], ?dst_ori[?dst_key])
)

eth_dos_distance_base(src_ori, src_key, dst_ori, dst_key, distance_ori, distance_key) = AND(
    Equal(?src_ori[?src_key], ?dst_ori[?dst_key])
    ValueOf(?distance_ori[?distance_key], 0)
)

eth_dos_distance_ind(src_ori, src_key, dst_ori, dst_key, distance_ori, distance_key, private: one_ori, one_key, shorter_distance_ori, shorter_distance_key, intermed_ori, intermed_key) = AND(
    eth_dos_distance(?src_ori, ?src_key, ?intermed_ori, ?intermed_key, ?shorter_distance_ori, ?shorter_distance_key)
    ValueOf(?one_ori[?one_key], 1)
    SumOf(?distance_ori[?distance_key], ?shorter_distance_ori[?shorter_distance_key], ?one_ori[?one_key])
    eth_friend(?intermed_ori, ?intermed_key, ?dst_ori, ?dst_key)
)

eth_dos_distance(src_ori, src_key, dst_ori, dst_key, distance_ori, distance_key) = OR(
    eth_dos_distance_base(?src_ori, ?src_key, ?dst_ori, ?dst_key, ?distance_ori, ?distance_key)
    eth_dos_distance_ind(?src_ori, ?src_key, ?dst_ori, ?dst_key, ?distance_ori, ?distance_key)
)
```

After:
```
eth_friend(src_key, dst_key, private: attestation_pod) = AND(
    ValueOf(?attestation_pod["_type"], 1)
    Equal(?attestation_pod["_signer"], SELF[?src_key])
    Equal(?attestation_pod["attestation"], SELF[?dst_key])
)

eth_dos_distance_base(src_key, dst_key, distance_key) = AND(
    Equal(SELF[?src_key], SELF[?dst_key])
    ValueOf(SELF[?distance_key], 0)
)

eth_dos_distance_ind(src_key, dst_key, distance_key, private: one_key, shorter_distance_key, intermed_key) = AND(
    eth_dos_distance(?src_key, ?intermed_key, ?shorter_distance_key)
    ValueOf(SELF[?one_key], 1)
    SumOf(SELF[?distance_key], SELF[?shorter_distance_key], SELF[?one_key])
    eth_friend(?intermed_key, ?dst_key)
)

eth_dos_distance(src_key, dst_key, distance_key) = OR(
    eth_dos_distance_base(?src_key, ?dst_key, ?distance_key)
    eth_dos_distance_ind(?src_key, ?dst_key, ?distance_key)
)
```

Summary:
- `eth_dos_distance, eth_dos_distance_ind, eth_dos_distance_base` go from 6 public arguments to 3
- `eth_dos_distance_ind` go from 12 arguments to 6

ed255:
Resolve #239 
I've managed to implement this feature with virtually no extra cost in the circuit by encoding the case `anchored_key.pod_id = SELF` in a statement template as a wildcard at index 0, and adding an offset +1 to regular wildcard indexes.  Then in the circuit the array of custom predicate arguments starts with a constant value of `SELF`, so resolving the wildcard index 0 returns `SELF`.  Since the random access in plonky2 has a fixed cost for any array up to 64 elements, and the number of arguments will always be much smaller than 64 the total number of constraints after this change is unaffected.

I've rewritten the ethdos to use `SELF` which allows the reductions described in the issue.  Now we can adjust the minimum parameters of a MainPod for ethdos which reduce the number of constraints from 30886 to 25347! (that's a 18% reduction of constraints needed).  Most importantly this change allows defining custom predicates that use less wildcards, so we can reduce the max number of wildcards per custom predicate in the parameters).

I've added a github action to build the code with the `metrics` feature because in the PR where I  implemented it I moved the code and broke it, but since the feature is disabled by default it wasn't checked to be building.  The bug in this PR.

### DIFF
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -2244,6 +2244,24 @@ mod tests {
         let expected_st_arg = StatementArg::Key(AnchoredKey::new(pod_id, Key::from("key")));
         helper_statement_arg_from_template(&params, st_tmpl_arg, args, expected_st_arg)?;
 
+        // case: AnchoredKey(SELF, key_literal)
+        let st_tmpl_arg = StatementTmplArg::AnchoredKey(
+            SelfOrWildcard::SELF,
+            KeyOrWildcard::Key(Key::from("foo")),
+        );
+        let args = vec![Value::from(1), Value::from(pod_id.0), Value::from(3)];
+        let expected_st_arg = StatementArg::Key(AnchoredKey::new(SELF, Key::from("foo")));
+        helper_statement_arg_from_template(&params, st_tmpl_arg, args, expected_st_arg)?;
+
+        // case: AnchoredKey(SELF, key_wildcard)
+        let st_tmpl_arg = StatementTmplArg::AnchoredKey(
+            SelfOrWildcard::SELF,
+            KeyOrWildcard::Wildcard(Wildcard::new("b".to_string(), 2)),
+        );
+        let args = vec![Value::from(1), Value::from(pod_id.0), Value::from("key")];
+        let expected_st_arg = StatementArg::Key(AnchoredKey::new(SELF, Key::from("key")));
+        helper_statement_arg_from_template(&params, st_tmpl_arg, args, expected_st_arg)?;
+
         // case: WildcardLiteral(wildcard)
         let st_tmpl_arg = StatementTmplArg::WildcardLiteral(Wildcard::new("a".to_string(), 1));
         let args = vec![Value::from(1), Value::from("key"), Value::from(3)];
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -1,4 +1,4 @@
-use std::{array, sync::Arc};
+use std::{array, iter, sync::Arc};
 
 use itertools::{zip_eq, Itertools};
 use plonky2::{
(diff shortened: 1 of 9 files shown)


src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -767,20 +767,26 @@ impl CustomOperationVerifyGadget {
         // optimization: ak_id_wc_index and wc_index use the same signals, so we only need to do one
         // random access to resolve both of them
         assert_eq!(ak_id_wc_index, wc_index);
+        // optimization: the wildcard indexes have an offset of +1.  This allows us to set a fixed
+        // SELF in args[0] to resolve SelfOrWildcard::SELF encoded as a wildcard at inde x0.
```
ax0, line 771:
```suggestion
        // optimization: the wildcard indices have an offset of +1.  This allows us to set a fixed
        // SELF in args[0] to resolve SelfOrWildcard::SELF encoded as a wildcard at index 0.
```

ax0 (approved):
LGTM!

src/backends/plonky2/circuits/mainpod.rs
ed255: thanks!

PR #241, opened 2025-05-22
calculate MainPod id in a dynamic-friendly way

ed255:
The MainPod id is now calculated with front padding and a fixed size independent of max_public_statements so that introduction gadgets can be verified by a MainPod while paying only for the number of statements they use.  This is because with front padding of none-statements we can precompute the poseidon state corresponding to absorbing all the padding statements and only pay constraints for the non-padding statements.

The id is calculated as follows:
`id = hash(serialize(reverse(statements || none-statements)))`

I introduced a new parameter `num_public_statements_id` which allows making differently-sized statement to id compatible.


arnaucube (approved):
LGTM!

PR #256, opened 2025-06-04
expose some interfaces for external usage (from introduction-pods)

arnaucube:
- expose some interfaces for external usage (from introduction-pods)
- add From<MainPod> for OperationArg, add copy op!


ax0 (approved):

ed255 (approved):
LGTM!

PR #257, opened 2025-06-05
Feat/add exp gate

ainta:
- added ExponentiationGate to common_data_for_recursion 
- ExponentiationGate is needed for ed25519pod in from introduction-pods(https://github.com/0xPARC/pod2/pull/256)

### DIFF
--- a/src/backends/plonky2/recursion/circuit.rs
+++ b/src/backends/plonky2/recursion/circuit.rs
@@ -342,7 +342,7 @@ pub fn common_data_for_recursion<I: InnerCircuit>(
         GateRef::new(plonky2::gates::random_access::RandomAccessGate::new_from_config(&config, 4)),
         GateRef::new(plonky2::gates::random_access::RandomAccessGate::new_from_config(&config, 5)),
         GateRef::new(plonky2::gates::random_access::RandomAccessGate::new_from_config(&config, 6)),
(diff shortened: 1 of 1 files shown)


ed255:
As discussed in the Telegram chat, the idea for introduction gadgets is that they can use their own configuration and set of gates without requiring modification of the standard set of gates from the recursive MainPod.  This is achieved by adding a wrapping circuit that "normalizes" the proof, see the current implementation of https://github.com/0xPARC/introduction-pods/pull/1 for reference.

src/backends/plonky2/recursion/circuit.rs
```diff
@@ -342,6 +342,7 @@ pub fn common_data_for_recursion<I: InnerCircuit>(
         GateRef::new(plonky2::gates::random_access::RandomAccessGate::new_from_config(&config, 4)),
         GateRef::new(plonky2::gates::random_access::RandomAccessGate::new_from_config(&config, 5)),
         GateRef::new(plonky2::gates::random_access::RandomAccessGate::new_from_config(&config, 6)),
+        GateRef::new(plonky2::gates::exponentiation::ExponentiationGate::new(66)),
```
ed255, line 345:
```suggestion
        GateRef::new(plonky2::gates::exponentiation::ExponentiationGate::new_from_config(&config)),
```
please use this constructor so that this still works if we ever change the config

ed255 (approved):
LGTM!

Issue #258, opened 2025-06-05
Deserialization of introduction pods defined in external libraries


ed255:
So far we have implemented the deserialization of pods with code that goes through each possible pod type as encoded in the serialization and deserializes data using the pod type methods:
https://github.com/0xPARC/pod2/blob/7bc0bd08d29499df15d147e818b3c1dbdfdd5e7e/src/frontend/serialization.rs#L80-L97

https://github.com/0xPARC/pod2/blob/7bc0bd08d29499df15d147e818b3c1dbdfdd5e7e/src/frontend/serialization.rs#L129-L163

Similarly for serialization we're also checking the exact type:
https://github.com/0xPARC/pod2/blob/7bc0bd08d29499df15d147e818b3c1dbdfdd5e7e/src/frontend/serialization.rs#L61-L73

https://github.com/0xPARC/pod2/blob/7bc0bd08d29499df15d147e818b3c1dbdfdd5e7e/src/frontend/serialization.rs#L103-L121

With introduction pods we can't keep doing this approach because everyone should be able to write their own introduction pods in their own libraries and it would be great if serialization and deserialization implemented in this library still work.

I think serialization is easier to resolve because we can add some methods to the `Pod` or `RecursivePod` trait.

As for deserialization it's going to be a bit more tricky.  My current thinking is that we can have a static HashMap under a mutex in this library that contains deserialization closures per pod type.  And then external libraries can register their deserialization methods into this HashMap that our library will use.

robknight:
Currently the main difference between the mock and circuit-based PODs is how the `proof` or `signature` is handled.

Is the same true for introduction PODs? So (de)serialization of things like statements would be shared between PODs, but (de)serialization of proof/signature components would need to be implementation-specific?

PR #260, opened 2025-06-06
Implement generic serialization/deserialization

ed255:
- Move all backend-specific serialization code to the backend.
- Extend the `Pod` trait to allow full serialization of any pod as a trait object so that serialization is fully generic.
- Implement deserialization for each pod in a generic way (following the same interface) and add a registry in the middleware to handle deserialization of any registered pod
  - Libraries that use pod2 can register new pod types to be deserialized.  They can do that by calling the new middleware function `register_pod_deserializer`
- The frontend no longer imports the backend for deserialization and instead uses the middleware functionality for generic deserialization.
- Change the implementation of `MainPod` and `MockMainPod` `pub_self_statements` method to return the non-padded list of statements so that they match with the data that the `PodBuilder` generates.  This way we can skip some data processing in serialization/deserialization.

Resolve https://github.com/0xPARC/pod2/issues/258


src/backends/plonky2/signedpod.rs
```diff
@@ -61,6 +64,12 @@ pub struct SignedPod {
     pub dict: Dictionary,
 }
 
+#[derive(Serialize, Deserialize)]
+struct Data {
+    signature: String,
+    dict: Dictionary,
```
robknight, line 70: The equivalent field for `MockSignedPod` is called `kvs`. I think it would be simpler for consumers if they had the same name, with the only difference being the pod_type and signature format.

robknight (commented):
lgtm! I think it would be worth aligning the name of `kvs` and `dict` but otherwise all looks good.

robknight (approved):

src/backends/plonky2/signedpod.rs
ed255: thanks, I've changed the field name :)

Issue #229, opened 2025-05-12
Allow literals anywhere in a statement


ed255:
Currently all statements take a list of anchored keys except for these two exceptions:
- `ValueOf` takes a literal `Value` as the last argument.
- `Custom` takes a list of literal values (in the form of `WildcardValue`, which can be a pod_id or a key)

`WildcardValue` is defined as:
```rust
pub enum WildcardValue {
    PodId(PodId),
    Key(Key),
}
```
But that's just for type checking, on the backend it all translates to the raw value of `PodId` or `Key`

Also `Value` and `WildcardValue` could be encoded the same way in the backend (we could make them overlap)

The custom predicate verification logic needs to resolve statement templates, which means the ability to build the list of arguments of a statement.  Because the statement could use a custom predicate, any of those arguments could be a `WildcardValue`, which means that the circuit has the logic to set values in any argument position.
Moreover the statement template could be written to expect a particular value in a particular argument.  Currently this is only supported for the value in a `ValueOf`.  But it would be easy to extend it to the WildcardValue.

So we have most of the circuit logic already to handle literal values anywhere in the arguments of a statement, yet we don't support literal values in arguments in statements, and instead we need to load them via `ValueOf`.

I would like to propose allowing literal values in any argument of a Statement.  So in general, the arguments to native predicates would be anchored keys or values.

There would be changes required in the circuit logic that verifies operations.  Currently we have many operations of the form `FooFromEntry` that take two statements of the type ValueOf.  We would need to change that so that any of those statements could be replaced by a literal.

Currently we have this:
```
ValueOf(SELF["a"], 10) <- NewEntry
ValueOf(SELF["b"], 20) <- NewEntry
Lt(SELF["a"], SELF["b"]) < LtFromEntries(ValueOf(SELF["a"], 10), ValueOf(SELF["b"], 20) )
```
And we would need to support this:
```
ValueOf(SELF["a"], 10) <- NewEntry
Lt(SELF["a"], 20) < LtFromEntries(ValueOf(SELF["a"], 10), 20)
```
I haven't evaluated how big this change would be, and how many extra constraints it would need.

## Why am I proposing this?

I believe this change would make the circuits leaner for actual use cases.  Since we encode all the verifying logic in circuits we need to set upper bounds to many things.  With the proposed change we would need:
- Less statements
- Less arguments
Which would allow us to reduce the total number of constraints needed for particular use cases.  Reducing those parameters means less values to hash, less cells, less values to do random access to.

[...]

PR #268, opened 2025-06-09
update docs to use pod lang v1

### LINKED ISSUE #266: Update docs and examples to use the POD2 language
The language was introduced in #225

ed255:
Resolve #266 

The examples in `examples.md` already use "literals anywhere" feature (that's because those examples were higher level descriptions of use-cases, although the language with literals anywhere is now rich enough to encode those examples).
I've only taken the liberty of nesting ANDs inside ORs, which the language doesn't support just to keep the examples as succinct as they were before.
The eth_dos example in `book/src/customexample.md` is using the old approach of passing wildcards for origin and keys, which don't work in recursion.  We can update that once #229 is resolved.

### DIFF
--- a/book/src/examples.md
+++ b/book/src/examples.md
@@ -11,16 +11,25 @@ An EthDos Pod exposes a single custom statement with two custom deduction
 rules, the inductive case and the base case.
 
 ```
-statement eth_dos_distance(src: PubKey, dst: PubKey, distance: Int):
-   - OR():
-      - AND(attestation_pod: Pod, intermediate: PubKey, n: int):
-         - eq(attetation_pod.attestation, dst)
-         - eq(attetation_pod.type, SIGNATURE)
-         - sum_of(distance, 1, n)
-         - eth_dos_distance(src, attetation_pod.signer, n)
-      - AND():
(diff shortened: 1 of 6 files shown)


book/src/examples.md
```diff
@@ -41,29 +50,32 @@ Date values:
...
+    Lt(?gov_id.date_of_birth, NOW_MINUS_18Y) # date_of_birdth is more than 18y old
+    Equal(?paystub[KEY_SIGNER], ZOO_DEEL)
+    Equal(?paystub[KEY_TYPE], SIGNATURE)
+    Equal(?paystub[ssn], ?gov_id.ssn)
+    // At least one year of consistent employment with your current employer
+    Lt(?paystub["start_date"], NOW_MINUS_1Y) # start_date is more than 1y old
+    Gt(?paystub["issue_date"], NOW_MINUS_7D) # issue_date is less than 7d old
+    // Annual salary is at least $20,000
+    Gt(?paystub["annual_salary"], 20000)
+    // Private key knowledge
```
ed255, line 71: I've updated this to proof knowledge of the sk by having a signed pod with an auth entry.

Issue #262, opened 2025-06-09
Explore num gates reduction in Schnorr signature with ECDouble


ed255:
Currently for the Schnorr signature we're using the ECAdd gate to implement doubling.  But a doubling gate requires less multiplications than an adding gate, so we could save some gates by implementing it.  On the other hand, we'd have 2 extra custom gates to verify in the recursive verifier.

For reference https://eprint.iacr.org/2022/274.pdf in section "5.2 Curve Operations" says that "Using fractional (x, u) coordinate" they get:
- point addition in 10M
- point doubling in 4M + 5S

PR #271, opened 2025-06-10
feat(backend): add EC double gate

ax0:
Towards #262.


ed255:
Here are the measurements comparing main https://github.com/0xPARC/pod2/commit/6ab0bc52fce656e08a1b2955c97ab9cb623a76eb with this PR https://github.com/0xPARC/pod2/pull/271/commits/d32fb88606ba6974a41a720e4ee5340b74153a02

# Main
- verify signature: 1703 gates
- verify recursive MainPod proof (2^15): 4717 gates
- proving time for recursive MainPod proof (2^15): 3.23 seconds (AMD Ryzen 5 3600)
# This PR
- verify signature: 985 gates
- verify recursive MainPod proof (2^15): 4778 gates
- proving time for recursive MainPod proof (2^15): 3.57 seconds (AMD Ryzen 5 3600)

# Summary

## Precise deltas
- Signature verification goes from 1703 -> 985 (-718) which is very significant.  The more signature verification we want, the more important this is.  For example if we want a configuration with 5 signature verifications, this would save 3590 gates!
- Recursive verification goes from 4717 -> 4778 (+61).  I think we'll only do 2 recursive verification, so this adds 124 gates in our recursive circuit, which is totally acceptable.

## Unprecise deltas
I don't have enough runs of the proving to get accurate results so I'll assume 10% error in each measure

- Proving time goes from 3.23 +/- 10% to 3.57 +/- 10%.  I don't think this increase is coming from more commitment work, because more routing wires means more fixed polynomials for the permutation argument, and I would assume these commitments are precomputed.

## Conclusion
The gate saving is very significant.  To make a decision of whether to apply this change as is or not we need to think about our target number of gates, and what do we want to fit in them, and also considering zk.

For example right now the config for the recursive MainPod would not have room for the blinders to add zk in 2^15.  If we apply this change we would save 1436, which is still not enough for zk, but we'd be closer to staying within 2^15.

A similar analysis can be done for the config with non-recursive MainPod without custom predicates.

Perhaps another avenue of exploration would be trying to apply this optimizations without changing the config (so we get all the benefits and no downsides in extra proving time).  We could do that by not having an automated recursive gate, and instead do the recursive verification using plonky2 already existing extension arithmetic gates.

ax0:
Superseded by #397.

PR #277, opened 2025-06-11
Better Hash and PartialEq for Keys and AnchoredKeys

robknight:
When adding the original serialization code, I added the Hash and PartialEq traits to some of our structs. The default implementation hashes both the `hash` and `name` fields together, which is unnecessary - we can hash just the `hash` field instead. Similarly, PartialEq does not need to compare both fields, it can compare just the hash field. This mirrors how the same traits were implemented for the `Value` type.

This should make storing Keys/AnchoredKeys in hashmaps/sets slightly more efficient.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -443,7 +443,7 @@ impl From<&Value> for Hash {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ed255 (approved):
LGTM!

PR #285, opened 2025-06-12
docs: add initial docs of Introduction PODs section, with some diagrams

arnaucube:


book/src/introductionpods.md
```diff
@@ -0,0 +1,63 @@
...
+There are some cases, where the operations needed for the `IntroductionPod`
+require a different circuit configuration than the standard recursive circuit
+configuration that is used for the `MainPods`.
+
+In those cases, we do a previous recursive step, where we verify the logic of
+the intended `IntroductionPod`, and then the proof of this circuit is the one
+verified in the real `IntroductionPod`.
+
+This is done in order to be able to *converge* the different circuit
+configuration shape (and thus different verifier_data) into a new proof that can
```
ed255, line 44: Maybe you mean `common_data` instead of `verifier_data` here?

book/src/introductionpods.md
```diff
@@ -0,0 +1,63 @@
...
+be verified with the standard configuration (used for the MainPods).
+
+<img src="img/introductionpod-2-steps.png" style="max-width:80%;" />
```
ed255, line 47: The last arrow in this diagram also mentions `verifier_data` but perhaps it should be `common_data`?

The `verifier_data` of different circuits will be different.  But a recursive in-circuit verifier is able to verify different circuits (each with a different `verifier_data`) as long as they share the `common_data`

ed255 (approved):
Overall looks good!  I've added two comments about something that I found confusion where you talk about `verifier_data` of different circuits being the same, and perhaps you meant `common_data`?
Please take a look.  I'm already approving the PR so that you can merge it after resolving the comments.

book/src/introductionpods.md
arnaucube: Oh right! Thanks for pointing it out ^^

PR #286, opened 2025-06-13
Basic 'use' syntax for importing custom predicates

### LINKED ISSUE #284: Import custom predicates in Podlang
Currently, Podlang code cannot refer to externally-defined custom predicates. This is fine for toy examples but breaks down for anything more complex.

Therefore, we need some kind of `use` or `import` syntax.

## Proposal

```
use (pred1, pred2) from BATCH_ID;

REQUEST(
  pred1(?A)
  pred2(?B)
)
```

Where BATCH_ID is a very long hex string, e.g. `e6aaf6d46d3c44e03dbbc41cc606c8a7228dbdea1f04e671c68aa056c723258e45a2c1835ceef52a958e5fd3440e38bcf63ed16a6c4360f79ddba55b8aa1b02b4f722f219528e6cafe1c89b4423d6ad667432b4ca4af199493c4fa39b57399a85918022e1556024a1f9b4f7ebfb5b89e1568632719d743e03f1834fe02665eab`, corresponding to the ID (hash) of the custom predicate batch you want to import. We _might_ want to consider alternate encodings for shorter strings, probably as a language syntactic sugar feature.

We can extend the parse API to allow a collection of batches to be provided, like so:

```rs
let batch = ...; // define some custom predicate batch
parse(program_text, [batch]);
```

The resulting proof request would contain references to the externally-defined `CustomPredicateBatch`.

We could chain several parse operations together, with the later ones importing the results of the earlier ones:

```rs
let eth_dos_code = r#"
  // ethdos predicates go here
"#;

// custom_batches is a vec of CustomPredicateBatch (arguably it should not be able to return multiple batches, that was future-proofing which may not have been necessary)
let ProcessedOutput{ custom_batches } = parse(eth_dos_code, []);

let proof_request_code = r#"
  // eth_dos_distance is the 4th predicate in the batch
  use (_, _, _, eth_dos_distance) from BATCH_ID;
  REQUEST(eth_dos_distance("alice", "charlie", ?distance)
"#;

let ProcessedOutput{ request_templates } = parse(proof_request_code, [batches[0]])
```

`request_templates` would then contain the proof request, with references to the `CustomPredicateBatch`.

A further optimization would be to allow the parsing of multiple batches at the same time:

```rs
// Parser figures out how to do dependency resolution
parse([eth_dos_code, proof_request_code]);
```

robknight:
Closes #284 

This PR defines a `use` syntax for Podlang code:

```
use p1, p2, _, p4 from BATCH_ID
```

Where `BATCH_ID` the hex string representing the custom predicate batch's ID.

When parsing, we can now provide known batches to the parser:

```
parse(program_text, params, &[batch1, batch2])
```

This will enable the parser to look up the referenced custom predicate by BATCH_ID and index.

Predicates are imported according to their index:

```
use _, p2, _, p4 from BATCH_ID
```

This imports the predicates at indices 1 and 3, and skips 0 and 2.

The names given are aliases, and do not need to match the original names given to the predicates. This can help to avoid name collisions.

## Possible enhancements

Managing batch IDs might be annoying, so I considered the possibility of making the `batches` vector a HashMap of `String -> CustomPredicateBatch`, with the strings providing aliases to make the `use` statements easier to read. This might also speed up development cycles in cases where we want to test a custom predicate or proof request that imports a CustomPredicateBatch, and we are making repeated changes to the definition of the imported batch. With static batch IDs in source code, changes to the imported code will require changes at the point of import, too, since the hash will change. I'm open to adding some kind of aliasing/indirection but wanted to avoid over-engineering on the first pass.

### DIFF
--- a/src/lang/error.rs
+++ b/src/lang/error.rs
@@ -17,7 +17,7 @@ pub enum LangError {
     Frontend(Box<frontend::Error>),
 }
 
-/// Errors that can occur during the processing of Podlog Pest tree into middleware structures.
+/// Errors that can occur during the processing of Podlang Pest tree into middleware structures.
 #[derive(thiserror::Error, Debug)]
 pub enum ProcessorError {
     #[error("Undefined identifier: '{name}' at {span:?}")]
@@ -69,6 +69,22 @@ pub enum ProcessorError {
     InvalidLiteralFormat {
         kind: String,
         value: String,
+        span: Option<(usize, usize)>,
+    },
+    #[error("Batch with ID '{id}' not found at {span:?}")]
+    BatchNotFound {
+        id: String,
+        span: Option<(usize, usize)>,
+    },
+    #[error("Number of predicates in 'use' statement ({found}) exceeds the number of predicates in the batch ({expected}) at {span:?}")]
+    ImportArityMismatch {
+        expected: usize,
+        found: usize,
+        span: Option<(usize, usize)>,
+    },
+    #[error("Duplicate import name '{name}' at {span:?}")]
+    DuplicateImportName {
+        name: String,
         span: Option<(usize, usize)>,
     },
     #[error("Frontend error: {0}")]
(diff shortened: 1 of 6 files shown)


ed255 (approved):
LGTM!  I've left a small suggestion.

PR #288, opened 2025-06-16
recursion circuit's verifier_data_hash include constant_sigmas_cap in the hash, and add explanation

### LINKED ISSUE #275: In Recursive circuit, verify that `circuit_digest` is calculated using `constants_sigmas_cap`
In the RecursiveCircuit we're creating the targets for the verifier data like this, which doesn't create any constraints:
https://github.com/0xPARC/pod2/blob/6258e52e1acecbf57e5d735a6b565063a9192d4b/src/backends/plonky2/recursion/circuit.rs#L190-L192

Then we're passing the `circuit_digest` to the inner circuit so that it can check whether it's a valid circuit or not:
https://github.com/0xPARC/pod2/blob/6258e52e1acecbf57e5d735a6b565063a9192d4b/src/backends/plonky2/recursion/circuit.rs#L204

But upon checking the `verify_proof` method I observed that it uses `circuit_digest` just as a commitment to derive Fiat Shamir challenges.
There's no check that the `circuit_digest` was calculated using the `constants_sigmas_cap`!

The `constants_sigmas_cap` determine the constants used in the circuit as well as the copy constraints, which basically allow us to completely change the behavior of the circuit, and this value is never verified!

We need to constraint that `circuit_digest` is obtained by hashing something that contains the `constants_sigmas_cap`!

(A) This is how the `circuit_digest` is calculated: https://github.com/0xPolygonZero/plonky2/blob/82791c4809d6275682c34b926390ecdbdc2a5297/plonky2/src/plonk/circuit_builder.rs#L1256-L1264
We'll need some extra witnesses for that.

(B) Another option is to stop using the `circuit_digest` and instead use:
- `circuit_hash = hash(circuit_digest, constants_sigmas_cap)`
Which is something we actually had in the original RecursiveCircuit PR by @arnaucube and I suggested removing that because `circuit_digest` already included the `constant_sigmas_cap`...

I think option (A) is more clear, but (B) is easier.

arnaucube:
resolves #275


ed255 (approved):
LGTM!

PR #300, opened 2025-06-18
move pod data's deserialization to RecursivePod trait (#294), expose get_common_data

### LINKED ISSUE #294: Add `deserialize` method to the `RecursivePod` trait
We need to add it like this:
```rust
    fn deserialize(
        params: Params,
        id: PodId,
        vds_root: Hash,
        data: serde_json::Value,
    ) -> Result<Box<dyn RecursivePod>>
    where
        Self: Sized;
```
because the `RecursivePod` is used with `dyn` (for trait objects).

Then implementations of introduction pods will be forced to implement that method.
See https://github.com/0xPARC/introduction-pods/pull/4#discussion_r2151749568

arnaucube:
resolves #294 , unblocks https://github.com/0xPARC/introduction-pods/pull/4

### DIFF
--- a/src/backends/plonky2/emptypod.rs
+++ b/src/backends/plonky2/emptypod.rs
@@ -158,23 +158,6 @@ impl EmptyPod {
         })
         .map_err(|e| Error::custom(format!("EmptyPod proof verification failure: {:?}", e)))
     }
-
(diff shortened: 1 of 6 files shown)


ed255 (approved):
LGTM!

PR #304, opened 2025-06-20
feat: compress EC subgroup points before serialising

### LINKED ISSUE #301: Compress/Uncompress public keys
The public keys used in our signature scheme are elliptic curve points.  The current representation is 2 coordinates where each one is built from 5 goldilocks fields.
For serialization/display purposes that's a lot of bits.

Having functions to compress the point and uncompress them would be ideal to have a more succint serialization/display of the public keys.

ax0:
Resolves #301.

### DIFF
--- a/src/backends/plonky2/primitives/ec/curve.rs
+++ b/src/backends/plonky2/primitives/ec/curve.rs
@@ -9,6 +9,7 @@ use std::{
 };
 
 use num::{bigint::BigUint, Num, One};
+use num_bigint::RandBigInt;
 use plonky2::{
     field::{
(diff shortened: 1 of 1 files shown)


src/backends/plonky2/primitives/ec/curve.rs
```diff
@@ -688,6 +744,13 @@ mod test {
         assert_eq!(p2, p3);
     }
 
+    #[test]
```
ed255, line 747: Would you mind adding a test that does the roundtrip from point in the subgroup -> compress_from_subgroup -> decompress_into_subgroup -> assert_eq with the original point?
For some pseudorandom points?

ed255 (approved):

PR #305, opened 2025-06-20
Display point in base58

ed255:
Depends on #304 

Implement Display for Point like this:
- In alternate mode, show the `x`, `u` coordinates using the display of the extension field
- In normal mode if the point is in the subgroup display the `u` coord in base58.  If not in the subgrup display the `x` and `u` coords concatenated in base58. 

Implement the serialization/deserialization using the base58 encoding.

The reason to use base58 VS base64 is to make it easier to select an entire public key and make it easy to compare two public keys or even read them aloud.


ax0 (approved):
LGTM!

PR #306, opened 2025-06-20
add zk config, enabled by a feature (on by default)

### LINKED ISSUE #248: Enable zk in our circuits
Most of our circuits are now built with a standard plonky2 config (without zk)

We need to use a config with zk.  Once we do that, most likely things in the recursion will break and we'll need to fix them

### LINKED ISSUE #252: Design of the "zk" feature
Related to https://github.com/0xPARC/pod2/issues/248

The current plan to add zk to our circuits (which is also the simplest) is to just use the standard configuration with zk in the `Recursive<MainPod>` circuit.  But from previous benchmarks we know that adding zk introduces a significant amount of rows.
Let's call it option A:

Image

Another approach is to configure our circuits without zk and then add a wrapping layer that adds zk.  The wrapping circuit would be smaller and has two goals: adding zk and normalizing the proof shape so that it can be verified by an universal verifier in the `Recursive<MainPod>`.  This may also be advantageous for introduction pods because we no longer need to pad them to a big circuit or wrap them with a big circuit to normalize them (in case they were already the proper size but used a different configuration / common)

Image

NOTE: We arrived to this two options in a call with @arnaucube and for now we decided to start with Option A (which is simpler) but consider Option B in the future.

arnaucube:
resolves #248, closes #252 (implements the described 'Option A')

### DIFF
--- a/src/backends/plonky2/recursion/circuit.rs
+++ b/src/backends/plonky2/recursion/circuit.rs
@@ -415,7 +415,7 @@ fn estimate_verif_num_gates(degree_bits: usize) -> usize {
     #[cfg(feature = "zk")]
     {
         // Formula obtained via linear regression using
-        // `test_measure_zk_recursion` results with `standard_recursion_config`.
+        // `test_measure_zk_recursion` results with `standard_recursion_zk_config`.
         num_gates = 244 * degree_bits + 1127;
     }
     #[cfg(not(feature = "zk"))]
--- a/src/backends/plonky2/recursion/circuit.rs
+++ b/src/backends/plonky2/recursion/circuit.rs
@@ -164,8 +164,12 @@ impl<I: InnerCircuit> RecursiveCircuit<I> {
 
     /// builds the targets and returns also a ProverCircuitData
     pub fn build(params: &RecursiveParams, inner_params: &I::Params) -> Result<Self> {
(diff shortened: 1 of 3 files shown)


src/backends/plonky2/recursion/circuit.rs
```diff
@@ -398,9 +410,20 @@ fn standard_gates(config: &CircuitConfig) -> Vec<GateRef<F, D>> {
...
 fn estimate_verif_num_gates(degree_bits: usize) -> usize {
-    // Formula obtained via linear regression using `test_measure_recursion` results with
-    // `standard_recursion_config`.
-    let num_gates: usize = 236 * degree_bits + 1171;
+    let num_gates: usize;
+
+    #[cfg(feature = "zk")]
+    {
+        // Formula obtained via linear regression using
+        // `test_measure_zk_recursion` results with `standard_recursion_config`.
```
ax0, line 418:
```suggestion
        // `test_measure_zk_recursion` results with `standard_recursion_zk_config`.
```

ax0 (approved):
LGTM!

src/backends/plonky2/recursion/circuit.rs
arnaucube: oh right! thanks, added

PR #314, opened 2025-06-25
resolve #273: remove global constant MAX_DEPTH, instead use the respective value from Params

### LINKED ISSUE #273: remove the global constant `MAX_DEPTH` and use the respective value from `Params` instead
The constant [`MAX_DEPTH`](https://github.com/0xPARC/pod2/blob/77c96d5dbee1dcddcfaa71b15d1617d7440368db/src/constants.rs#L1) is only used at the 4 following locations:
- src/frontend/serialization.rs#85 for the serialization of `SignedPod`
- src/backends/plonky2/signedpod.rs#77 at the method `_verify`
- src/backends/plonky2/mock/signedpod.rs#72 at the method `_verify`
- src/lang/processor.rs method `process_literal_value`

Ideally we can get rid of the `MAX_DEPTH` global constant, and use everywhere the [`params.max_depth_mt_containers`](https://github.com/0xPARC/pod2/blob/45fdef9304441fc38b8b360350230ae66bd5c02d/src/middleware/mod.rs#L600).

arnaucube:
resolves #273

### DIFF
--- a/src/backends/plonky2/mock/mainpod.rs
+++ b/src/backends/plonky2/mock/mainpod.rs
@@ -40,7 +40,7 @@ impl PodProver for MockProver {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct MockMainPod {
     params: Params,
     id: PodId,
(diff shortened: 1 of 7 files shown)


ax0 (approved):
LGTM!

PR #318, opened 2025-06-27
Restore typed data to serialized PODs

### LINKED ISSUE #310: Serialization lacks type information for JSON Schema
The previous serialization code used strongly-typed `SerializedSignedPod` and `SerializedMainPod` structs, which allowed us to produce a JSON Schema describing their contents. When we switched to using the `data` field to hold most serialized data, we typed this as `serde_json::Value`, which is a generic type that can hold any kind of data. As a result, the JSON Schema would treat anything at all in the `data` field as being valid. This also breaks use-cases like generating TypeScript types from the JSON Schema.

The schema for SignedPod entries and MainPod public statements should be part of the "contract" with third party systems/clients that rely on our serialization format, and should be strongly-typed. "Opaque" data like proofs or signatures can remain as part of the untyped data field.

robknight:
Closes #310 

This reintroduces the `public_statements` and `entries` fields for `SerializedMainPod` and `SerializedSignedPod` respectively. This allows us to generate meaningful JSON Schemas that tell third-party (non-Rust) clients what data they can find in the serialization format.

This does introduce a smaller problem: there is now duplication of data between the typed `entries`/`public_statements` field and the untyped `data` blob, which includes its own copies of that data. However, this is less bad than not having any meaningful schema for our serialization format.

### DIFF
--- a/src/frontend/serialization.rs
+++ b/src/frontend/serialization.rs
@@ -1,10 +1,14 @@
+use std::collections::HashMap;
+
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use super::Error;
 use crate::{
     frontend::{MainPod, SignedPod},
-    middleware::{deserialize_pod, deserialize_signed_pod, Params, PodId, VDSet},
+    middleware::{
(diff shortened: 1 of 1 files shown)


ed255 (approved):
LGTM!

PR #321, opened 2025-06-30
Detect invalid wildcards in the language processor

### LINKED ISSUE #320: Unwrap on None with certain (invalid?) predicates
Hi,

Trying to parse the following custom predicate leads to an unwrap on a None option.

```
        identity_verified(username, private: identity_pod) = AND(
            Equal(?identity_pod["{key_type}"], {signed_pod_type})
            Equal(?identity_pod["{key_signer}"], {identity_server_pk:#})
            Equal(?identity_pod["username"], ?username)
            Equal(?identity_pod["user_public_key"], ?user_public_key)
        )
```
Notice that ?user_public_key is not defined.

Tracing the issue leads to https://github.com/0xPARC/pod2/blob/3306f360f794c6a2d06043342bbf98ed343fdd94/src/frontend/custom.rs#L218.

I assume this predicate is invalid, though the parser should throw an error which is more interpretable to the user.

robknight:
Closes #320 

The language parser/processor currently does not detect invalid wildcards (that is, the use of a wildcard not declared in the argument list) directly, and instead allows `CustomPredicateBuilder` to throw an error. However, this loses a lot of context about the exact point in the code where the error occurs. By checking this earlier, we can return more useful errors.

### DIFF
--- a/src/lang/error.rs
+++ b/src/lang/error.rs
@@ -60,9 +60,10 @@ pub enum ProcessorError {
     Internal(String),
     #[error("Middleware error: {0}")]
     Middleware(middleware::Error),
-    #[error("Undefined wildcard: '?{name}' at {span:?}")]
+    #[error("Undefined wildcard: '?{name}' in predicate '{pred_name}' at {span:?}")]
     UndefinedWildcard {
         name: String,
+        pred_name: String,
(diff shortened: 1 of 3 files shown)


ed255 (approved):
LGTM!

PR #322, opened 2025-06-30
Assorted tweaks to support external playground crate

robknight:
These are various tweaks which allow the playground crate to import pod2 as an external crate. In particular, the `examples` feature allows external crates to import things like the Ethdos batch.

The other changes are mostly adding some extra `derive`s or methods to make crate-private fields accessible.


src/middleware/basetypes.rs
```diff
@@ -125,6 +125,32 @@ impl fmt::Display for RawValue {
     }
 }
 
+impl ToHex for RawValue {
```
arnaucube, line 128: One question, I see that before this PR, the type `Hash` had implementations for the traits `ToHex` and `FromHex`. Here it is introducing `ToHex` for the type `RawValue`; should we add also the implementation of `FromHex` for the type `RawValue`? (so that it is available in the future for when converting back and forth the `RawValue` from & to hex)

arnaucube (commented):
LGTM! just left a small question

src/middleware/basetypes.rs
robknight: I've added the `FromHex` trait as well.

PR #327, opened 2025-07-01
Automatically implement Pod::equals

dgulotta:
This patch introduces an automatic implementation of Pod::equals, so that it does not have to be manually implemented for each pod.


arnaucube (approved):
Nice!!

PR #330, opened 2025-07-07
chore(middleware): additional error reporting for custom predicates

ax0:


arnaucube (approved):
LGTM! tagged also @ed255 as reviewer

src/middleware/operation.rs
```diff
@@ -370,79 +370,87 @@ pub fn check_st_tmpl(
...
             StatementArg::Key(AnchoredKey { pod_id, key }),
         ) => {
             let pod_id_ok = check_or_set(Value::from(*pod_id), pod_id_wc, wildcard_map);
-            pod_id_ok && (key_tmpl == key)
+            pod_id_ok.and_then(|_| {
+                (key_tmpl == key).then_some(()).ok_or(
+                    Error::mismatched_anchored_key_in_statement_tmpl_arg(
+                        key_tmpl.clone(),
+                        key.clone(),
+                    ),
```
ed255, line 403: This error only has context for the check `key_tmpl == key`.
But it will also be returned if `pod_id_ok = false`.

I have two alternative suggestions:
a. Extend the error to also include the context of the pod_id
b. Add a new error just for `pod_id_ok = false`.

src/middleware/operation.rs
```diff
@@ -370,79 +370,87 @@ pub fn check_st_tmpl(
...
     // NOTE: We set unresolved wildcard slots with an empty value.  They can be unresolved because
     // they are beyond the number of used wildcards in this custom predicate, or they could be
     // private arguments that are unused in a particular disjunction.
-    Some(
-        wildcard_map
-            .into_iter()
-            .map(|opt| opt.unwrap_or(Value::from(0)))
-            .collect(),
-    )
+    // TODO: Should we keep track of this for less cryptic error reporting?
```
ed255, line 441: Do you mean keep track of the `Option<_>` instead of turning `None` into `0`?

src/middleware/operation.rs
```diff
@@ -370,79 +370,87 @@ pub fn check_st_tmpl(
...
+        .collect())
 }
 
 fn check_custom_pred(
     params: &Params,
     custom_pred_ref: &CustomPredicateRef,
     args: &[Statement],
     s_args: &[Value],
-) -> Result<bool> {
+) -> Result<()> {
```
ed255, line 453: :partying_face:

src/middleware/operation.rs
```diff
@@ -476,23 +484,31 @@ fn check_custom_pred(
...
-        None => return Ok(false),
-    };
+    let wildcard_map = resolve_wildcard_values(params, pred, args)?;
 
-    // Check that the resolved wildcard match the statement arguments.
+    // Check that the resolved wildcards match the statement arguments.
     for (s_arg, wc_value) in s_args.iter().zip(wildcard_map.iter()) {
         if *wc_value != *s_arg {
-            return Ok(false);
+            return Err(Error::mismatched_wildcard_value_and_statement_arg(
```
ed255, line 492: Maybe we can add the `pred` here and also the index of the argument for better context?

src/middleware/error.rs
```diff
@@ -18,6 +20,18 @@ pub enum MiddlewareInnerError {
...
+    #[error("{0} does not match {1}")]
+    MismatchedAnchoredKeyInStatementTmplArg(Key, Key),
+    #[error("{0} does not match against {1}")]
+    MismatchedStatementTmplArg(StatementTmplArg, StatementArg),
+    #[error("{0} does not match {1}")]
+    MismatchedWildcardValueAndStatementArg(Value, Value),
+    #[error("Not all statement templates of {0} have been matched")]
+    UnsatisfiedCustomPredicateConjunction(CustomPredicate),
+    #[error("None of the statement templates of {0} have been matched")]
+    UnsatisfiedCustomPredicateDisjunction(CustomPredicate),
```
ed255, line 34: The `Display` of a `CustomPredicate` has many new lines: https://github.com/0xPARC/pod2/blob/2c41a6c5543a3167c816d6e668478c3cd6f13700/src/middleware/custom.rs#L326-L357

I would suggest rewriting the error message so that the `{0}` appears at the end after a new line like:
`"Not all statement templates of the following custom predicate have been matched:\n{0}"`

ed255 (commented):
Overall looks good!  I've left some suggestions for improvement.  Please take a look.

PR #344, opened 2025-07-16
MerkleTree insertion proofs

arnaucube:
- implement `tree.insert` method to add new leaves to the tree (till now we could only add leaves at tree initial creation, not after), which returns a proof of correct insertion (proof of correct *state transition*)
- implement verification of proofs of correct insertion of leaves
- implement circuit that verifies proofs of correct insertion
- fix graphviz method after endianness change

Update: thanks @ax0 for pointing out the [tampering proofs case](https://github.com/0xPARC/pod2/pull/344/commits/79a532e97bc4e63d465b0445469e576adf8d3ef6), which required adding the [check 5.3](https://github.com/0xPARC/pod2/pull/344/commits/2da2e8237b35b8d50e9a4dcbeefb2a0285a677d6).

### DIFF
--- a/src/backends/plonky2/primitives/merkletree/circuit.rs
+++ b/src/backends/plonky2/primitives/merkletree/circuit.rs
@@ -604,7 +604,7 @@ pub mod tests {
     use crate::{
         backends::plonky2::{
             basetypes::C,
-            primitives::merkletree::{keypath, kv_hash, MerkleTree},
+            primitives::merkletree::{keypath, kv_hash, MerkleProof, MerkleTree},
         },
         middleware::{hash_value, RawValue},
     };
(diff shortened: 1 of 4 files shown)


src/backends/plonky2/primitives/merkletree/circuit.rs
```diff
@@ -385,6 +391,244 @@ fn kv_hash_target(
...
+            |acc, (old, new)| {
+                let eq_at_i = builder.is_equal(old, new);
+                builder.and(acc, eq_at_i)
+            },
+        );
+        let old_is_noteq_new = builder.not(old_is_eq_new);
+        let in_case_5_2 = builder.and(old_is_noteq_new, is_divergence_level);
+
+        // do the case2's checks
+        let old_leaf_hash = kv_hash_target(
```
ax0, line 534: This is also computed in the call to `verify_merkle_proof_circuit` above. We could return it from there, but that may seem unnatural. I'm also not sure if the gate saving is worth it.

ax0 (approved):
LGTM!

src/backends/plonky2/primitives/merkletree/circuit.rs
arnaucube: Update: as commented in the call, the difference is of just 2 constraints, so we keep it without modifying the `verify_merkle_proof_circuit` interface; but, [this commit](https://github.com/0xPARC/pod2/pull/344/commits/898c3dc4f944b1fbf8ad23b4d3fa8a8faaec9941) moves the computation of the old_leaf_hash to outside the loop, and simplifies the boolean operations for the check 5.3. Thanks for pointing it out!

src/backends/plonky2/primitives/merkletree/circuit.rs
```diff
@@ -385,6 +391,234 @@ fn kv_hash_target(
...
+        builder.conditional_assert_eq(
+            proof.enabled.target,
+            proof.proof_non_existence.key.elements[j],
+            proof.new_key.elements[j],
+        );
+        builder.conditional_assert_eq(
+            proof.enabled.target,
+            proof.proof_non_existence.value.elements[j],
+            proof.new_value.elements[j],
+        );
```
ed255, line 473: Isn't the `value` field irrelevant in a non-existence proof?

src/backends/plonky2/primitives/merkletree/circuit.rs
```diff
@@ -385,6 +391,234 @@ fn kv_hash_target(
...
+            .collect();
+        for j in 0..HASH_SIZE {
+            builder.conditional_assert_eq(proof.enabled.target, old_sibling_i[j], new_sibling_i[j]);
+        }
+
+        // 5.2) when i==d && if old_siblings[i] != new_siblings[i], check that:
+        //   old_siblings[i] == EMPTY_HASH && new_siblings[i] == old_leaf_hash
+
+        // in_case_5_2=true if: i==d (= is_divergence_level) && old_siblings[i]!=new_siblings[i]
+        let old_is_eq_new = zip_eq(old_siblings[i].elements, new_siblings[i].elements).fold(
```
ed255, line 518: you could use `reduce` (https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce) instead of `fold` here because the len of siblings is always > 0.

src/backends/plonky2/primitives/merkletree/circuit.rs
```diff
@@ -385,6 +391,234 @@ fn kv_hash_target(
...
+    // 5.1) old_siblings[i] == new_siblings[i] ∀ i \ {d}
+    // 5.2) at i==d, if old_siblings[i] != new_siblings[i]: old_siblings[i] ==
+    //   EMPTY_HASH new_siblings[i] == old_leaf_hash
+    // 5.3) assert that if old_key!=empty, both old_leaf_path&new_leaf_path
+    //   should diverge at the inputted divergence level
+    let old_siblings = proof.proof_non_existence.siblings.clone();
+    let new_siblings = proof.new_siblings.clone();
+    for i in 0..proof.max_depth {
+        let i_targ = builder.constant(F::from_canonical_u64(i as u64));
+        let is_divergence_level = builder.is_equal(i_targ, proof.divergence_level);
```
ed255, line 500: what would happen if `proof.divergence_level` is not the first level where the paths diverge but a later one?

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -175,6 +204,89 @@ impl MerkleTree {
...
+            if old_siblings[d] != EMPTY_HASH {
+                return Err(TreeError::state_transition_fail(
+                    "siblings don't match: old[d]!=empty".to_string(),
+                ));
+            }
+            let k = proof
+                .proof_non_existence
+                .other_leaf
+                .map(|(k, _)| k)
+                .unwrap_or(proof.new_key);
```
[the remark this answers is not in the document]
arnaucube: Yes, but, imagine a prover that feeds a 'malicious' proof that does not contain the expected `other_leaf` when it should; if we have the `.unwrap()` the library will panic.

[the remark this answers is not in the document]
ed255: That's true!
Then what about returning an error?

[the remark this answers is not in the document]
arnaucube: right, added the error return

PR #345, opened 2025-07-16
always replace SELF when copying statements

### LINKED ISSUE #342: Pod IDs in custom statements
If I understand correctly, the system currently does not distinguish whether a custom statement argument is a pod ID, key, or value.  In particular, this means that if SELF appears in the argument to a custom statement, it won't get converted to the pod ID when the statement is copied.

For example, suppose we have the following custom predicate.
```
custom_equal(pod, key, value) = AND(
   Equal(?pod[?key], ?value)
)
```
If the user makes a `custom_equal` statement about a key in the current pod, then even after the statement is copied into another pod, it will still have `SELF` in the first argument.  This is probably not what the user expects.

I can see two ways that this could be fixed.

1. Add typing to arguments in custom operations.  If an argument has type "pod ID", then it can only be used in native statements as either as the first half of an anchored key or in Equals/NotEquals statements comparing it to other "pod ID" arguments.  Then when SELF appears in a "pod ID" argument, it's replaced with the pod ID when the statement is copied to another pod.
2. Change SELF to a value that is unlikely to show up by coincidence, and always replace SELF with the pod ID when copying custom statements.

dgulotta:
Closes #342 .

### DIFF
--- a/src/middleware/basetypes.rs
+++ b/src/middleware/basetypes.rs
@@ -33,10 +33,7 @@ use std::{
 
 use hex::{FromHex, FromHexError, ToHex};
 use plonky2::{
-    field::{
-        goldilocks_field::GoldilocksField,
-        types::{Field, PrimeField64},
-    },
+    field::types::{Field, PrimeField64},
     hash::poseidon::PoseidonHash,
     plonk::config::Hasher,
 };
@@ -58,12 +55,7 @@ pub const HASH_SIZE: usize = 4;
 pub const VALUE_SIZE: usize = 4;
 
 pub const EMPTY_VALUE: RawValue = RawValue([F::ZERO, F::ZERO, F::ZERO, F::ZERO]);
-pub const SELF_ID_HASH: Hash = Hash([
-    GoldilocksField(0x5),
-    GoldilocksField(0xe),
-    GoldilocksField(0x1),
-    GoldilocksField(0xf),
-]);
+pub const SELF_ID_HASH: Hash = Hash([F(0x5), F(0xe), F(0x1), F(0xf)]);
 pub const EMPTY_HASH: Hash = Hash([F::ZERO, F::ZERO, F::ZERO, F::ZERO]);
 
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
--- a/src/backends/plonky2/basetypes.rs
+++ b/src/backends/plonky2/basetypes.rs
@@ -2,16 +2,15 @@
 //! is enabled.
 //! See src/middleware/basetypes.rs for more details.
 
+/// F is the native field we use everywhere.  Currently it's Goldilocks from plonky2
+pub use plonky2::field::goldilocks_field::GoldilocksField as F;
 use plonky2::{
-    field::{extension::quadratic::QuadraticExtension, goldilocks_field::GoldilocksField},
+    field::extension::quadratic::QuadraticExtension,
     hash::{hash_types, poseidon::PoseidonHash},
     plonk::{circuit_builder, circuit_data, config::GenericConfig, proof},
 };
 use schemars::JsonSchema;
 use serde::{Deserialize, Deserializer, Serialize};
-
-/// F is the native field we use everywhere.  Currently it's Goldilocks from plonky2
-pub type F = GoldilocksField;
 
 /// D defines the extension degree of the field used in the Plonky2 proofs (quadratic extension).
 pub const D: usize = 2;
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -766,7 +766,7 @@ impl Params {
     }
 }
 
-/// Replace references to SELF by `self_id` in anchored keys of the statement.
+/// Replace references to SELF by `self_id`.
 pub fn normalize_statement(statement: &Statement, self_id: PodId) -> Statement {
     let predicate = statement.predicate();
     let args = statement
@@ -775,6 +775,9 @@ pub fn normalize_statement(statement: &Statement, self_id: PodId) -> Statement {
         .map(|sa| match &sa {
             StatementArg::Key(AnchoredKey { pod_id, key }) if *pod_id == SELF => {
                 StatementArg::Key(AnchoredKey::new(self_id, key.clone()))
+            }
+            StatementArg::Literal(value) if value.raw.0 == SELF.0 .0 => {
+                StatementArg::Literal(self_id.into())
             }
             _ => sa.clone(),
         })


robknight:
Are there any security implications to this? If I send someone a POD Request containing this magic value as a RawValue, will it get replaced with the POD ID?

I _think_ this means that clients, including the solver, would need to become aware of the effect of this constant. As it stands, this breaks the use-case of "run the request through the solver again, but with the response as the sole input", because now the request has a different value to the one returned in the response.

Might it be worth having a special TypedValue for this rather than relying on magic numbers? I suppose not, since the backend wouldn't know about it.

I'm still pretty skeptical of the use-case that motivates this, and of adding magic. In isolation each bit of magic seems reasonable, until it becomes hard to reason about what's going on.

dgulotta:
I don't think of #342 as being tied to a specific use case of self-referential pods.  A custom statement could take a pod ID, and be agnostic as to whether that ID is SELF.  Currently, the system will not produce a correct statement if the ID happens to be SELF.

robknight:
I don't think it's really incorrect - if I get a POD Request which says `REQUEST(some_custom_pred(1))` then the resulting statement will contain the number 1. This is what I would expect. Having a resulting statement which contains a different value to the one that was in the request is what would be surprising to me.

robknight:
Having thought a bit more about it: would it be possible to go back to the earlier solution of having SELF as a special-case wildcard, rather than a special-case literal?

Representing placeholder values as wildcards just seems much cleaner, and "wildcards in requests become literal values in responses" is an existing pattern, whereas "literal values can change to become other literal values" isn't.

src/middleware/basetypes.rs
```diff
@@ -33,7 +33,10 @@ use std::{
 
 use hex::{FromHex, FromHexError, ToHex};
 use plonky2::{
-    field::types::{Field, PrimeField64},
+    field::{
+        goldilocks_field::GoldilocksField,
```
ed255, line 37: We rename this type as `F` here https://github.com/0xPARC/pod2/blob/143a8c9d4eb43046ef1c24d4df0a7fc8b0fc647c/src/backends/plonky2/basetypes.rs#L14

And then re-export it in the middleware here https://github.com/0xPARC/pod2/blob/143a8c9d4eb43046ef1c24d4df0a7fc8b0fc647c/src/middleware/basetypes.rs#L49

To be consistent could you remove this import and rewrite the `SELF_ID_HASH` as `[F(0x5), F(0xe), ...`?

artwyman:
I think I'd need to have a deeper in-person conversation to fully understand the issue under discussion here, so take my thoughts as only loosely-held opinions.  But I've been following along and have some knee-jerk reactions.

1. Replacing a magic literal does feel risky to me, though I don't have a specific problematic case in mind.  If podlang already has a notion of wildcards which will be filled in by the prover that seems like a better conceptual fit here.  At least enough to be worth exploring.

2. I do think that it's useful for the prover of a MainPOD to have some way to identify themselves and "attest" to the contents of that MainPOD, without signing another SignedPOD to do it.  That's specifically so that copied statements don't have the same value as the original.  This is something we can discuss at the use case level, but I think it's an important option when taking PODs from "statement of fact" to "statement of intent".

PublicKeyOf is one way to get there.  Rob and I talked through some ways to get there using the RSAPOD or a known secret entry some other POD.  I think the construction right now requires a custom predicate in order to ensure the known value stays private, while confirming it came from the same MainPOD rather than being copied from another one.  A part of me wants this concept of "prover identity" to be more of a first-class notion in the language.

ed255:
> > > I think this means that clients, including the solver, would need to become aware of the effect of this constant. As it stands, I think this breaks the use-case of "run the request through the solver again, but with the response as the sole input", because now the request might stipulate a different value to the one returned in the response.
> 
> > Could you elaborate on that?
> 
> If I write `REQUEST(some_pred(SOME_LITERAL))` and send this to someone, then I want to be able to validate the response by running the same request locally, with the response POD as the sole input. But if the response contains a different literal, then it won't work.

I see, I understand that now.

This may be a bit of stretch but currently if you do `REQUEST(Equal(SELF["foo"], bar))` you'll encounter the same issue, as `SELF` will be replaced by the pod_id value.  Do you see this as problematic?

Technically you could simulate the environment of the prover (where you still have `SELF` instead of the `pod_id`) if you use `pod.pub_self_statements()` instead of `pod.pub_statements`.  So perhaps there's no technical blocker here, but more a design that is confusing / not intuitive?

> We already have a system for dealing with prover-supplied values, which is wildcards. `REQUEST(some_pred(?some_wildcard))` can come back with a prover-supplied value in the `some_pred` statement without causing a problem. This is a really neat division that is easy to reason about - literals are literals and never change, and wildcards are always "replaced" by literals during proving.

I don't see the `SELF` as competing against wildcards.  I see the SELF as a way to internally refer to the pod_id which you don't know yet.  And in order for pod_ids to be unique, this SELF must become unique after it appears in conjunction with statements from other pods (hence the replacement).

> > After writing this example I realize it conveys the idea, but it's weird. But I believe that once we have the PublicKeyOf statement, the Entries of a MainPod have new semantics. Because proving the knowledge of a public key in a MainPod can be seen as a way of signing with that key all the Entries of the pod. And then it's a bug if I can say that several statements are proved together in the same pod when in reality those statements have been proven independently.
> 
> This seems wrong to me. `PublicKeyOf` is not equivalent to a signature, for the reason that it doesn't say anything about the data that it "signs". You can freely copy the `PublicKeyOf` statement to any other MainPod, unless you only ever use it as as private statement. Aside from this, I'm not sure why we would _want_ to do this.
> 
> We already have a system for signing things - SignedPod! If you want to make a signature over some data, you can create a SignedPod to do it! If you want to sign a message saying something like "I endorse the contents of this MainPod", then you can make a SignedPod with an entry whose value is the hash of the MainPod. This also has the advantage of being much cheaper than making a new MainPod if you just want to prove that you control the keypair of a public key mentioned in the MainPod whilst also, say, signing a nonce to prevent credential reuse.

I understand what you say.  I also wonder what use-cases `PublicKeyOf` uncovers that can't be realized with SignedPods.
But I disagree with saying that "`PublicKeyOf` is not equivalent to a signature".  I think there's a way to use the `PublicKeyOf` such that the only way to make a MainPod is by knowing the a secret key corresponding to a public key without revealing the secret key.  Whether this construction is useful or not, I'm not sure because we already have SignedPods.  I would prefer moving the discussion about the usefulness of `PublicKeyOf` to another thread if possible.  I used it in my example for simplicity.  Something similar could be achieved with an introduction pod that derives a public key, or a statement shows knowledge of a hash pre-image without revealing it.  And of course this needs to be properly constructed such that the knowledge proof can't be isolated and reused.

> > > Having thought a bit more about it: would it be possible to go back to the earlier solution of having SELF as a special-case wildcard, rather than a special-case literal?
> 
> > I think that implies some kind of type checking in-circuit. You mean that a value used as pod_id to be placed as a wildcard for the pod_id of an anchored key is different than a value to be placed as a wildcard in a value slot right?
> 
> I'm just thinking of the special SELF wildcard, which is what we had before the literals-in-statements change. That wildcard can become the real POD ID during proving.

Ah, you mean when an anchored key in a statement template argument was something like `(SelfOrWildcard, KeyOrWildcard)`?

PR #347, opened 2025-07-18
chore(middleware): add container op checks and tests

ax0:


src/middleware/operation.rs
```diff
@@ -542,3 +559,79 @@ pub(crate) fn value_from_op(input_st: &Statement, output_ref: &ValueRef) -> Opti
...
+
+    use crate::{
+        backends::plonky2::primitives::merkletree::MerkleTree,
+        middleware::{
+            hash_value, AnchoredKey, Error, Key, Operation, Params, PodId, Result, Statement,
+        },
+    };
+
+    #[test]
+    fn check_container_ops() -> Result<()> {
```
ed255, line 575: Thanks for adding these tests!  They should be useful for the work that Andrew and Brian Liu are doing.

ed255 (approved):
LGTM!

ed255:
While reviewing another PR I just saw this:
https://github.com/0xPARC/pod2/blob/8429cd224d24538357e22bd09327904398fa7a1e/src/frontend/mod.rs#L426
https://github.com/0xPARC/pod2/blob/8429cd224d24538357e22bd09327904398fa7a1e/src/frontend/mod.rs#L436
Which are probably related to this PR

PR #353, opened 2025-07-23
Add Podlang pretty-printing

robknight:
This PR adds Podlang pretty-printing from the AST (the middleware data structures). This is useful for visualizing a POD which contains custom statements, which we want to do in the Podnet client, but this feels like a feature which ought to belong in the core language crate.

I think we might want to revisit how to pretty-print cross-batch predicate calls at some point; for now I'm assuming that all printing is happening in the context of a POD that already exists, and so we will just want to print out all of the batches that are used by the POD. However, there might be cases where we want to represent cross-batch calls with the `use` syntax, though I can't immediately think of what that would be.


src/lang/pretty_print.rs
```diff
@@ -0,0 +1,669 @@
...
+use crate::middleware::{
+    CustomPredicate, CustomPredicateBatch, NativePredicate, Predicate, StatementTmpl,
+    StatementTmplArg, TypedValue, Value,
+};
+
+/// Trait for converting AST nodes to Podlang source code
+///
+/// This trait provides a consistent interface for pretty-printing different
+/// types of AST nodes back to their Podlang source representation.
+pub trait PrettyPrint {
```
ed255, line 17: Have you considered changing the API of the trait methods to receive a mutable `Writer` instead of returning a `String`?  Like this:
```rust
fn fmt_podlang_with_indent(&self, w: &mut dyn fmt::Write, indent: usize) -> Result<(), fmt::Error>;
```

Then you don't need to allocate a string for each implementation and you can just continuously write to the `w` (as a stream)

src/lang/pretty_print.rs
```diff
@@ -0,0 +1,669 @@
...
+impl StatementTmpl {
+    fn to_podlang_string_with_batch_context(
+        &self,
+        batch_context: Option<&CustomPredicateBatch>,
+    ) -> String {
+        let mut result = String::new();
+
+        match &self.pred {
+            Predicate::Native(native_pred) => {
+                let _ = write!(&mut result, "{}", format_native_predicate(native_pred));
```
ed255, line 50: I guess this `let _ = ` is to capture the `Result` because otherwise clippy complains?
And this `write!` will always return Ok because the writer is a `&mut String`, and a string will never return error on a write.

src/lang/pretty_print.rs
ed255: I see you're already using the `write!` macro extensively, those calls would look the same but instead of writing to `&mut result` you'd write to `w`.
And whenever you have `format!(...)` you''ll have to change it for `write!(w, ...)`

src/lang/pretty_print.rs
```diff
@@ -0,0 +1,669 @@
...
+            }
+            let _ = write!(&mut result, "{}", arg.to_podlang_string());
+        }
+        let _ = write!(&mut result, ")");
+
+        result
+    }
+}
+
+impl PrettyPrint for StatementTmplArg {
```
ed255, line 81: I believe this implementation matches the `impl Display`:
https://github.com/0xPARC/pod2/blob/8429cd224d24538357e22bd09327904398fa7a1e/src/middleware/custom.rs#L108-L122

src/lang/pretty_print.rs
```diff
@@ -0,0 +1,669 @@
...
+}
+
+impl CustomPredicateBatch {
+    fn format_predicate_with_context(&self, predicate: &CustomPredicate, indent: usize) -> String {
+        format_predicate_definition(predicate, indent, Some(self))
+    }
+}
+
+impl PrettyPrint for Value {
+    fn to_podlang_string_with_indent(&self, _indent: usize) -> String {
```
ed255, line 122: There's some overlap between this implementation and `Display for TypedValue` (which is used in `Display for Value`).

The biggest changes are that in `Display` the containers are not shown fully, only their commitment is shown.

Then the syntax for `PodId, Raw, PublicKey` is different, and I think `Display for Value` would work well with the podlang syntax for those.

Perhaps we can just move this implementation to `Display for TypedValue`?  I think it's fine to show the entire contents of the containers.

src/lang/pretty_print.rs
```diff
@@ -0,0 +1,669 @@
...
+            if private_args.peek().is_some() {
+                let _ = write!(result, ", ");
+            }
+        }
+    }
+
+    let _ = write!(result, ")");
+}
+
+fn format_native_predicate(pred: &NativePredicate) -> &'static str {
```
ed255, line 238: Could you move this to an `impl Display for NativePredicate` in `src/middleware/statement.rs`?
This is useful beyond podlang.

ed255 (commented):
Overall looks good to me!

Please take a look at my comments.  Most of them are related to moving some of this formatting code outside of `lang` to reduce code deduplication and because we don't have any commitment for the format in the `Display` implementations, so it's fine if we change it to make it more podlang-y.

Also consider my suggestion to receive a writer instead of returning a string.

src/lang/pretty_print.rs
robknight: This is eliminated due to the writer change.

src/lang/pretty_print.rs
robknight: It does. I'm not sure what to do about this - there are three options:

1) Replace the existing `Display` code for all relevant types with the implementations in `pretty_print.rs`, and mostly remove `pretty_print.rs` (except the tests)
2) Have the `Display` implementations call the `PrettyPrint` methods
3) Have two separate implementations, which can sometimes display different things

The argument for (3) is that there might be cases where we want `Display` to work differently from how `PrettyPrint` works (e.g. some types have an "alternate" display, whereas the pretty-printer can only output one canonical form).

If we use `Display` then I will probably need to add a lot more tests (or property-based tests) to ensure that display changes don't break the round-trip through the parser.

src/lang/pretty_print.rs
robknight: I've moved this to `TypedValue`.

src/lang/pretty_print.rs
robknight: Done

PR #359, opened 2025-07-28
Check a single POD against a POD Request

robknight:
To date, most of the tests in our repo do not check that the PODs contain their expected statements after building. They are typically verified using the `verify` method, but their content is not inspected. This is not unreasonable for some of the tests, as their purpose is to test specific features. However, it may be useful for the frontend tests in particular to demonstrate how to check the contents of a POD, for the benefit of third-party developers.

The test for the solver in the pod2-client repo do validate the contents of created PODs, since the entire purpose of the solver is pattern-matching on the contents of PODs. In the "document server" use-case, we use the solver as a mechanism of validating incoming PODs. This sort-of works, but has two wrinkles:

1) The solver is a bit heavy if all you really want to do is validate that one POD contains the statements that you expect
2) The solver checks that the input PODs logically support the statements you want to make, which most of the time means "contains exactly the statements you want to make". But if you want to make a statement like `Lt(?pod["some_key"], 50)`, and the response POD contains the statement `Equal(?pod["some_key"], 10)`, then this will be treated as a match because, logically, it does state the thing you want to prove. But if you want to check that the POD contains _exactly_ the narrow statement `Lt(?pod["some_key"], 50)` then the solver won't guarantee that. I personally quite like this, as it is entirely _logical_, but that doesn't prevent it from being a bit surprising.

Given this, I've written a simpler system for checking if a given POD contains exactly the statements specified in a `PodRequest`, with consistent wildcard bindings. This is a simple DFS with backtracking, which is fine because we're only doing syntactic matching rather than handling the semantics of the statements as the solver does, and only with a single POD as input.

I've re-jigged the ZuKYC tests a bit, and these now test that the ZuKYC POD really does contain the expected statements. I've also done the same for Ethdos - it's fairly trivial as each POD only contains one `eth_dos` statement, but it demonstrates the principle. I also switched the ZuKYC PODs test back to the approach of having the sanctions list as a literal provided by the requestor, rather than relying simply on a signature from a third party. This aligns with the original GPC version, and makes the tests and PodRequest simpler to read.

To validate the contents of a POD, you can create a `PodRequest` like so:
```rust
let request = parse(
      r#"
  REQUEST(
      NotEqual(4, 5)
      // Add statement templates here
  )
  "#,
      &params,
      &[],
  )
  .unwrap()
  .request;
```
This creates a `PodRequest` struct, which has a method `exact_match_pod`, which takes a single `&dyn Pod` and returns true if all of the statement templates in the request can be satisfied by statements in the `Pod`, or false otherwise. Additional statements not mentioned in the request are ignored.
```rust
request.exact_match_pod(pod); // returns true if all statement templates matched, false if not
```

I think this ought to be sufficient as a simple and general-purpose mechanism for checking the content of a POD.

### DIFF
--- a/src/frontend/pod_request.rs
+++ b/src/frontend/pod_request.rs
@@ -6,14 +6,14 @@ use crate::{
     middleware::{Pod, Statement, StatementArg, StatementTmpl, StatementTmplArg, Value},
 };
 
+/// Represents a request for a POD, in terms of a set of statement templates.
+/// The response should be a POD (or PODs) containing a set of statements which
+/// satisfy the templates, with consistent wildcard bindings across all templates.
 #[derive(Debug, Clone, PartialEq)]
 pub struct PodRequest {
     pub request_templates: Vec<StatementTmpl>,
 }
 
-/// Represents a request for a POD, in terms of a set of statement templates.
-/// The response should be a POD (or PODs) containing a set of statements which
-/// satisfy the templates, with consistent wildcard bindings across all templates.
 impl PodRequest {
     pub fn new(request_templates: Vec<StatementTmpl>) -> Self {
         Self { request_templates }
@@ -23,9 +23,6 @@ impl PodRequest {
     /// This checks for exact matches to the statement templates; that is to say
     /// that it performs a "syntactic" match, not a "semantic" match; no
     /// processing of the semantics of the statements is performed.
-    ///
-    /// TODO: matching on SELF as part of https://github.com/0xPARC/pod2/issues/351
-    /// Until then, matching on SELF anchored keys is not supported.
     pub fn exact_match_pod(&self, pod: &dyn Pod) -> Result<HashMap<String, Value>> {
         let pod_statements = pod.pub_statements();
         let mut bindings: HashMap<String, Value> = HashMap::new();


artwyman:
This looks great.  I skimmed the implementation, but will defer to someone else who knows Rust better to review fully.

One request from a usability standpoint: can the matcher return the bindings?  As a verifier the next thing I'm going to want to do after checking the "shape" of the result is to look at any revealed values, which in this case will be the wild card.  E.g. prove you have a ticket matching various criteria, and reveal your email address, in which case the "reveal" part is a template like `Eq($ticket_pod_id["attendee_email"], $email)`, and I want to know what `email` bound to.

robknight:
That's a good idea - it makes the test more complete too, as we can check if the bindings are what we expected.

src/frontend/mod.rs
```diff
@@ -904,7 +906,11 @@ pub mod tests {
             gov_id.get("_signer").unwrap(),
             pay_stub.get("_signer").unwrap(),
         )?;
-        assert!(request.exact_match_pod(&*kyc.pod));
+        assert!(request.exact_match_pod(&*kyc.pod).is_ok());
```
artwyman, line 909: This pattern is executing exact_match_pod twice.  I think the unwrap() below is sufficient to cover this assert.
Same pattern repeats below.

artwyman (commented):
👍

src/frontend/pod_request.rs
```diff
@@ -0,0 +1,260 @@
...
+};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct PodRequest {
+    pub request_templates: Vec<StatementTmpl>,
+}
+
+/// Represents a request for a POD, in terms of a set of statement templates.
+/// The response should be a POD (or PODs) containing a set of statements which
+/// satisfy the templates, with consistent wildcard bindings across all templates.
```
ed255, line 16: Shouldn't this doc string be above the definition of the struct?

ed255 (approved):
LGTM!
The checker implementation is very clean :)

I made a small comment, and I agree with @artwyman's nitpick here https://github.com/0xPARC/pod2/pull/359#discussion_r2238319592

PR #365, opened 2025-07-29
Secret keys in Podlang

robknight:
This implements the SecretKey literal in Podlang, as follows:

```
SecretKey(base_64_encoded_string)
```

This follows the base64 encoding used for serialization of SecretKeys.


ed255 (approved):
LGTM!

PR #366, opened 2025-07-29
remove unsafe code in cache

dgulotta:
This PR removes the unsafe code in `pod2::cache::mem` by using `Box::leak` instead.

Although it is unlikely to matter for our use case, I believe the current cache code is unsound in general because it does not require `T` to be `Sync`.  If `T` is not `Sync` (for example, if `T` is a `RefCell`), then it should not be possible to get a reference to `T` that outlives the lock on the `Mutex`.

Separately, I replaced an instance of `String::from_utf8_unchecked` in `DebugGate` because there did not seem to be a good reason not to check.

### DIFF
--- a/src/backends/plonky2/circuits/utils.rs
+++ b/src/backends/plonky2/circuits/utils.rs
@@ -7,7 +7,7 @@ use plonky2::{
         witness::{PartitionWitness, Witness},
     },
     plonk::circuit_data::CommonCircuitData,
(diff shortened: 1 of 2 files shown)


ed255 (approved):
LGTM!

Thanks for improving these two previous usages of `unsafe`!

I think you're right in that the usage of `unsafe` in the cache was unsound, so thanks again for fixing that :)

PR #367, opened 2025-07-29
support longer arrays in vec_ref

ed255:


dgulotta:
Could you explain why `high` is a `Vec<BoolTarget>` instead of a single `Target`?  Otherwise, it seems fine.

ed255:
> Could you explain why `high` is a `Vec<BoolTarget>` instead of a single `Target`? Otherwise, it seems fine.

Thanks for asking this question!  I now realized that a single `Target` just works, so I'll simplify the implementation with this change.

ed255:
@dgulotta I've replaced the bit decomposition of the `high` part with a single `Target`, could you take a look again?

src/backends/plonky2/circuits/common.rs
```diff
@@ -867,6 +873,50 @@ impl Flattenable for StatementTmplArgTarget {
...
+        2
+    }
+    pub fn new_virtual(array_len: usize, builder: &mut CircuitBuilder) -> Self {
+        // Limit the maximum array length to avoid abusing `vec_ref`
+        assert!(array_len <= 256);
+        let bits = usize_bits(array_len);
+        Self {
+            bits,
+            low: builder.add_virtual_target(),
+            high: if bits > 6 {
```
dgulotta, line 904: Seems like this can be simplified by checking if `array_len <= 64`.

src/backends/plonky2/circuits/common.rs
```diff
@@ -1293,18 +1351,41 @@ impl CircuitBuilderPod<F, D> for CircuitBuilder {
...
+        // We do several random accesses over chunks of CHUNK_LEN using the lowest bits of the
+        // index.  Then we combine them using the highest bits of the index.
+        let mut chunk_res = Vec::new();
+        let num_chunks = array.len().div_ceil(CHUNK_LEN);
+        for chunk in array.chunks(CHUNK_LEN) {
+            let mut index_chunk = i.low;
+            // I we have several chunks and the last one is smaller (it's index needs less than 6
+            // bits), make it zero except when it's used so that the range check over the index
+            // passes.
+            if chunk.len() <= CHUNK_LEN / 2 && i.bits > 6 {
```
dgulotta, line 1370: `i.bits > 6` can be replaced with `num_chunks > 1`

dgulotta (approved):
There are a few very small suggestions below, but otherwise it looks good!

PR #368, opened 2025-07-29
Support SELF keyword in Podlang parser

### LINKED ISSUE #351: Syntactic sugar for magic 0x5e1f constant
In Podlang, we want to be able to write `SELF`, but have this be substituted for a magic constant at runtime.

The aim is to make it easy for frontend-aware clients to understand that the user's intent is to refer to the SELF PodId, which might not be clear from the use of a `Raw()` value.

Roughly speaking, given this POD Request:

```
REQUEST(
  Equal(?pod_id, Raw(0x5e1f))
  Equal(?pod_id["key"], 123)
)
```

If the response contains this:

```
Equal(0x123456, 0x123456)
Equal(0x123456["key"], 123)
```

Then the client is entitled to reject the response as the literal values do not match.

However, given:

```
REQUEST(
  Equal(?pod_id, SELF)
  Equal(?pod_id["key"], 123)
)
```

Then the client should accept the response as SELF is a special value which should match against any PodId.

This issue depends on https://github.com/0xPARC/pod2/pull/345 to define the magic constant.

robknight:
Closes #351 

This adds support for a syntactic sugar SELF keyword.

I've simplified my original design, so that this is pure syntactic sugar that gets removed during parsing. It's simply a shortcut for writing out the full SELF keyword in hex.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -266,7 +266,11 @@ impl fmt::Display for TypedValue {
             TypedValue::PublicKey(p) => write!(f, "PublicKey({})", p),
             TypedValue::SecretKey(p) => write!(f, "SecretKey({})", p),
             TypedValue::PodId(p) => {
-                write!(f, "0x{}", p.0.encode_hex::<String>())
+                if *p == SELF {
+                    write!(f, "SELF")
+                } else {
+                    write!(f, "0x{}", p.0.encode_hex::<String>())
+                }
             }
             TypedValue::Raw(r) => {
                 write!(f, "Raw(0x{})", r.encode_hex::<String>())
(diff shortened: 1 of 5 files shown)


src/middleware/mod.rs
```diff
@@ -341,6 +341,18 @@ impl JsonSchema for TypedValue {
             ..Default::default()
         };
 
+        let pod_id_schema = schemars::schema::SchemaObject {
```
robknight, line 344: This is unrelated but I noticed that this is missing during testing. The JSON schema is used for validation and TypeScript code generation in the Podnet client, and I noticed that it wasn't rendering PodIds correctly because this is missing.

PR #369, opened 2025-07-29
fix: missing gate and generator for serialization

ed255:
Add the missing gates and generator in the serializer that were added with the PublicKeyOf operation.

Add a test for CircuitData serialization+deserialization to avoid these kind of bugs in the future.


ax0 (approved):
LGTM!

PR #370, opened 2025-07-30
fix frontend::Operation::new_entry + doc improvements

dgulotta:
This PR fixes the `frontend::operation::Operation::NewEntry` function, which previously created an operation with the wrong number of arguments.  It also makes a few improvements to error handling / documentation:
* In `frontend::MainPodBuilder::op_statement`, `invalid_op_args` errors now always include the operation type.  Previously, they would sometimes have "malformed operation" where the operation type was supposed to be.
* `frontend::MainPodBuilder::op_statement` now checks that statements that appear as an argument to a custom predicate have the correct predicate and correct number of arguments.
* `frontend::Operation::custom` has documentation explaining that `args` should be the the set of statements required to prove the custom statement, and `Statement::None` should be used with `or` predicates.
* `middleware::custom::CustomPredicate::new` has documentation explaining its arguments.  In particular, it clarifies that `args_len` is the number of public arguments while `wildcard_names` should include public and private arguments.


ed255 (approved):
LGTM!

PR #372, opened 2025-07-30
Return an error instead of panicking when too many statements

### LINKED ISSUE #371: Error handling for MainPodBuilder
In `MainPodBuilder` there are some cases where we panic instead of returning an error when some parameter limit is exceeded. For example:

https://github.com/0xPARC/pod2/blob/f10a5adb41474b81e774c7c7999f5a0740514734/src/frontend/mod.rs#L176-L183

Any method on `MainPodBuilder` that currently might panic should instead return a `Result` with a proper error type. As `MainPodBuilder` is a frontend component, the error types and messages need to be clear and helpful.

robknight:
Closes #371 - instead of panicking when there are too many statements, instead return a Result with an error.

### DIFF
--- a/src/frontend/error.rs
+++ b/src/frontend/error.rs
@@ -35,6 +35,10 @@ pub enum InnerError {
     PodlangParse(String),
     #[error("POD Request validation error: {0}")]
     PodRequestValidation(String),
+    #[error("too many public statements")]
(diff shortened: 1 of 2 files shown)


ed255 (approved):
LGTM!

Issue #361, opened 2025-07-28
Support index for secret key coming from an input POD


artwyman:
After #355 lands, PublicKeyOf statements will be supported, but with a known limitation in the code which fills in the secret key witness in MainPodVerifyTarget::set_targets().  The calculation of `adjusted_index` and the resultant lookup only works if the secret key literal came from a statement in the current POD.  This is the likely use case, given secret keys should generally not be public.  However this code will fail (crash due to a negative index into a vector) if the secret key ever came from a statement in an input pod (signed or recursive).  See this TODO in the code:

```
                    // TODO: This adjustment only works if the secret key came
                    // from a statement in the current POD, which is the most
                    // common case.  A more general solution needs to be able
                    // index across the virtual array of statements from all
                    // input PODs, similar to what's done in
                    // plonky2::mainpod::layout_statements.
```

ed255:
I would like to propose solving this issue once we introduce the table approach for the public key derivation.

In that approach, we can extract the secret keys easily in a function similar to this one
https://github.com/0xPARC/pod2/blob/9f8335756c00e524247d90ea7672f4364a0a7312/src/backends/plonky2/mainpod/mod.rs#L128-L177
which sets up the inputs to the circuit and is being called before reaching the `set_targets` method.

artwyman:
> I would like to propose solving this issue once we introduce the table approach for the public key derivation.

Makes sense.  So this issue should be revisited when #357.  (Mentioning that issue here will hopefully help that to happen.)

Issue #357, opened 2025-07-25
Multiplexed table lookup in the MainPod for operation checks


ed255:
Some operation checks are quite expensive in terms of constraints, so doing them for each operation slot in the MainPod is very costly overall.  For this reason in those situations we choose a smaller number of `k` checks and do them outside of the operation slots.  Then in the operation slot we just index the result of the outside check via a `random_access`.

There's an improvement over this solution when the table entries we're accessing are very big in terms of `Target`s: instead of doing the `random_access` to each `Target` we hash all the `Targets` of a table entry and do `random_access` over this hash.

Currently we have two operation checks that use table lookups with `random_access`:
- table of `MerkleClaimTarget`s: ContainsFromEntries, NotContainsFromEntries
- table of `CustomPredicateVerifyQueryTarget`s: Custom

In the following weeks we'll have a new table for the `MerkleProofStateTransitionTarget` that will be used by `ContainerUpdateFromEntries`.
And for https://github.com/0xPARC/pod2/pull/355 we'll most likely need a table of `hash(pk), hash(sk)` generated from verifications of `pk = derive(sk)`.

So that's 4 tables in total, and a `CircuitBuilderPod::vec_ref` for each, which internally uses `random_access`.

The `random_access` method uses gates that can do random access to arrays of size power of two up to 64 elements.  With https://github.com/0xPARC/pod2/issues/356 we will be able to do more than 64 (perhaps the new implementation will support array sizes of sums of a few powers of two like 64 + 16, 64 + 32, 64 + 64, 2*64 + 16, ...)

To get a better utilization of the arrays we're using for `random_access` and reduce a few `Target`s overhead per table access, I propose multiplexing all tables into one.

This is what the table could look like:
| tag | data | hash |
| ---     | ---     | --- |
| MerkleProof | MerkleClaimTarget | hash(tag \| data) |
| CustomPredVerify | CustomPredicateVerifyQueryTarget | hash(tag \| data) |
| MerkleProofStateTransition | MerkleProofStateTransitionTarget | hash(tag \| data) |
| PublicKeyDerive | (hash_pk, hash_sk) | hash(tag \| data) |

The tag column would be a fixed value.
The data column would have targets that are used in constraints that verify their correctness unconditionally.
The hash column would have a hash of of the tag and data.  The goal of this column is to compact all table entries into just 4 `Target`s, and we'd do the `vec_ref` over the hash.

[...]

PR #376, opened 2025-07-31
Add table multiplexer (and use it for container, custom pred & PublicKeyOf ops)

ed255:
- Extend the `Flattenable` trait to include a `size` method that returns the number of `Target`s the type requires.  This is used in the table to figure out the max length of an array that must fit all entry types.
- Move the circuit methods to precalculate hash states and do hashes started from a precomputed state to a new module
- Introduce `MuxTableTarget` which allows easy multiplexing of tables where each sub-table may have entries of different lengths.  The table access is done via hashing + unhashing automatically (via use of a generator)
- Use the `MuxTableTarget` to access merkle tree claims and custom predicate verification, which where previously in different tables and accessed with independent random accesses each
- Move the public key derivation for the PublicKeyOf operation check to the same multiplexed table.  Now we can choose how many of those operations a circuit supports.

Resolve https://github.com/0xPARC/pod2/issues/357
Resolve https://github.com/0xPARC/pod2/issues/361

### DIFF
--- a/src/backends/plonky2/circuits/mainpod.rs
+++ b/src/backends/plonky2/circuits/mainpod.rs
@@ -95,8 +95,7 @@ impl StatementCache {
             // converting a length 1 array into a scalar.
             op.args
                 .iter()
-                .flatten()
-                .map(|&i| builder.vec_ref(params, prev_statements, i))
+                .map(|i| builder.vec_ref(params, prev_statements, i))
                 .collect::<Vec<_>>()
         };
         assert!(params.max_operation_args >= 3);
@@ -224,7 +223,7 @@ fn verify_operation_circuit(
     // been verified, so we need only look up the claim.
     let measure_resolve_merkle_claim = measure_gates_begin!(builder, "ResolveMerkleClaim");
     let resolved_merkle_claim =
-        (!merkle_claims.is_empty()).then(|| builder.vec_ref(params, merkle_claims, op.aux[0]));
+        (!merkle_claims.is_empty()).then(|| builder.vec_ref(params, merkle_claims, &op.aux[0]));
     measure_gates_end!(builder, measure_resolve_merkle_claim);
 
     // Operations from custom statements will refer to one
@@ -233,7 +232,7 @@ fn verify_operation_circuit(
     let measure_resolve_custom_pred_verification =
         measure_gates_begin!(builder, "ResolveCustomPredVerification");
     let resolved_custom_pred_verification = (!custom_predicate_verification_table.is_empty())
-        .then(|| builder.vec_ref(params, custom_predicate_verification_table, op.aux[1]));
+        .then(|| builder.vec_ref(params, custom_predicate_verification_table, &op.aux[1]));
     measure_gates_end!(builder, measure_resolve_custom_pred_verification);
 
     // The verification may require aux data which needs to be stored in the
@@ -938,7 +937,7 @@ fn make_statement_arg_from_template_circuit(
     let first_index = ak_id_wc_index;
     let is_first_index_valid = builder.or(is_ak, is_wc_literal);
     let first_index = builder.select(is_first_index_valid, first_index, zero);
-    let resolved_ak_id = builder.vec_ref(params, &args, first_index);
+    let resolved_ak_id = builder.vec_ref_small(params, &args, first_index);
     let resolved_wc = resolved_ak_id;
 
     // If the index is not used, use a 0 instead to still pass the range constraints from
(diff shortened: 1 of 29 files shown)


src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1078,53 +1184,6 @@ fn normalize_statement_circuit(
     }
 }
 
-/// Precompute the hash state by absorbing all full chunks from `inputs` and return the reminder
```
ed255, line 1081: moved to `hash.rs`

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1078,53 +1184,6 @@ fn normalize_statement_circuit(
...
-    // Absorb all inputs up to the biggest multiple of RATE.
-    for input_chunk in inputs.chunks(P::RATE) {
-        perm.set_from_slice(input_chunk, 0);
-        perm.permute();
-    }
-
-    (perm, inputs_rem)
-}
-
-/// Hash `inputs` starting from a circuit-constant `perm` state.
```
ed255, line 1096: moved to `hash.rs`

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1216,50 +1275,6 @@ fn build_custom_predicate_table_circuit(
     Ok(custom_predicate_table)
 }
 
-/// Build table of [batch_id, custom_predicate_index, custom_predicate, args, st, op, op_args]
```
ed255, line 1219: removed in favor of `build_operation_aux_table_circuit`

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1612,41 +1617,6 @@ impl InnerCircuit for MainPodVerifyTarget {
         for (i, (st, op)) in zip_eq(&input.statements, &input.operations).enumerate() {
             self.input_statements[i].set_targets(pw, &self.params, st)?;
             self.operations[i].set_targets(pw, &self.params, op)?;
-            if matches!(
```
ed255, line 1615: This logic is now handled in `src/backends/plonky2/mainpod/mod.rs` `extract_public_key_of`

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1752,24 +1734,29 @@ mod tests {
...
             .collect();
+
         let merkle_proofs_target: Vec<_> = merkle_proofs
             .iter()
             .map(|_| {
-                let mt_proof = MerkleClaimAndProofTarget::new_virtual(
-                    params.max_depth_mt_containers,
-                    &mut builder,
-                );
-                verify_merkle_proof_circuit(&mut builder, &mt_proof);
```
ed255, line 1762: Removed this verification because it's also happening later via `build_operation_aux_table_circuit`

dgulotta:
Could you explain the motivation for putting multiple types of data in the same table?  Does this make some step in the verification process more efficient?

ed255:
> Could you explain the motivation for putting multiple types of data in the same table? Does this make some step in the verification process more efficient?

The motivation behind that is to reduce the overall number of constraints.

The table entries are accessed via random access, and we're using the plonky2 random access gates which works with powers of 2: in a single row you can do 1 random access of 64 elements, 2 random access of 32 elements, 4 random access of 16 elements, etc.

By merging multiple tables we can get better utilization.  Let's say we have 3 tables of 10.  Previously we'd need 3 accesses of 16 elements each (that take 16 * 3 = 48 "random access slots").  But if we merge them we end up with 30 elements that can use single a random access of 32 elements (that take 32 "random access slots").

The other benefit is the hashing+unhashing.  For random access to data that is big enough from an array that is big enough, it's cheaper to do hash + 4 random access + hash than doing n random accesses.  In this scenario, by merging multiple tables into one we only pay the access hash once: we just has an entire "flattened" entry and then interpret it as different Target types as necessary.  Currently we have some entries that are somehow big (I think `CustomPredicateVerifyQueryTarget` is over 100 `Target`s).
Also note that to calculate the hashed entry of the table we only pay for the entry size of that subtable, not the max entry size of the entire table.  This is because we follow a padding strategy that allows precomputing the hash state corresponding to the padding.

Please let me know if this makes sense, if something is not well explained, if you see any flaw in this explanation, etc.

dgulotta:
It's not clear to me why putting Merkle proof verifications, custom predicate verifications, and PublicKeyOf verifications in the same table is reducing the number of queries that need to be made.  If seems like that is replacing 3 accesses of 16 elements with 3 accesses of 32 elements, rather than 1 access of 32 elements.  But maybe I'm just missing something.

PR #378, opened 2025-08-01
remove op! macro

### LINKED ISSUE #352: Replace the `op!` macro for a more rusty solution
This `op!` macro has been confusing to some users: https://github.com/0xPARC/pod2/blob/89dfc4e214144ef71b4d1a51e4f5d382c9f0cbe7/src/frontend/mod.rs#L769-L827

One reason behind the confusion is that the first argument of the macro is not a Rust expression but a special token that doesn't exist in the Rust language (that is matched by the macro pattern matcher).

I think we could get away with this big macro by having functions to construct each operation where each argument is a different generic.  Something like this:
```rust
fn eq(lhs: impl Into<frontend::OperationArg>, rhs: impl Into<frontend::OperationArg>) -> frontend::Operation {
  frontend::Operation(
    middleware::OperationType::Native(middleware::NativeOperation::EqualFromEntries),
    vec![lhs.into(), rhs.into()]
  )
}
```

The only case where we may still need a macro is for a custom predicate operation, because the number of arguments is variadic.

dgulotta:
Fixes #352 .

### DIFF
--- a/examples/main_pod_points.rs
+++ b/examples/main_pod_points.rs
@@ -15,10 +15,9 @@ use pod2::{
         basetypes::DEFAULT_VD_SET, mainpod::Prover, mock::mainpod::MockProver,
         primitives::ec::schnorr::SecretKey, signedpod::Signer,
     },
-    frontend::{MainPodBuilder, SignedPodBuilder},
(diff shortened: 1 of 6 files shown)


ed255 (approved):
Nice! This is much better than the macro :)

PR #379, opened 2025-08-01
add missing constructors to frontend::Operation

dgulotta:
Added the `frontend::Operation` constructors that were missing from #378.  Also changed some of the constructors to take `Statement` instead of `&Statement`, since they were just cloning their arguments.

### DIFF
--- a/src/frontend/operation.rs
+++ b/src/frontend/operation.rs
@@ -143,7 +143,7 @@ macro_rules! op_impl_oa {
 
 macro_rules! op_impl_st {
     ($fn_name: ident, $op_name: ident, 1) => {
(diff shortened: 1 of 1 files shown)


ed255 (approved):
LGTM!

Although I have a comment: the `op_impl_st` with 2 arguments (at line 155) takes the statements as ref.  Perhaps it can take them without ref to be consistent with the case of 1 argument?

dgulotta:
> LGTM!
> 
> Although I have a comment: the `op_impl_st` with 2 arguments (at line 155) takes the statements as ref. Perhaps it can take them without ref to be consistent with the case of 1 argument?

Thanks for catching that!

PR #383, opened 2025-08-04
feat(backend): add updates and deletions to Merkle tree state transition proofs

### LINKED ISSUE #335: implement merkletree circuit to verify tree state transition
Implement merkletree update circuit, which proves that the merkletree state transition (from `old_root` to `new_root`) has been done correctly for the given `new_key`. This will allow verifying correct new leaf insertion, and leaf edition&deletion (if needed).

Related to https://github.com/0xPARC/pod2/issues/259 .

Edit: this will require implementing also the methods `insert`, `update`, `remove` leaf; which return the merkle proof of state transition that can be verified by the circuit.

ax0:
Resolves #335.

### DIFF
--- a/src/backends/plonky2/primitives/merkletree/circuit.rs
+++ b/src/backends/plonky2/primitives/merkletree/circuit.rs
@@ -30,7 +30,9 @@ use crate::{
         basetypes::D,
         circuits::common::{CircuitBuilderPod, ValueTarget},
         error::Result,
-        primitives::merkletree::{MerkleClaimAndProof, MerkleProofStateTransition},
+        primitives::merkletree::{
+            MerkleClaimAndProof, MerkleTreeOp, MerkleTreeStateTransitionProof,
+        },
     },
     measure_gates_begin, measure_gates_end,
     middleware::{EMPTY_HASH, EMPTY_VALUE, F, HASH_SIZE},
@@ -396,19 +398,18 @@ fn kv_hash_target(
 /// Verifies that the merkletree state transition (from old_root to new_root)
 /// has been done correctly for the given new_key. This will allow verifying
 /// correct new leaf insertion, and leaf edition&deletion (if needed).
-pub struct MerkleProofStateTransitionTarget {
+/// See `MerkleTreeStateTransitionProof` struct for an explanation of the fields.
+pub struct MerkleTreeStateTransitionProofTarget {
     pub(crate) max_depth: usize,
     // `enabled` determines if the merkleproof state transition verification is enabled
     pub(crate) enabled: BoolTarget,
-    pub(crate) typ: Target,
-
+    pub(crate) op: Target,
     pub(crate) old_root: HashOutTarget,
-    pub(crate) proof_non_existence: MerkleClaimAndProofTarget,
-
+    pub(crate) op_proof: MerkleClaimAndProofTarget,
     pub(crate) new_root: HashOutTarget,
-    pub(crate) new_key: ValueTarget,
-    pub(crate) new_value: ValueTarget,
-    pub(crate) new_siblings: Vec<HashOutTarget>,
+    pub(crate) op_key: ValueTarget,
+    pub(crate) op_value: ValueTarget,
+    pub(crate) siblings: Vec<HashOutTarget>,
 
     // auxiliary witness
     pub(crate) divergence_level: Target,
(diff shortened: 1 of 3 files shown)


arnaucube (approved):
LGTM! Very cool! also, thanks for the in-line comments, made the reasoning more easy to follow ^^

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -82,30 +76,83 @@ impl MerkleTree {
...
+
+        let old_root: Hash = self.root.hash();
+        self.root
+            .apply_op(self.max_depth, MerkleTreeOp::Update, *key, Some(*value))?;
+        let new_root = self.root.compute_hash();
+
+        let (v, proof) = self.prove(key)?;
+        assert!(proof.existence);
+        assert_eq!(v, *value);
+        assert!(proof.other_leaf.is_none());
```
ed255, line 119: Does `apply_op` with `Update` guarantee that we the entry at `key` exists?  which also implies that this `proof.other_leaf.is_none() = true`?
Also, `proof.siblings == old_proof.siblings` right?

I see that all the operations follow this pattern of making another proof after the `apply_op`.  This is not a suggestion for any change, I just want to confirm my understanding of the behavior.

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -82,30 +76,83 @@ impl MerkleTree {
...
+
+        Ok(MerkleTreeStateTransitionProof {
+            op: MerkleTreeOp::Delete,
+            old_root,
+            op_proof: proof,
+            new_root,
+            op_key: *key,
+            op_value: value,
+            value: None,
+            siblings: proof_existence.siblings,
```
ed255, line 152: In the other operations `op_proof` contains the proof before the operation, and `siblings` the siblings after the operation.  Why is this different?
Then I see that in `verify_state_transition` for `delete` you call `insert` just swapping the `old_root` and `new_root`, so you end up with something that is correct.

Wouldn't it be more consistent if
- `fn delete` returns `op_proof: proof_existence` and `siblings: proof.siblings`
- In `verify_state_transition` for the delete case you rearrange the `op_proof` and `siblings`

ed255: Update: I just saw that this behavior is documented in the `MerkleTreeStateTransitionProof` type, so I think this is fine :)

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -564,53 +681,166 @@ impl Node {
...
+        // path.
+        match op {
+            MerkleTreeOp::Delete => self.normalise_path(&key_path),
+            _ => Ok(()),
+        }
+    }
+
+    /// Normalises a Merkle tree along a specified path. Useful
+    /// post-deletion.
+    fn normalise_path(&mut self, key_path: &[bool]) -> TreeResult<()> {
```
ed255, line 723: nitpick: this function always returns `Ok(())`, so maybe there's no need to return `TreeResult<()>`?

src/backends/plonky2/primitives/merkletree/circuit.rs
```diff
@@ -396,105 +398,136 @@ fn kv_hash_target(
...
+
+    // 1) Verify the provided op proof.
+    verify_merkle_proof_circuit(builder, &proof.op_proof);
+
+    // 2) Check that the provided siblings yield an existence proof of
+    // (op_key, op_value) with root specified by op:
+    // Insert: `new_root`
+    // Update: `new_root`
+    // Delete: `old_root`
+    let root = HashOutTarget {
```
ed255, line 448: in `src/backends/plonky2/circuits/common.rs` we have the method `select_flattenable` and also `impl Flattenable for HashOutTarget`.
This would allow implementing this as:
```rust
let root = builder.select_flattenable(is_deletion, proof.old_root, proof.new_root)
```

I understand you may not want to import `common` because it has many pod-isms.  Would it make sense to split `common` into pod-isms and non-pod-isms to be used around?

src/backends/plonky2/primitives/merkletree/mod.rs
ed255: Update2: After reviewing the circuit implementation I'm even more convinced that this is the way to go.  The circuit treats the siblings/proof "differently" for the delete in order to skip constraints.  By having the native implementation like this we get a native implementation that matches the behavior of the circuits.  I think it's more important that the native implementation matches the circuits than having the consistency I was mentioning in the first message.

ed255 (approved):
LGTM! I've left a few comments, questions/discussions.

PR #396, opened 2025-08-25
Serialize Rust data structures to / from pods

dgulotta:
This PR allows Rust data structures to be serialized to / from pods.  There are some example tests in `middleware/serialization.rs`.

The implementation seems good enough to cover most use cases, but there are few ways that it could be improved:
- Ideally, serializing a `Value` would return the same `Value`, and serializing any of  `RawValue`, `Point`, `SecretKey`, `Dictionary`, `Set`, `Array` would just wrap it in a `TypedValue` in the obvious way.  This behavior is implemented for  `RawValue`, `Point`, and `SecretKey`, but not for `Dictionary`, `Set`, `Array` and `Value`.  Implementing the remaining four would be feasible, but the code would be ugly.
- If the user happens to create their own newtype struct named `RawValue`, `Point`, or `SecretKey`, then the serializer will think that it is one of the POD types.  We could fix this by having our types pass custom names to `Serializer::serialize_newtype_struct`.
- It would be convenient to have a serializer and deserializer for `Dictionary`, and maybe having deserializers for owned types in addition to reference types.
- If the deserializer receives a request for a specific data type, and it can't fulfill that request, it will send whatever it has and let the `Deserialize` implementation decide if it can do anything with that.  It seems customary to return an error instead.

Where possible, I tried to mimic `serde_json` behavior, but a few things had to be done differently since POD does not have null or floating point numbers.
- `()` and unit structs are serialized as an empty `Array` (the JSON serializer uses `null`)
- `None` is serialized as the string `"None"` (the JSON serializer uses `null`)
- `Some(x)` is serialized as a dictionary mapping `"Some"` to `x` (the JSON serializer just uses `x`)
    - Should we use an empty `Array/Set` for `None` and a one-element `Array/Set` for `Some(x)` instead?
- `f32` and `f64` are serialized as strings using scientific notation
- Bytes are serialized as a base64 encoded string (the JSON serializer uses an array of numbers but I figured that we may not to build a large Merkle tree for a byte array)


dgulotta:
- Added a deserializer for `Dictionary`
- Added `pod2::` prefix to serde names of `RawValue`, `Point`, `SecretKey` so the serializer won't misidentify user defined types with these names
- Changed `()` to serialize to `Bool(false)`
- Changed `None` to serialize to an empty set and `Some(x)` to serialize to a one-element set

dgulotta:
After trying the procedural macro approach for a while, I concluded that it's probably better in principle but also more time-consuming to implement.  So the serde approach seems more realistic for now.

Issue #413, opened 2025-09-08
check_custom_pred incorrectly rejecting operations with free parameters


dgulotta:
Consider the following code.
```rust
let params = Default::default();
let batch = parse(
    r#"
    two_equal(x,y,z) = OR(
        Equal(?x,?y)
        Equal(?y,?z)
        Equal(?x,?z)
    )  
    "#,
    &params,
    &[],
)
.unwrap()
.custom_batch;
let mut builder = MainPodBuilder::new(&params, &VDSet::new(6, &[]).unwrap());
let cpr = CustomPredicateRef { batch, index: 0 };
let eq_st = builder.priv_op(Operation::eq(1, 1)).unwrap();
let op = Operation::custom(cpr.clone(), [eq_st, Statement::None, Statement::None]);
let st = Statement::Custom(cpr, [1, 1, 2].into_iter().map(Value::from).collect());
builder.insert(true, (st, op)).unwrap();
let prover = MockProver {};
builder.prove(&prover).unwrap();
```
This code should produce a valid POD, but instead it gives an error:
```
called `Result::unwrap()` on an `Err` value: Inner: Value 0 does not match argument 2 with index 2 in the following custom predicate:
two_equal(x, y, z) = OR(
  Equal(?x, ?y)
  Equal(?y, ?z)
  Equal(?x, ?z)
)
   0: pod2::middleware::error::Error::mismatched_wildcard_value_and_statement_arg
             at /home/dan/.cargo/git/checkouts/pod2-6221e9414caa177b/51eb101/src/middleware/error.rs:64:33
   1: pod2::middleware::operation::check_custom_pred
             at /home/dan/.cargo/git/checkouts/pod2-6221e9414caa177b/51eb101/src/middleware/operation.rs:725:24
   2: pod2::middleware::operation::Operation::check
             at /home/dan/.cargo/git/checkouts/pod2-6221e9414caa177b/51eb101/src/middleware/operation.rs:555:17
   3: pod2::frontend::MainPodCompiler::compile_st_op
             at /home/dan/.cargo/git/checkouts/pod2-6221e9414caa177b/51eb101/src/frontend/mod.rs:725:36
   4: pod2::frontend::MainPodCompiler::compile
             at /home/dan/.cargo/git/checkouts/pod2-6221e9414caa177b/51eb101/src/frontend/mod.rs:753:18
   5: pod2::frontend::MainPodBuilder::prove
[...]

PR #415, opened 2025-09-09
Fix for #413

dgulotta:
This commit makes the middleware accept the example pod in #413 , but the backend still rejects it.


ed255 (approved):
LGTM!

PR #416, opened 2025-09-10
remove unused file gates/generic.rs

dgulotta:
The code in this file is not being used any more.

If we ever want to use SimpleGate in the future, there is an improved version at https://github.com/dgulotta/plonky2-gate-utils/.


ed255 (approved):
LGTM!

PR #417, opened 2025-09-10
fix cargo doc warnings

dgulotta:


ed255 (approved):
I've added a comment with a suggestion, please take a look, but otherwise LGTM!

PR #427, opened 2025-09-15
Allow creation of opaque custom predicate batch

ed255:
This is useful for a verifier that wants to verify a MainPod that contains a custom predicate in a public statement without knowledge of how the custom predicate is defined.

### DIFF
--- a/src/middleware/custom.rs
+++ b/src/middleware/custom.rs
@@ -404,6 +404,14 @@ impl CustomPredicateBatch {
         Arc::new(cpb)
     }
 
+    pub fn new_opaque(name: String, id: Hash) -> Arc<Self> {
+        Arc::new(Self {
+            id,
+            name,
+            predicates: vec![],
+        })
+    }
+
     /// Cryptographic identifier for the batch.
     fn calculate_id(&self, params: &Params) -> Hash {
         // NOTE: This implementation just hashes the concatenation of all the custom predicates,


robknight:
But if you don't know how the batch is defined then you're not actually verifying anything?

ed255:
> But if you don't know how the batch is defined then you're not actually verifying anything?

I think this is similar to a zk proof.
You are verifying that someone knows enough of the definition of the custom predicate to build a valid statement out of it.
And you're verifying that the statement was constructed in a valid way, even-though you don't know it's meaning.

In the AD system this can be used as follows:
- An AD is initialized with an opaque commitment to a custom predicate that defines valid updates
- A synchronizer can track committed state updates by veryfing mainPods that prove state updates using the committed custom predicate.
- At any point someone who knows the AD full state and the definition of the custom predicate can prove to a third party properties of the committed state.

robknight:
Sure, but why is there any benefit in withholding the definition of the custom predicate? The definition being of variable (and potentially long) length?

ed255:
> Sure, but why is there any benefit in withholding the definition of the custom predicate? The definition being of variable (and potentially long) length?

The benefit is saving some precious bytes from the payload that needs to be broadcasted to entities which don't require the definition to validate a correct state update.

artwyman:
> But if you don't know how the batch is defined then you're not actually verifying anything?

Can you clarify this?  My understanding is that so long as you know the hash of a custom predicate you can verify it.  Doing so without the definition seems likely to be part of the agreement of a distributed system like an AD.  I.e. When I start my AD I agree to a specific update predicate and distribute its hash, then everyone can verify that my updates properly follow that predicate.  I likely also publish the source-code so that it can be audited, but most users don't need to do that if they can simply verify the predicate by hash.

ax0 (approved):
LGTM!

Issue #381, opened 2025-08-01
Support "anonymous" in-line custom predicates


robknight:
Support custom predicates which contain inline conjunctions/disjunctions without requiring the declaration of a separate named custom predicate:

```
some_custom_pred(a, b) = AND(
  AND(
    Equal(?a["foo"], 123)
  )
  Equal(?a["bar"], ?b)
)
```

Such predicates should have the following characteristics:

- They should be anonymous, meaning that they have no name (or name is an empty string)
- They should have only public arguments
- The public arguments should be the same as, or a subset of, their caller's arguments

In the above example, `a` is declared by `some_custom_pred` and is used in the first nested `AND`. This means that the nested custom predicate must declare arguments whose names belong to the "parent" custom predicate's argument list.

These restrictions should be enforced by StatementTmplBuilder. There is nothing we can do to prevent people from violating these conventions by constructing statement templates manually, and so long as the logical structure of their custom predicates is valid then the compiler will accept them. However, in this case we make no guarantee about how these custom predicates will be rendered by pretty-printing.

An alternative, stricter mechanism would be to introduce a new kind of CustomPredicateRef, e.g. InlineCustomPredicateRef, which enforces these restrictions more forcefully.

PR #432, opened 2025-10-09
Frontend AST for Podlang

### LINKED ISSUE #380: Full AST for Podlang frontend
Before converting to middleware data structures for use by the compiler, we should parse Podlang into a frontend AST, which would preserve certain features:

- Syntactic sugar predicates like `Gt` or `DictContains`
- When implemented, inline ANDs/ORs #381
- When implemented, longer custom predicates #382 
- Comments, which would allow for experimenting with structured comments for type-checking etc.

This would also allow for full two-way parsing/pretty-printing, allowing us to have a standard code formatter.

We currently desugar custom predicates in the `StatementTmplBuilder`. However, this breaks two-way conversion between Podlang source and pretty-printed output based on the custom predicate data structures.

Instead, we should parse to the sugared version, and desugar in the compiler.

### LINKED ISSUE #382: Support longer custom predicates
Custom predicates are limited by the number of arguments a statement can take, since each statement template in the predicate is to be matched by an argument to the resulting custom statement.

However, we can create a set of logically linked custom predicates, such that any statement about the "head" predicate will necessarily depend on the other predicates in the chain.

We can use this to allow users to write longer-than-usual custom predicates, which are then broken down into a "linked list" of custom predicates.

This has some overlap with #381, as perhaps the "linked" predicates would be anonymous custom predicates.

We will need to give some thought to _where_ in the system these modifications are applied. In #380 we want to delay "de-sugaring" of predicates until a later stage of the compilation process, or to find some way to keep a copy of the original de-sugared content; we will need to do the same here.

robknight:
Closes #380 
Closes #382 

This PR adds support for longer custom predicates. It does so by defining a new AST which matches the structure of a Podlang document, preserving the original custom predicate lengths and sugared predicates (e.g. `Gt`, `DictContains`). This AST can then be:

- Validated, ensuring that predicates are correctly declared or imported, with error messages which identify the exact line/column on which any error occurs (similar to how the parser identifies lines/columns for parsing errors)
- "Lowered" to middleware, which performs two steps:
	1. Splitting of predicates into chains of linked predicates in cases where a predicate exceeds the length limit
	2. Desugaring via `CustomPredicateBuilder`/`StatementTmplBuilder`
	
The "frontend" AST can evolve to include new features, such as #381. It may also support better analysis tools, since the desugared predicates preserve some "type hints" for predicates like `DictContains`.

The external API for the `lang` module remains unchanged, so all existing code should continue to work and all tests pass. However, new code will be able to take advantage of longer predicate lengths.

## Future changes

There are a few more things we might want to do with this capability:

### Distribute the high-level source with serialized PODs

It is already the case that desugared middleware data is insufficient to recover the original Podlang source, but with this change (and future changes) the gap between the source and the middleware data will grow. We could switch to an approach of attaching the source to serialized PODs, which would allow the recipient of a POD to see the source code for the predicates in the original form rather than a "decompiled" form.

### Versioning

The frontend->middleware compilation process will change over time, so there is a possibility that the same source might later compile to a different middleware representation. To ensure reproducibility, we may want to introduce a versioning system for the language and compiler, e.g. a `#pragma version 1` kind of syntax in Podlang source.

### Splitting into multiple batches

Currently the splitting system will only split a predicate up to the limit of a single batch. Assuming a predicate length limit of 5 and a batch size of 4, this means that a 17-statement predicate would take up an entire batch (the first three chunks are length 4 with 1 "chaining" call, and the final chunk has 5 statements). The splitting algorithm currently does not create a new batch for the overflow. Given the limit of two batches per MainPod proof, we may want to think carefully about how to handle this.

### DIFF
--- a/src/lang/frontend_ast.rs
+++ b/src/lang/frontend_ast.rs
@@ -527,10 +527,10 @@ pub mod parse {
     use pest::iterators::Pair;
 
     use super::*;
-    use crate::lang::parser::Rule;
+    use crate::lang::parser::{self, Rule};
 
     /// Convert a Pest document pair to an AST Document
-    pub fn parse_document(pair: Pair<Rule>) -> Result<Document, crate::lang::parser::ParseError> {
+    pub fn parse_document(pair: Pair<Rule>) -> Result<Document, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::document);
         let mut items = Vec::new();
 
@@ -653,7 +653,7 @@ pub mod parse {
 
     fn parse_custom_predicate_def(
         pair: Pair<Rule>,
-    ) -> Result<CustomPredicateDef, crate::lang::parser::ParseError> {
+    ) -> Result<CustomPredicateDef, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::custom_predicate_def);
         let span = get_span(&pair);
         let mut inner = pair.into_inner();
@@ -722,7 +722,7 @@ pub mod parse {
         }
     }
 
-    fn parse_request_def(pair: Pair<Rule>) -> Result<RequestDef, crate::lang::parser::ParseError> {
+    fn parse_request_def(pair: Pair<Rule>) -> Result<RequestDef, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::request_def);
         let span = get_span(&pair);
         let mut statements = Vec::new();
@@ -743,7 +743,7 @@ pub mod parse {
         })
     }
 
-    fn parse_statement(pair: Pair<Rule>) -> Result<StatementTmpl, crate::lang::parser::ParseError> {
+    fn parse_statement(pair: Pair<Rule>) -> Result<StatementTmpl, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::statement);
         let span = get_span(&pair);
         let mut inner = pair.into_inner();
@@ -768,9 +768,7 @@ pub mod parse {
         })
     }
 
-    fn parse_statement_arg(
-        pair: Pair<Rule>,
-    ) -> Result<StatementTmplArg, crate::lang::parser::ParseError> {
+    fn parse_statement_arg(pair: Pair<Rule>) -> Result<StatementTmplArg, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::statement_arg);
         let inner = pair.into_inner().next().unwrap();
 
@@ -782,9 +780,7 @@ pub mod parse {
         }
     }
 
-    fn parse_anchored_key(
-        pair: Pair<Rule>,
-    ) -> Result<AnchoredKey, crate::lang::parser::ParseError> {
+    fn parse_anchored_key(pair: Pair<Rule>) -> Result<AnchoredKey, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::anchored_key);
         let span = get_span(&pair);
         let mut inner = pair.into_inner();
@@ -813,14 +809,12 @@ pub mod parse {
         }
     }
 
-    fn parse_literal_value(
-        pair: Pair<Rule>,
-    ) -> Result<LiteralValue, crate::lang::parser::ParseError> {
+    fn parse_literal_value(pair: Pair<Rule>) -> Result<LiteralValue, parser::ParseError> {
         assert_eq!(pair.as_rule(), Rule::literal_value);
         let inner = pair.into_inner().next().unwrap();
 
         match inner.as_rule() {
-            Rule::literal_int => Ok(LiteralValue::Int(parse_literal_int(inner))),
+            Rule::literal_int => Ok(LiteralValue::Int(parse_literal_int(inner)?)),
             Rule::literal_bool => Ok(LiteralValue::Bool(parse_literal_bool(inner))),
             Rule::literal_string => Ok(LiteralValue::String(parse_literal_string(inner)?)),
             Rule::literal_raw => Ok(LiteralValue::Raw(parse_literal_raw(inner))),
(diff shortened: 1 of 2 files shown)


src/lang/frontend_ast.rs
```diff
@@ -0,0 +1,1262 @@
...
+    Array(LiteralArray),
+    Set(LiteralSet),
+    Dict(LiteralDict),
+}
+
+/// Integer literal
+#[derive(Debug, Clone, PartialEq)]
+pub struct LiteralInt {
+    pub value: i64,
+    pub raw: String, // Original string representation
```
ed255: Why is the `raw` needed, can't we just stringify the `value` when necessary?  The same question applies to `LiteralString`

src/lang/frontend_ast_validate.rs
```diff
@@ -0,0 +1,803 @@
...
+    Custom {
+        index: usize,
+    },
+    BatchImported {
+        batch: Arc<CustomPredicateBatch>,
+        index: usize,
+    },
+    IntroImported {
+        name: String,
+        args_len: usize,
```
ed255: what's the difference between this value and `PredicateInfo.public_arity`?

src/lang/frontend_ast_lower.rs
```diff
@@ -0,0 +1,798 @@
...
+                } => Predicate::Intro(IntroPredicateRef {
+                    name: name.clone(),
+                    args_len: *args_len,
+                    verifier_data_hash: *verifier_data_hash,
+                }),
+            }
+        } else {
+            return Err(LoweringError::PredicateNotFound {
+                name: pred_name.clone(),
+            });
```
ed255: Shouldn't this case be impossible since we already validated the ast?  If that's the case, I think an `unreachable!` would be more appropriate here.

src/lang/frontend_ast_lower.rs
```diff
@@ -0,0 +1,798 @@
...
+                middleware::Value::from(point)
+            }
+            LiteralValue::SecretKey(sk) => {
+                // Parse base64 to SecretKey
+                let bytes = deserialize_bytes(&sk.base64)
+                    .map_err(|_| LoweringError::InvalidArgumentType)?;
+                let secret_key = SecretKey::from_bytes(&bytes)
+                    .map_err(|_| LoweringError::InvalidArgumentType)?;
+                middleware::Value::from(secret_key)
+            }
```
ed255: I would like to propose moving the parsing of these three types to `frontend_ast` so that they are parsed as early as possible.
This will mean that the `frontend_ast::parse::parse_document` would need to return a `Result`.

The reason for this proposal is:
- we can parse these values without knowing the params
- we can parse them in a single pass during the pest pairs tree traversal
- we can return from an error earlier
- the lowering is more complex than the pest tree traversal & parsing, so removing a bit of complexity from the lowerer is good

ed255: It would also be nice to move the parsing of literal containers to `parse_document` but they depend on params (#436 would fix that)

src/lang/frontend_ast_lower.rs
```diff
@@ -0,0 +1,798 @@
...
+        }
+
+        // Convert to &str slices for builder API
+        let public_args_str: Vec<&str> = public_arg_names.iter().map(|s| s.as_str()).collect();
+        let private_args_str: Vec<&str> = private_arg_names.iter().map(|s| s.as_str()).collect();
+
+        // Add predicate to batch using builder
+        let conjunction = pred_def.conjunction_type == ConjunctionType::And;
+
+        if conjunction {
```
ed255: You could use `cpb_builder.predicate(name, conjunction, ...)` to avoid this `if/else`.

src/lang/frontend_ast_split.rs
```diff
@@ -0,0 +1,780 @@
...
+    pub private_args: Vec<String>,
+    /// Public arguments promoted to pass to next link (empty if last link)
+    pub public_args_out: Vec<String>,
+}
+
+/// Wildcard usage information
+#[derive(Debug, Clone)]
+struct WildcardUsage {
+    _wildcard_name: String,
+    /// Indices of constraints using this wildcard
```
ed255: I guess "constraint == statement template" ?

ed255:
In general the PR looks good to me!  I've left many small comments and suggestions, please take a look.
The only comment that is perhaps not small is https://github.com/0xPARC/pod2/pull/432#discussion_r2470122234 please let me know what you think about it.

I'm really looking forward to future iterations of the compiler!  In particular I would be very interested in generating multiple batches from a single podlang document.

After this PR I guess it's not straight forward to create statements for a compiled custom predicate if you're only looking at the podlang source right?  Do you have any thoughts on being able to use something like the MainPodBuilder where the user can just declare statements following the same shape as the predicate defined in podlang?  That is, if a custom predicate has 7 statement templates, then I would be able to create 7 statements and use them as input to create the custom statement.

src/lang/frontend_ast.rs
```diff
@@ -0,0 +1,1262 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct Identifier {
+    pub name: String,
+    pub span: Option<Span>,
+}
+
+/// Hash value in hex format (0x...)
+#[derive(Debug, Clone, PartialEq)]
+pub struct HashHex {
+    pub value: String, // Includes the 0x prefix
```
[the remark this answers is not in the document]
robknight: Good idea!

src/lang/frontend_ast_validate.rs
robknight: It is redundant.

src/lang/frontend_ast_validate.rs
```diff
@@ -0,0 +1,803 @@
...
+        };
+
+        // Check arity:
+        // - For native predicates: always use total arity
+        // - For custom predicates: always use public_arity (private args are not passed at call sites)
+        let expected_arity = match pred_info.kind {
+            PredicateKind::Custom { .. }
+            | PredicateKind::BatchImported { .. }
+            | PredicateKind::IntroImported { .. } => pred_info.public_arity,
+            PredicateKind::Native(_) => pred_info.arity,
```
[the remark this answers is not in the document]
robknight: Good point.

src/lang/frontend_ast_lower.rs
robknight: Done, except for the containers.

src/lang/frontend_ast_lower.rs
```diff
@@ -0,0 +1,798 @@
...
+                }),
+            }
+        } else {
+            return Err(LoweringError::PredicateNotFound {
+                name: pred_name.clone(),
+            });
+        };
+
+        // Check args count
+        if stmt.args.len() > self.params.max_statement_args {
```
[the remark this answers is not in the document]
ed255: I see, I hadn't considered this case.  This makes sense.

src/lang/frontend_ast_split.rs
```diff
@@ -0,0 +1,780 @@
...
+            predicate: pred.name.name.clone(),
+            count: public_args,
+            max_allowed: params.max_statement_args,
+            message: "Public arguments exceed max operation args - cannot call this predicate"
+                .to_string(),
+        });
+    }
+
+    // Check: total args must fit in wildcard limit
+    if total_args > params.max_custom_predicate_wildcards {
```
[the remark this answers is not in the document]
ed255: sounds good!

PR #442, opened 2025-12-15
Remove max_depth in native MerkleTree

### LINKED ISSUE #436: Consider delaying depth check in MerkleTree
Currently when building a MerkleTree we have to set the maximum depth in advance.  This affects construction of each container type as well as serialization / deserialization of Container types.

An alternative approach would be to build MerkleTrees without checking for a maximum depth.  Then only proving merkle proofs in circuits (which currently have a max depth) we would return error if the proof is too long.

Pros:
- More ergonomic creation of containers
- Simpler serialization/deserialization of containers
Cons:
- When a MerkleTree gets too big we get the error later in the chain

ed255:
This simplifies the MerkleTree (and container) API.
Defer the max depth check when assigning the witness (merkle proof siblings) to the merkle tree circuit.

In this implementation the native Merkle Tree branches grow as much as they needed.  There are no checks of max depth in the merkle tree.  All keys are 256 bits (I added a debug_assert for this); so in the worst case a path will have depth 256.  It can't have a longer depth because the `insert` method calls `prove_nonexistence` which errors if the key already exists; another one may exist which must be different and thus require a path <= 256 depth. 

Resolve #436

### DIFF
--- a/src/frontend/serialization.rs
+++ b/src/frontend/serialization.rs
@@ -96,7 +96,7 @@ mod tests {
             (TypedValue::Bool(true), "true"),
             (
                 TypedValue::Array(Array::new(vec!["foo".into(), false.into()])),
-                "{\"max_depth\":32,\"array\":[\"foo\",false]}",
+                "{\"array\":[\"foo\",false]}",
             ),
             (
                 TypedValue::Dictionary(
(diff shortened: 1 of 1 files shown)


arnaucube (approved):
LGTM! Much cleaner now to use merkletree related interfaces ^^

PR #443, opened 2025-12-15
[docs] Port the example that @ax0 told me months ago about a theorem-ish perspective on predicates, and updates the signature section adding a note (& references) on the current signature scheme.

arnaucube:
Port the example that @ax0 told me months ago about a theorem-ish perspective on predicates, and updates the signature section adding a note (& references) on the current signature scheme.

### DIFF
--- a/book/src/simpleexample.md
+++ b/book/src/simpleexample.md
@@ -149,39 +149,41 @@ eth_friend(?1, ?2, ?3, ?4) = and<
 ## Another perspective
 When working with statements and operations, it might be useful to see them from another perspective:
 
-- A *statement* can be seen as the *constraints* of a traditional zk-circuit,
-  which can be true or false.
 - A *predicate* is a relation formula, which when filled with values becomes a
   *statement*.
+- A *statement* can be seen as the *constraints* of a traditional zk-circuit,
+  which can be true or false.
 - An *operation* comprises the deduction rules, which are rules used to deduce
   new statements from previous statements or used to construct new statements
   from values.
 
 $$
-predicate \cong circuit/relation to be fulfilled\\
-statement \cong constraints filled with the witness\\
+predicate \cong circuit/relation~to~be~fulfilled\\
+statement \cong constraints~filled~with~the~witness\\
 operations \cong deduction~rules
 $$
 
 
 For example,
-- `Equal` for integers is a predicate.
-- `5 Equal 5` is a statement (true)
-- `5 Equal 4` is a statement (false)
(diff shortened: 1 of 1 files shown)


book/src/simpleexample.md
```diff
@@ -145,3 +145,45 @@ eth_friend(?1, ?2, ?3, ?4) = and<
...
+
+</div>
+<div style="padding:10px;max-width:45%;">
+    
+We can view it as:
+
+The *statement* `IsWood(item)` is `true` if and only if:<br>
+ $\exists$ `ingreds`, `inputs`, `key`
+ such that the following statements hold:
+     `ItemDef`, `Equal`, `DictContains`.
```
ed255: I find this a bit confusing because `ItemDef, Equal, DictContains` are predicates, not statements.
I would rewrite this like:

The *statement* `IsWood(0x1234)` is `true` if and only if:<br>
 $\exists$ `ingreds`, `inputs`, `key`
 such that the following statements hold:
     `ItemDef(0x1234, ingreds, inputs, key)`, `Equal(inputs, {})`, `DictContains(ingreds, "blueprint", "wood")`.

book/src/simpleexample.md
```diff
@@ -145,3 +145,56 @@ eth_friend(?1, ?2, ?3, ?4) = and<
...
+statement \cong constraints filled with the witness\\
+operations \cong deduction~rules
+$$
+
+
+For example,
+- `Equal` for integers is a predicate.
+- `5 Equal 5` is a statement (true)
+- `5 Equal 4` is a statement (false)
+
```
artwyman: This could use an example of an "operation", i.e. deduction rule.  Like transitive equality.  (Doesn't necessarily have to map 1:1 to POD2 operations to be a helpful example, but it would be good if it were.)

book/src/simpleexample.md
```diff
@@ -145,3 +145,56 @@ eth_friend(?1, ?2, ?3, ?4) = and<
...
+
+
+<div style="display:flex;margin-left:-60px;">
+<div style="padding:10px;max-width:50%; border-right:1px solid #ccc;">
+    
+So, for example, for the given predicate:
+
+```
+IsWood(item, private: ingreds,inputs,key)=AND(
```
artwyman: I think using digital objects details in examples at this level feels like it will cause more confusion.  At this level nobody knows what ItemDef means.  I'd suggest using a simpler made-up example similar to the mathematical examples above.  Maybe a predicate like IsComposite(n) which proves a factorization using ProductOf statements?

PR #444, opened 2025-12-29
Create multiple PODs where resource limits for a single POD are exceeded

robknight:
## Summary

Adds `MultiPodBuilder`, a wrapper for `MainPodBuilder` that automatically handles cases where statements exceed per-POD limits by splitting across multiple PODs.

When building complex proofs, it's possible to exceed per-POD resource limits (max statements, Merkle proofs, custom predicate batches, etc.). Previously, this would result in builder errors that required manual intervention to split work across PODs. `MultiPodBuilder` automates this process.

## Implementation

- Uses MILP (Mixed Integer Linear Programming) via `good_lp` with the `microlp` backend to find the optimal POD packing that minimises the number of PODs needed
- Respects all per-POD resource constraints: statements, public statements, Merkle proofs, Merkle state transitions, custom predicate verifications, signed_by, public_key_of, and custom predicate batch limits
- Handles dependencies between statements: if statement B depends on statement A and they end up in different PODs, A is automatically made public and copied to B's POD
- Supports re-proving statements in multiple PODs when that's more efficient than copying

## API

```rust
// Basic usage (same as MainPodBuilder)
let mut builder = MultiPodBuilder::new(&params, &vd_set);
builder.pub_op(Operation::eq(1, 1))?;
builder.priv_op(Operation::lt(1, 2))?;

// Optionally inspect solution before proving
let solution = builder.solve()?;
println!("Will create {} PODs", solution.pod_count);

// Prove all PODs
let result = builder.prove(&prover)?;
// result.pods - all PODs in dependency order
// result.output_pods() - PODs containing user-requested public statements
// result.intermediate_pods() - supporting PODs containing intermediate statements

// Custom options
let options = MultiPodOptions { max_pods: 50 };
let builder = MultiPodBuilder::new_with_options(&params, &vd_set, options);
```

## New Dependencies

- `good_lp` with `microlp` feature - pure Rust MILP solver, no external dependencies

### DIFF
--- a/src/frontend/multi_pod/mod.rs
+++ b/src/frontend/multi_pod/mod.rs
@@ -754,7 +754,7 @@ mod tests {
 
         // pred_a accepts a Contains statement
         // pred_b accepts a pred_a statement (Custom statement from pred_a)
-        let batch = parse(
+        let parsed = parse(
             r#"
             pred_a(X) = AND(Contains(X, "k", 1))
             pred_b(X) = AND(pred_a(X))
@@ -762,8 +762,10 @@ mod tests {
             &params,
             &[],
         )
-        .expect("parse predicates")
-        .custom_batch;
+        .expect("parse predicates");
+        let batch = parsed
+            .first_batch()
+            .expect("parse predicates should have a batch");
 
         let mut builder = MultiPodBuilder::new(&params, vd_set);
 
@@ -1478,21 +1480,29 @@ mod tests {
 
         // Create 4 separate batches using podlang parser
         // Each batch has a simple predicate that checks a Contains statement
-        let batch1 = parse(r#"pred1(A) = AND(Contains(A, "x", 1))"#, &params, &[])
-            .expect("parse batch1")
-            .custom_batch;
+        let parsed1 =
+            parse(r#"pred1(A) = AND(Contains(A, "x", 1))"#, &params, &[]).expect("parse batch1");
+        let batch1 = parsed1
+            .first_batch()
+            .expect("parse batch1 should have a batch");
 
-        let batch2 = parse(r#"pred2(A) = AND(Contains(A, "x", 2))"#, &params, &[])
-            .expect("parse batch2")
-            .custom_batch;
+        let parsed2 =
+            parse(r#"pred2(A) = AND(Contains(A, "x", 2))"#, &params, &[]).expect("parse batch2");
+        let batch2 = parsed2
+            .first_batch()
+            .expect("parse batch2 should have a batch");
 
-        let batch3 = parse(r#"pred3(A) = AND(Contains(A, "x", 3))"#, &params, &[])
-            .expect("parse batch3")
-            .custom_batch;
+        let parsed3 =
+            parse(r#"pred3(A) = AND(Contains(A, "x", 3))"#, &params, &[]).expect("parse batch3");
+        let batch3 = parsed3
+            .first_batch()
+            .expect("parse batch3 should have a batch");
 
-        let batch4 = parse(r#"pred4(A) = AND(Contains(A, "x", 4))"#, &params, &[])
-            .expect("parse batch4")
-            .custom_batch;
+        let parsed4 =
+            parse(r#"pred4(A) = AND(Contains(A, "x", 4))"#, &params, &[]).expect("parse batch4");
+        let batch4 = parsed4
+            .first_batch()
+            .expect("parse batch4 should have a batch");
 
         let mut builder = MultiPodBuilder::new(&params, vd_set);
 
@@ -1577,7 +1587,7 @@ mod tests {
         let vd_set = &*MOCK_VD_SET;
 
         // Chain of predicates: each accepts the output of the previous
-        let batch = parse(
+        let parsed = parse(
             r#"
             pred_a(X) = AND(Contains(X, "k", 1))
             pred_b(X) = AND(pred_a(X))
@@ -1587,8 +1597,10 @@ mod tests {
             &params,
             &[],
         )
-        .expect("parse predicates")
-        .custom_batch;
+        .expect("parse predicates");
+        let batch = parsed
+            .first_batch()
+            .expect("parse predicates should have a batch");
 
         let mut builder = MultiPodBuilder::new(&params, vd_set);
 
@@ -1719,7 +1731,7 @@ mod tests {
         // pred_a takes TWO custom statement arguments (b_out and c_out)
         // pred_b and pred_c each take a Contains
         // Note: AND clauses are newline-separated, not comma-separated
-        let batch = parse(
+        let parsed = parse(
             r#"
             pred_b(X) = AND(Contains(X, "k", 1))
             pred_c(X) = AND(Contains(X, "k", 1))
@@ -1731,8 +1743,10 @@ mod tests {
             &params,
             &[],
         )
-        .expect("parse predicates")
-        .custom_batch;
+        .expect("parse predicates");
+        let batch = parsed
+            .first_batch()
+            .expect("parse predicates should have a batch");
 
         let mut builder = MultiPodBuilder::new(&params, vd_set);
 
@@ -1833,9 +1847,11 @@ mod tests {
         let vd_set = &*MOCK_VD_SET;
 
         // Create two SEPARATE batches (parsed separately to get different batch IDs)
-        let batch_a = parse(r#"pred_a(X) = AND(Contains(X, "k", 1))"#, &params, &[])
-            .expect("parse batch_a")
-            .custom_batch;
+        let parsed_a =
+            parse(r#"pred_a(X) = AND(Contains(X, "k", 1))"#, &params, &[]).expect("parse batch_a");
+        let batch_a = parsed_a
+            .first_batch()
+            .expect("parse batch_a should have a batch");
 
         // batch_b's pred_b accepts pred_a statements
         // Must use "use batch" syntax to reference external predicates
@@ -1846,9 +1862,11 @@ mod tests {
             pred_b(X) = AND(pred_a(X))
             "#
         );
-        let batch_b = parse(&batch_b_src, &params, std::slice::from_ref(&batch_a))
-            .expect("parse batch_b")
-            .custom_batch;
+        let parsed_b =
+            parse(&batch_b_src, &params, std::slice::from_ref(batch_a)).expect("parse batch_b");
+        let batch_b = parsed_b
+            .first_batch()
+            .expect("parse batch_b should have a batch");
 
         let mut builder = MultiPodBuilder::new(&params, vd_set);
 


src/frontend/multi_pod/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+            if solution.output_pod_indices.contains(&idx) {
+                output_indices.push(idx);
+            } else {
+                intermediate_indices.push(idx);
+            }
+        }
+
+        Ok(MultiPodResult {
+            pods,
+            output_indices,
```
ed255: Do you think there will be real use cases with multiple output pods?
Because the API would be simpler if there's a single output pod here.

src/frontend/multi_pod/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+        solution: &MultiPodSolution,
+        earlier_pods: &[MainPod],
+        prover: &dyn MainPodProver,
+    ) -> Result<MainPod> {
+        let mut builder = MainPodBuilder::new(&self.params, &self.vd_set);
+
+        // Add external input PODs
+        for input_pod in &self.input_pods {
+            builder.add_pod(input_pod.clone())?;
+        }
```
ed255: This doesn't seem right.  Shouldn't the input pods only be added to the pods that need them, instead of adding them to all of them?

src/frontend/multi_pod/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+        // Create a mapping from statement to its source (for copy operations)
+        let mut stmt_sources: HashMap<usize, StmtSource> = HashMap::new();
+        for &stmt_idx in statements_in_this_pod {
+            stmt_sources.insert(stmt_idx, StmtSource::Local);
+        }
+        for earlier_pod_idx in 0..pod_idx {
+            for &stmt_idx in &solution.pod_public_statements[earlier_pod_idx] {
+                stmt_sources
+                    .entry(stmt_idx)
+                    .or_insert(StmtSource::FromPod(earlier_pod_idx));
```
ed255: This only inserts if the entry doesn't exist.
In which situation will an entry for `stmt_idx` already exist?

src/frontend/multi_pod/cost.rs
```diff
@@ -0,0 +1,278 @@
...
+/// running the full MILP solver.
+pub fn estimate_pod_count(operations: &[Operation], params: &Params) -> usize {
+    let mut aggregate = AggregateCost::new();
+    for op in operations {
+        aggregate.add(&StatementCost::from_operation(op));
+    }
+    // Add a fudge factor for dependency overhead and public statement constraints
+    let lower_bound = aggregate.lower_bound_pods(params);
+    // Heuristic: add ~20% overhead for dependencies
+    lower_bound + (lower_bound / 5).max(0)
```
ed255: Isn't the `max(0)` here a no-op?

src/frontend/multi_pod/deps.rs
```diff
@@ -0,0 +1,211 @@
...
+        for deps in &self.statement_deps {
+            for dep in &deps.depends_on {
+                if matches!(dep, StatementSource::Internal(_)) {
+                    in_degree[deps.index] += 1;
+                }
+            }
+        }
+
+        // Start with nodes that have no dependencies
+        let mut queue: Vec<usize> = (0..n).filter(|&i| in_degree[i] == 0).collect();
```
ed255: It would be nice if this queue was a FIFO instead of a LIFO so that if the original graph was already topologically sorted, it keeps the same ordering.

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+    if min_pods > input.max_pods {
+        return Err(super::Error::Solver(format!(
+            "Problem requires at least {} PODs, but max_pods is set to {}. \
+             Increase Options::max_pods to allow more PODs.",
+            min_pods, input.max_pods
+        )));
+    }
+
+    // Upper bound: add slack but cap at configured max_pods
+    let max_pods = (min_pods * 2).max(2).min(n).min(input.max_pods);
```
ed255: `min_pods >= 1` because `min_pods_by_public >= 1`.
Then `min_pods * 2 >= 2`, which makes the `max(2)` a no-op.

Also, what's the goal of doing `min(n)`?

And is there any reason for the slack to be `+100%`?

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+        .collect();
+
+    // Objective: minimize number of PODs used
+    let objective: Expression = pod_used.iter().sum();
+    let mut model = vars.minimise(objective).using(default_solver);
+
+    // Constraint 1: Each statement must be proved at least once
+    for s in 0..n {
+        let sum: Expression = prove[s].iter().sum();
+        model = model.with(constraint!(sum >= 1));
```
ed255: I see that the `SolverModel` has the method `add_constraint(&mut self, ...)`, which doesn't require reassigning `model` each time.

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    // Constraint 2: Output-public statements must be public in at least one POD
+    for &s in input.output_public_indices {
+        let sum: Expression = public[s].iter().sum();
+        model = model.with(constraint!(sum >= 1));
+    }
+
+    // Constraint 3: Public implies proved
```
ed255: We want the public statements to appear in a small subset of pods, instead of having them spread around all pods.
Do these constraints factor this in?

ed255: Related question: can a private statement appear in a public slot of a pod that also contains a public statement?

For example imagine I have this graph of statements:
a -> b -> c -> d
b -> e

I want `d` and `e` to be public.  

Is this arrangement possible with the current solver constraints?
pod0:
- priv: a, b, e
- pub: b, e

pod 1:
- in: pod0 [b, e]
- priv: c, d
- pub: d

This arrangement is leaking `b` through pod0.

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+    for s in 0..n {
+        for p in 0..max_pods {
+            model = model.with(constraint!(prove[s][p] <= pod_used[p]));
+        }
+    }
+
+    // Constraint 5: Dependencies
+    // If s depends on d (internal), and s is proved in p, then either:
+    // - d is proved in p, OR
+    // - d is public in some earlier POD p' < p
```
ed255: Not only this, but the input pods of `p` must contain `d` as public statement.
Is this considered?

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+    // Constraint 8: Input POD limits using uses_input
+    // uses_input[p][pp] >= prove[s][p] + public[d][pp] - 1 for each dependency (s depends on d)
+    for s in 0..n {
+        for dep in &input.deps.statement_deps[s].depends_on {
+            if let StatementSource::Internal(d) = dep {
+                for p in 1..max_pods {
+                    for pp in 0..p {
+                        // If s is proved in p and d is public in pp, then uses_input[p][pp] = 1
+                        model = model.with(constraint!(
+                            uses_input[p][pp] >= prove[s][p] + public[*d][pp] - 1.0
```
ed255: I wonder if there are any issues here because:
- A statement could be proved in multiple pods
- What happens if `d` is proved in `p` and `pp`?  Then there's no need for `pp` to be input to `p`.  But this constraint will force that.

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use super::*;
+    use crate::frontend::multi_pod::deps::StatementDeps;
+
+    fn make_simple_deps(n: usize) -> DependencyGraph {
```
ed255: All these tests use this method, which a graph with disconnected nodes.

I think at least you should have one test with a connected graph that needs to be cut.

src/frontend/multi_pod/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+    }
+
+    /// Build and prove all PODs.
+    ///
+    /// This first solves if not already solved, then builds and proves
+    /// all necessary PODs in dependency order.
+    pub fn prove(&mut self, prover: &dyn MainPodProver) -> Result<MultiPodResult> {
+        // Ensure we have a solution
+        self.solve()?;
+        let solution = self.cached_solution.as_ref().unwrap();
```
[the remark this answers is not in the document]
robknight: The borrow checker doesn't like that, unfortunately.

src/frontend/multi_pod/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+                pods: vec![],
+                output_indices: vec![],
+                intermediate_indices: vec![],
+            });
+        }
+
+        // Build PODs in order
+        let mut pods: Vec<MainPod> = Vec::with_capacity(solution.pod_count);
+
+        for pod_idx in &solution.prove_order {
```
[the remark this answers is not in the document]
robknight: I'm not sure that I understand the question. The prove order ensures that dependencies are proven in the correct order, such that earlier PODs contain statements that are copied to later PODs.

src/frontend/multi_pod/mod.rs
robknight: I've removed this feature as it adds complexity. I could imagine hypothetical scenarios where I want to reveal a large number of digital objects and perhaps that could require more than one output POD, but that is quite speculative at this point and if we need that feature then we could choose to add it later.

src/frontend/multi_pod/mod.rs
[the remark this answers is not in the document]
ed255: I found the answer here https://github.com/0xPARC/pod2/pull/444#discussion_r2668759347
My suggestion would be to remove the `prove_order` field and here just do:
```rust
for pod_idx in 0..solution.pod_count {
```

src/frontend/multi_pod/solver.rs
robknight: I've switched to using this instead.

src/frontend/multi_pod/solver.rs
robknight: I'm not sure that I understand this fully. Is this no longer an issue, given that there is only a single output POD?

src/frontend/multi_pod/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+
+            // Now add the actual statement
+            let is_public = public_set.contains(&stmt_idx);
+            let op = self.operations[stmt_idx].clone();
+
+            let stmt = if is_public {
+                builder.pub_op(op)
+            } else {
+                builder.priv_op(op)
+            }
```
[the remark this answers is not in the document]
robknight: Done.

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+    // Solve
+    let solution = model
+        .solve()
+        .map_err(|e| super::Error::Solver(format!("{:?}", e)))?;
+
+    // Extract solution
+    let mut pod_count = 0;
+    for p in 0..max_pods {
+        if solution.value(pod_used[p]) > 0.5 {
+            pod_count = p + 1;
```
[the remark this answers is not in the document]
robknight: Good point, I'll change it to that.

src/frontend/multi_pod/solver.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    // Upper bound: add slack but cap at configured max_pods
+    let max_pods = (min_pods * 2).max(2).min(n).min(input.max_pods);
+
+    // Collect all custom batch IDs used
+    let all_batches: Vec<CustomBatchId> = input
+        .costs
+        .iter()
+        .flat_map(|c| c.custom_batch_ids.iter().cloned())
+        .collect::<HashSet<_>>()
```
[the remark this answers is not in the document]
robknight: I've removed HashSet usage entirely.

Issue #466, opened 2026-01-28
Merkle trees for custom predicate "modules"


robknight:
Currently custom predicates are arranged in small fixed-size batches, and the field elements making up the entire batch are hashed, giving us a hash for the batch. Individual predicates are identified by the combination of batch hash and their index within the batch.

Mutually recursive cycles between custom predicates are OK so long as they happen inside the same batch, because these can be handled using BatchSelf references. Mutually recursive cycles between predicates in different batches are _not_ OK, because the hashes for the two blocks depend on each other.

In some cases it is possible to refactor the Podlang code to avoid such cycles, but in other cases it is not. In some cases it is possible to automatically rearrange the batches so that cycles happen inside a batch, but for longer cycle chains it is not possible to do this.

This problem arises because the batch sizes are much smaller than we need for some new use-cases that were not anticipated during the original design. While examples such as Ethdos manage mutual recursion inside a single batch, newer examples require tens of connected custom predicates, which fit poorly with a fixed batch size of 4. We have tried to handle these larger sets of predicates by automatically splitting them into batches, but this has exposed the issue with cyclical dependencies between batches.

Even leaving aside this problem, the optimal layout of a large predicate collection in batches is a difficult problem. Imagine a collection of 50 predicates, and then a POD which wants to use eight of these predicates. In the happy case, this requires only two batches of four, if the predicates happen to be laid out optimally for this usage pattern. This might allow all of the statements to be made in a single POD. The worst case is that the predicates are each in distinct batches, requiring eight batches to be used, which can only be achieved by using four PODs.

---

## Proposal

The proposal is to store custom predicates in "modules", where each module has a Merkle tree. References between predicates in the module can be cyclical, and chains of cyclically-connected predicates can be much longer than 4 in length (precise maximum distance depends on the size of the Merkle tree, but this could easily be measured in hundreds or thousands, giving us plenty of headroom).

Cyclical references between modules would still be problematic, but this seems much easier to understand as many mainstream programming languages disallow cyclical dependencies.

With the Merkle tree approach, it would not be necessary to limit the number of modules that can be used in a single POD. Instead, the limit would be on the number of Merkle tree openings, with one opening for each unique custom predicate that we want to use with a custom operation. Presumably multiple uses of the same predicate would require only a single opening.

Points for discussion/consideration:

- What is the appropriate maximum number of predicates used?
- How will predicates within the same module be identified? (what is the replacement for BatchSelf)
[...]

PR #472, opened 2026-02-02
merkletree: reduce gate amount (-23%) by custom poseidon to use flag as initial state

arnaucube:
This implements what @ed255 suggested in today's call: use the initial state of poseidon for the node-type flag instead of passing it as input to be hashed.

It allows to do the merkletree related hashing in 1 gate instead of 2, reducing ~23% of gates per merkle proof.

| tree levels   | 10 | 16 | 32  | 40  | 64  | 128 | 130 | 250  | 256  |
|---------------|----|----|-----|-----|-----|-----|-----|------|------|
| old num gates | 50 | 76 | 144 | 178 | 280 | 554 | 564 | 1076 | 1102 |
| new num gates | 39 | 59 | 111 | 137 | 215 | 425 | 433 | 825  | 845  |

### DIFF
--- a/src/backends/plonky2/primitives/merkletree/circuit.rs
+++ b/src/backends/plonky2/primitives/merkletree/circuit.rs
@@ -421,6 +421,10 @@ fn hash_with_flag_target<H: AlgebraicHasher<F>>(
 ) -> HashOutTarget {
     assert_eq!(inputs.len(), H::AlgebraicPermutation::RATE);
 
+    // here we set `state` to a  `SPONGE_RATE+SPONGE_CAPACITY` (8+4=12) in our
+    // case to a vector of repeated `flag` value. Later at the absorption step,
+    // it will fit the inputs values at positions 0-8, keeping the flag values
+    // at positions 8-12.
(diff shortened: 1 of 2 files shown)


gubsheep:
Out of curiosity, what would the overall effect on gate count of the MainPOD circuit be?

ax0 (approved):
LGTM!

arnaucube:
> @gubsheep : Out of curiosity, what would the overall effect on gate count of the MainPOD circuit be?

Without changing any parameters and just with the upgrade of this PR, we would go from `50230` to `49154` gates (thus, not reducing from the current power of two).

But the direct effect would be at the amount of predicates that we can fit: from the numbers reported at https://github.com/0xPARC/pod2/issues/466#issuecomment-3823531286 , with trees of 16 levels, the average cost per predicate is `75 + 25.2 + 170/4 = 142.7 gates`, where with this PR the `75` is reduced to `59`, so that the cost is `126.7 gates`.
The old `142.7 gates` allowed for `⌊1184 / 142.7⌋ = 8 predicates`, which with this PR it becomes `⌊1184 / 126.7⌋ = 9 predicates`.

PR #474, opened 2026-02-04
use BTreeSet and BTreeMap for containers

ed255:
Previously we were using HashMap and HashSet for the Dictionary and Set containers.  This was very efficient but it was slightly inconvenient because sometimes we iterate over the entries of those collections to generate statements for recursive predicates, leading to non-deterministic order of statements which makes debugging issues harder.

By using BTreeMap and BTreeSet we always get deterministic iteration over the container entries, at a small cost of performance (which should be negligible because we target consumer devices and the bottleneck is in the proving process).

As an added bonus, now when we iterate over the Dictionary and Set we visit the entries in the same order as they are laid out in the Merkle Tree.

I've also done a small change in `MultiPodBuilder` to make an internal structure deterministic.

I've requested review of 2 people to get critical feedback on this change.

### DIFF
--- a/src/frontend/multi_pod/solver.rs
+++ b/src/frontend/multi_pod/solver.rs
@@ -171,11 +171,7 @@ pub fn solve(input: &SolverInput) -> Result<MultiPodSolution> {
     // Incremental approach: try solving with increasing POD counts
     // Start with min_pods and increment until we find a feasible solution
     for target_pods in min_pods..=input.max_pods {
-        println!("DBG try solve for {} pods", target_pods);
         if let Some(solution) = try_solve_with_pods(input, target_pods, &all_batches)? {
-            println!("DBG solution found!");
-            dbg!(&input);
-            dbg!(&solution);
             return Ok(solution);
         }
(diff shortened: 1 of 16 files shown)


src/frontend/multi_pod/solver.rs
```diff
@@ -170,7 +171,11 @@ pub fn solve(input: &SolverInput) -> Result<MultiPodSolution> {
     // Incremental approach: try solving with increasing POD counts
     // Start with min_pods and increment until we find a feasible solution
     for target_pods in min_pods..=input.max_pods {
+        println!("DBG try solve for {} pods", target_pods);
```
robknight: I don't think that we should merge these debug prints, though perhaps this suggests that we need better logging/tracing in general.

robknight (commented):
I think this is a good idea in principle but we ought to consider the issue pointed out in https://github.com/0xPARC/pod2/pull/393 before merging.

ed255:
Closing this for now because the current implementation works and we have workarounds for the nondeterminism in hashmaps.

PR #476, opened 2026-02-05
Fix accidental inclusion of extra input PODs

robknight:
The final `prove` step has to compute which PODs are required for the statements we are using, ensuring that the dependencies are satisfied.

A bug was causing this to include dependencies from input PODs when the same dependency is already satisfied in the local POD. The solver correctly accounts for this, but the prover step mistakenly ignores the fact that the solver planned for that dependency to be local. This can cause too many input PODs to be added.

The fix is to refactor the logic for computing inputs, with an assertion to ensure that the total number of input PODs never exceeds `params.max_input_pods`. To aid in debugging I have also implemented the `Display` trait for `MultiPodSolution`.


ed255 (commented):
Overall LGTM! But I have a suggestion to make one part more robust, please take a look

PR #484, opened 2026-02-19
fix: points wildcard name same as predicate

dhvanipa:
`cargo run --release --example main_pod_points` was failing with `Error: error: wildcard 'points' collides with a predicate name`


ed255:
> > LGTM! Thanks for the fix!
> 
> btw I don't have the ability to merge, so you'll need to do it

sorry I meant to do it after approving; but I noticed the test hadn't run so I requested them to run and then I forgot to come back and merge.

PR #487, opened 2026-03-06
Re-implement merkletree with persistent storage (key-value db)

### LINKED ISSUE #435: Persistent MerkleTree storage
Our MerkleTree implementation uses memory to store the leaves.  To store the MerkleTree we need to serialize it entirely.

For applications where the MerkleTree is big and persistent storage is needed, this is very inconvenient.

A possible solution for this would be that the MerkleTree uses a generic key-value storage, and we can plug in a memory or persistent storage implementation when needed.

This change may affect many places of the code base.

### LINKED ISSUE #439: Underflow in MerkleTree
When the tree is empty `new_siblings.len()` is 0.  This triggers an underflow in the code:

https://github.com/0xPARC/pod2/blob/c382bf487c3901b5d9ff2df1e96850d2a169a2f7/src/backends/plonky2/primitives/merkletree/circuit.rs#L649

The `self.divergence_level` is not used in this situation so the value assigned there doesn't matter, but I think we should fix the underflow condition, which is checked by rust in debug mode and panics.

A solution could be to do a conditional assignment of 0 if `new_siblings.len() == 0`

arnaucube:
Joint work with @ax0 . 

This PR re-implements the merkletree to allow it to use a key-value database in-disk. Needed a reimplementation of the core of the tree logic, since the previous implementation was designed to first build the tree structure in memory and at the end of the operations to compute the hashes, which in a key-value db would not be ideal.

The database interface is defined by the `DB` trait, which works with (atomic) transactions, and we added a [rocksdb](https://github.com/rust-rocksdb/rust-rocksdb) implementation of it as an example ready to use (aside from a naive in-memory db).

We also extend the tests of the merkletree to cover more edge cases (for future iterations to catch potential issues).

resolves #435 , resolves #439 (this is done by setting the `divergence_level` to `MAX_DEPTH` when `new_siblings.len()` is zero)

### DIFF
--- a/src/backends/plonky2/primitives/merkletree/mod.rs
+++ b/src/backends/plonky2/primitives/merkletree/mod.rs
@@ -622,18 +622,18 @@ impl MerkleTree {
             op,
         )?;
 
-        let leaf: Leaf = match (op, maybe_value) {
+        let node: Node = match (op, maybe_value) {
             (MerkleTreeOp::Insert, Some(value)) | (MerkleTreeOp::Update, Some(value)) => {
-                Leaf::new(k, value)
+                Node::Leaf(Leaf::new(k, value))
             }
             (MerkleTreeOp::Delete, None) => {
-                // return a leaf whose hash is 'empty', to indicate that there is no leaf
-                Leaf {
+                // return an intermediate node whose hash is 'empty', to
+                // indicate that there is no leaf
+                Node::Intermediate(Intermediate {
                     hash: EMPTY_HASH,
-                    path: vec![],
-                    key: EMPTY_VALUE,
-                    value: EMPTY_VALUE,
-                }
+                    left: EMPTY_HASH,
+                    right: EMPTY_HASH,
+                })
             }
             _ => {
                 return Err(TreeError::invalid_state_transition_proof_arg(format!(
@@ -642,11 +642,11 @@ impl MerkleTree {
                 )))
             }
         };
-        let leaf_hash = leaf.hash; // variable to avoid cloning `leaf` later
-        db.store_node(Node::Leaf(leaf))?;
+        let node_hash = node.hash(); // variable to avoid cloning `node` later
+        db.store_node(node)?;
         if siblings.is_empty() {
-            // return the leaf's hash as root
-            return Ok(leaf_hash);
+            // return the node's hash as root
+            return Ok(node_hash);
         }
 
         let new_root = if op == MerkleTreeOp::Delete {
@@ -689,7 +689,7 @@ impl MerkleTree {
                 true,
             )?
         } else {
-            Self::up(db, path, siblings.len() - 1, leaf_hash, siblings, op, true)?
+            Self::up(db, path, siblings.len() - 1, node_hash, siblings, op, true)?
         };
 
         Ok(new_root)


src/backends/plonky2/primitives/merkletree/db/mod.rs
```diff
@@ -0,0 +1,140 @@
+#![allow(dead_code)]
```
ed255: This should be removed.

src/backends/plonky2/primitives/merkletree/db/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct MemTxn {
+    db: Arc<Mutex<HashMap<RawValue, Node>>>,
+    write: bool,
+}
+
+impl Txn for MemTxn {
```
ed255: The trait says that a `Txn` implements an atomic transaction, but this is not an atomic transaction.
For a markle tree with in-memory storage this doesn't matter, but it feels wrong according to the documentation and the expectation of the behavior of the trait.

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -37,29 +41,201 @@ impl Eq for MerkleTree {}
 
 impl MerkleTree {
     /// builds a new `MerkleTree` where the leaves contain the given key-values
+    // NOTE: this method is kept for legacy compatibility at higher levels of the pod2 repo. It
+    // should be replaced by the `new_with_b` or `from_db` or `empty_with_db` or similar, and the
```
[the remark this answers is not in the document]
ed255: I would personally remove this `NOTE`.  I think we should keep the `new` method and not replace it because it's very convenient as a default.

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -37,29 +41,201 @@ impl Eq for MerkleTree {}
...
+    /// stored. In this way, the same method `down` can be used by MerkleTree
+    /// methods `get`, `contains`, `prove` and `prove_nonexistence`.
+    ///
+    /// Be aware that this method will return the found leaf at the given path,
+    /// which may contain a different key and value than the expected one. And
+    /// while it does not return explicitly a `siblings` variable, the input
+    /// `siblings` is modified adding there the siblings found along the path.
+    fn down(
+        txn: &dyn Txn,
+        path_and_lvl: (Vec<bool>, usize), // path and lvl
```
ed255: is this a "hack" to avoid clippy complaining about the function having too many arguments?

ed255 (commented):
Overall looks good to me, but I would like to discuss the DB api

src/backends/plonky2/primitives/merkletree/db/mod.rs
arnaucube: Right; the memdb is only thought for testing; ideally we can use any of the offered dbs (not the mem, for now only available rocksdb) with an in-memory version of it.

src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -37,29 +41,201 @@ impl Eq for MerkleTree {}
...
-        // Iterate over key-value pairs (if any) and add them.
-        for (k, v) in kvs.iter() {
-            root.apply_op(MerkleTreeOp::Insert, *k, Some(*v)).unwrap();
-        }
+        let root = {
+            let mut txn = db.begin_txn(true).unwrap();
+
+            // Iterate over key-value pairs (if any) and add them.
+            let mut root = EMPTY_HASH;
+            txn.store_node(Node::Leaf(Leaf::new(root.into(), EMPTY_VALUE)))
```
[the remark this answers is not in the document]
arnaucube: right, updated so that the `Leaf::new` method does not have the case of 'empty leaf' (also removed this line which was not actually needed)

src/backends/plonky2/primitives/merkletree/db/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        let mut db = self
+            .db
+            .lock()
+            .map_err(|e| anyhow!("failed to acquire memdb lock for write: {}", e))?;
+        db.insert(node.hash().into(), node);
+        Ok(())
+    }
+
+    fn commit(&mut self) -> Result<()> {
```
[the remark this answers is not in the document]
arnaucube: update: as agreed in the call, we got rid of the txns https://github.com/0xPARC/pod2/pull/487/commits/202c9190d34982e7c013829c41128c41ce70abc5

ed255 (approved):
LGTM!

Issue #426, opened 2025-09-15
Raw value collisions


dgulotta:
There are some known collisions between the raw values of various typed values:
- 0, false, and an empty container
- 1 and true
- 56-byte strings and containers with 1 element (strings are hashed by appending the byte 1, packing groups of 7 bytes into a field element, and hashing; Merkle leaves are hashed by concatenating the key, value, and field element 1 and hashing)

At the very least, the 0 / empty container and string / container collisions should be fixed.  It is worth considering whether we want to have types in the backend as well.

artwyman:
Out of curiosity: does the hash of an empty-string collide with anything else?

The non-empty container vs. non-empty string collision Dan mentions is particularly worrying since it's most non-intuitive and abusable, since it's a collision between two things which contain actual data.

My personal view is that at the very least unrelated types shouldn't collide with each other.  I.e. I'm okay with 0==false and {}==[], but not 0=={}.  The design of POD1 reflected that, where all numeric types (bool/int/cryptographic) hashed the same, and string/bytes hashed the same, but they wouldn't collide with each other..  I'm not necessarily against fully-distinct types (probably with some form of tagging) but it adds some complexity which is worth discussing more deeply.

There might be relatively simple fixes here.  It seems like there are changeable constants involved in most of the collisions here.  Dan mentioned the string and container hashes have a prefix byte 1 - couldn't one of the types use a 2 instead?  Similarly couldn't Ints be packed with non-zero values, and empty containers be represented as different nonzero values?

dgulotta:
It should be simple to fix the collisions mentioned above.  We can change the hash of the empty container to an arbitrary value and/or pad integers with something other than zeros.  We compute the hash of a non-leaf Merkle tree node by appending a 2, but it should be fine to use 3 for strings. 

I don't think the hash of the empty string collides with anything.  All of the hash computations in containers hash 9 field elements, while the empty string would hash 1 field element.

ed255:
In this PR https://github.com/0xPARC/pod2/pull/493 the handling of types for Value was changed via the exposed APIs:
- Raw is no longer a type, instead it's treated as "lack of type".  All values can be casted to raw
- bool is treated like Int(0) or Int(1)
- Value now doesn't expose a type, instead it can be casted to a type, and for some values multiple casts will return Some (where some of the collisions exists).  As a general rule a cast will succeed if we have the preimage information (for types that are encoded by hashing).

[...]

PR #493, opened 2026-03-17
Support persistent storage in Containers

ed255:
This PR extends the work of https://github.com/0xPARC/pod2/pull/487 to the Containers (Dictionary, Set, Array).

The merkle tree only stores `RawValue` for both the key and the value, so it is the responsibility of the Container to store the rich value.

In order to handle containers with persistent storage efficiently (which means, cloning them or updating them should not cause an O(n) data copy) I figured we need to have a database of `Value`s indexed by their raw value; as this gives us deduplication and free cloning of containers.
The issue with this approach is that in the current design we have collisions between Value's of different types: https://github.com/0xPARC/pod2/issues/426 and the current API relies on the single type of values.

To resolve this issue I decided to change the API, instead of assuming that a Value has a fixed type, let the value be possibly multiple compatible types and let the user of the library try casting the Value to a particular type.
For this I deprecated the public access of everything related to `TypedValue` and I propose for it to be considered an implementation detail and a blackbox from the external developer point of view.  The `Value` type is now used like this:
- To create a new Value use `Value::from(...)` where you can pass any compatible type (the same types as before)
- To access the Value in typed form you cast it like `value.as_foo()` which returns `Option<Foo>`.

Previously we had a collision between `true` and `1` (and `false` and `0`).  Now it doesn't matter whether a value holds a `true` or a `1`, both should be seen as the same and both return `Some` when doing `as_int` and `as_bool`.

Similarly we had collisions with containers.  For example `set(0, 1, 2) == array[0, 1, 2]` and `set("a", "b") = dict("a": "a", "b": "b")`.  Now any container can be casted to any of `set, array, dict`.  There's a caveat here: each of these types expects a particular encoding of keys, so casting to the wrong type will return errors on some operations.

With this design it no longer matters what is being stored and recovered because the API requires the user to express the expected type and any type with collisions for particular values can be casted to the right type.

There's only one case where it's not desirable to swap one `TypedValue` for another: the `TypedValue::Raw`.  If a non-`RawValue` in the DB is replaced by the corresponding `RawValue` we erase the required information to recover the rich value.  For this reason the implementations of the database treat the `RawValue` as a special case: if an value is stored in non-`RawValue`, the corresponding `RawValue` can never overwrite it.  If a value is stored in `RawValue`, a matching non-`RawValue` will overwrite it (promoting it to a rich value).  This way we never lose data.

A consequence of this is that the serialization, `Display` and `Debug` of a container is not stable.  At any point any of the entries can be swapped for a "compatible" one if they share the storage with other containers that introduce collisions.

In this PR I rewrote all containers as wrapper to a generic `Container` which holds a `Map` from `Value` to `Value`.  The serialization of each container now uses the single implementation of the generic `Container`.

### DIFF
--- a/src/middleware/containers.rs
+++ b/src/middleware/containers.rs
@@ -1,96 +1,45 @@
 //! This file implements the types defined at
 //! <https://0xparc.github.io/pod2/values.html#dictionary-array-set> .
 
+#![allow(unused_imports)] // TODO: Remove
+
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     fmt::Debug,
     sync::{Arc, RwLock},
 };
 
 use anyhow::bail;
-use dyn_clone::DynClone;
+use dyn_clone::{clone_box, DynClone};
 use schemars::JsonSchema;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{de, ser, Deserialize, Deserializer, Serialize};
 
 use super::serialization::{ordered_map, ordered_set};
 #[cfg(feature = "backend_plonky2")]
-use crate::backends::plonky2::primitives::merkletree::{self, MerkleProof, MerkleTree};
+use crate::backends::plonky2::primitives::merkletree::{self, MerkleProof, MerkleTree, TreeError};
 use crate::{
     backends::plonky2::primitives::merkletree::MerkleTreeStateTransitionProof,
-    middleware::{hash_str, Error, Hash, Key, RawValue, Result, Value, EMPTY_HASH, EMPTY_VALUE},
+    middleware::{
+        db::{mem::MemDB, DB},
+        hash_str, Error, Hash, Key, RawValue, Result, TypedValue, Value, EMPTY_HASH, EMPTY_VALUE,
+    },
 };
 
-pub trait DB: Debug + DynClone + Sync + Send + merkletree::db::DB {
-    fn load_value(&self, raw: RawValue) -> anyhow::Result<Value>;
-    fn store_value(&mut self, value: Value) -> anyhow::Result<()>;
-}
-dyn_clone::clone_trait_object!(DB);
-
-/// MemDB implements the DB trait in a in-memory HashMap.
-#[derive(Clone, Debug, Default)]
-pub struct MemDB {
-    nodes: Arc<RwLock<HashMap<RawValue, merkletree::Node>>>,
-    values: Arc<RwLock<HashMap<RawValue, Value>>>,
-}
-
-impl MemDB {
-    pub fn new() -> Self {
-        Self::default()
-    }
-}
-
-impl merkletree::db::DB for MemDB {
-    fn load_node(&self, hash: RawValue) -> anyhow::Result<merkletree::Node> {
-        let nodes = self.nodes.read().expect("lock not poisoned");
-
-        if let Some(node) = nodes.get(&hash) {
-            return Ok(node.clone());
-        }
-
-        if hash == EMPTY_VALUE {
-            return Ok(merkletree::Node::Leaf(merkletree::Leaf::new(
-                hash,
-                EMPTY_VALUE,
-            )));
-        }
-
-        bail!("MemDB: node not found: {}", hash);
-    }
-
-    fn store_node(&mut self, node: merkletree::Node) -> anyhow::Result<()> {
-        let mut nodes = self.nodes.write().expect("lock not poisoned");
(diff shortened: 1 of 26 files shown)


src/middleware/db/rocks.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn store_node(&mut self, node: merkletree::Node) -> Result<()> {
+        self.db
+            .put(node_key(node.hash()), node.encode()?)
+            .map_err(|e| anyhow!("rocksdb transaction put failed: {e}"))
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+enum ValueAux {
```
dhvanipa: I need this in the zk-craft repo, but it is not exported, so I copied it plus the `from` and `into_value` implementations. However, those don't compile because `TypedValue` is not exported.

src/middleware/db/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+pub mod rocks;
+
+// Trait for database that stores values.  Must be cheap to clone.
+pub trait DB: Debug + DynClone + Sync + Send + merkletree::db::DB {
+    fn load_value(&self, raw: RawValue) -> anyhow::Result<Option<Value>>;
+    // If the DB is persistent, for containers only the root needs to be stored because the
+    // Container type makes sure the underlying merkle tree is stored in the DB independently, so
+    // that it can be recovered back just with the root and the DB.
+    // If the value is RawValue and a previous non-RawValue exists, no store overwrite it.
+    // should be done. If the value is non-RawValue and a previous RawValue exists, store
```
ax0, line 24: I assume this should be:
```suggestion
    // If the value is RawValue and a previous non-RawValue exists, store should not
    // overwrite it. If the value is non-RawValue and a previous RawValue exists, store
```

src/middleware/containers.rs
```diff
@@ -1,29 +1,252 @@
...
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: SeqAccess<'de>,
+    {
+        let mut kvs = HashMap::<Value, Value>::new();
+        while let Some(mut elem) = seq.next_element::<Vec<Value>>()? {
+            if elem.len() == 2 {
+                let (v, k) = (elem.pop().unwrap(), elem.pop().unwrap());
+                kvs.insert(k, v);
+            } else {
```
ax0: Would it be worth adding a branch to catch elements of improper length?

src/middleware/containers.rs
```diff
@@ -1,29 +1,252 @@
...
+        self.root = mt.root();
+        Ok(mtp)
+    }
+    pub fn update(
+        &mut self,
+        key_raw: RawValue,
+        value: Value,
+    ) -> Result<MerkleTreeStateTransitionProof> {
+        let value_raw = value.raw();
+        store_value(self.db.as_mut(), value).unwrap();
```
robknight: In `insert` above, we use `?` to return error results, but here we are calling `unwrap`, which will panic on an error. I can't see a reason for this being different from `insert`.

robknight (approved):
Looks good apart from one question about a possible panic.

ax0 (approved):
LGTM!

src/middleware/containers.rs
ed255: thanks for spotting this!  I've replaced `unwrap()` by `?` in https://github.com/0xPARC/pod2/pull/493/commits/3916dc4d1480ba8a6be2977803a612cb5f41495c

src/middleware/db/mod.rs
ed255: Thanks!  I must have edited this at some point and then I left it in an inconsistent state.  Fixed in https://github.com/0xPARC/pod2/pull/493/commits/3916dc4d1480ba8a6be2977803a612cb5f41495c

src/middleware/containers.rs
ed255: Great idea, I've added a case to return a deserialize error on improper length in https://github.com/0xPARC/pod2/pull/493/commits/3916dc4d1480ba8a6be2977803a612cb5f41495c

src/middleware/db/rocks.rs
ed255: You're totally right, I missed this originally.  I've done a few changes so that a DB implementation can be defined outside of this crate.
- I renamed `ValueAux` to `TypedValueNoRec` to be more descriptive.
- I've added the public methods `to_bytes` and `from_bytes` to `Value` to be used by a persistent DB implementation.  Those methods internally use `TypedValueNoRec` but that's now an implementation detail.

PR #494, opened 2026-03-24
Self-referential predicate hashes as statement template args

robknight:
This adds a new statement template argument type `SelfPredicateHash`, which is specified as a reference to a predicate within the same batch/module. This is encoded into the predicate, and then during proving is replaced by the real predicate hash. The circuit verifies that the correct predicate hash is substituted.

### DIFF
--- a/src/backends/plonky2/circuits/common.rs
+++ b/src/backends/plonky2/circuits/common.rs
@@ -771,7 +771,8 @@ impl CustomPredicateEntryTarget {
         pw.set_target_arr(&self.id.elements, &predicate.batch.id().0)?;
         pw.set_target(self.index, F::from_canonical_usize(predicate.index))?;
 
-        // Replace statement templates of batch-self with (id,index)
+        // Replace BatchSelf predicates with Custom(batch, i), and
+        // SelfPredicateHash args with Literal(hash(Custom(batch, i)))
         let batch = &predicate.batch;
         let predicate = predicate.predicate();
         let statements = predicate
--- a/src/backends/plonky2/circuits/common.rs
+++ b/src/backends/plonky2/circuits/common.rs
@@ -788,10 +788,22 @@ impl CustomPredicateEntryTarget {
                     }
                     x => x.clone(),
                 };
-                StatementTmpl {
-                    pred_or_wc,
-                    args: st_tmpl.args,
-                }
+                let args = st_tmpl
+                    .args
+                    .into_iter()
+                    .map(|arg| match arg {
+                        StatementTmplArg::SelfPredicateHash(i) => {
+                            let pred_hash = Predicate::Custom(CustomPredicateRef {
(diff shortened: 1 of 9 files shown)


src/middleware/custom.rs
```diff
@@ -97,6 +101,13 @@ impl ToFields for StatementTmplArg {
                     .take(Params::statement_tmpl_arg_size())
                     .collect_vec()
             }
+            StatementTmplArg::SelfPredicateHash(index) => {
```
ed255, line 105: At the beginning of the match expression there's a comment with the encoding for each case, could you extend it with this new case?

src/backends/plonky2/circuits/common.rs
```diff
@@ -788,10 +788,22 @@ impl CustomPredicateEntryTarget {
                     }
                     x => x.clone(),
                 };
-                StatementTmpl {
-                    pred_or_wc,
-                    args: st_tmpl.args,
-                }
+                let args = st_tmpl
```
ed255, line 792: Could you extend the comment at line 774 to also mention the replacement of SelfPredicateHash arguments?

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1564,7 +1564,42 @@ fn normalize_st_tmpl_circuit(
     );
     let pred_hash_or_wc =
         PredicateHashOrWildcardTarget::new(st_tmpl.pred_hash_or_wc().elements[0], data);
-    StatementTmplTarget::new(pred_hash_or_wc, st_tmpl.args.clone())
+
```
ed255, line 1567: Could you extend the comment for the function `normalize_st_tmpl_circuit` to mention that it also replaces the SelfPredicateHash arguments?

src/backends/plonky2/circuits/mainpod.rs
```diff
@@ -1564,7 +1564,42 @@ fn normalize_st_tmpl_circuit(
...
+    let zero = builder.zero();
+    let normalized_args = st_tmpl
+        .args
+        .iter()
+        .map(|arg| {
+            let is_sph = builder.is_equal(arg.elements[0], prefix_sph);
+
+            // The predicate index is in elements[1] (same slot as WildcardLiteral).
+            // Use 0 when not SelfPredicateHash to satisfy range constraints in vec_ref.
+            let pred_index = builder.select(is_sph, arg.elements[1], zero);
```
ed255: I don't think the predicate index will be used in any `vec_ref`.  It will just be used in the encoding of `PredicateTarget` and then used for hashing; that's it.
So there's no need for this select, you can just do `let pred_index = args.elements[1]`

ed255 (approved):
This looks great!
I've added a few minor comments, mostly about keeping some code comments up to date.

PR #495, opened 2026-03-24
Podlang syntax for quoted predicates

### LINKED ISSUE #491: Podlang literal forms for predicates as values

robknight:
Closes #491 
Replaces #492

This adds the following new syntax to Podlang:

```
@native(Equal) // Refers to the predicate hash for the `Equal` statement
@external(tx, TxMutated) // refers to the predicate hash for the `TxMutated` statement from the `tx` imported module
@self(some_pred) // refers to the predicate hash for the `some_pred` hash contained within the same module
```

The first two are substituted during lowering to middleware forms. `@self` is turned into a `SelfPredicateHash` statement argument, which is resolved to a predicate hash during proving.

The `@self` syntax depends on #494.

### DIFF
--- a/src/lang/frontend_ast_lower.rs
+++ b/src/lang/frontend_ast_lower.rs
@@ -157,7 +157,9 @@ fn resolve_local_predicate(
 
 /// Lower a literal value from AST to middleware Value.
 ///
-/// This is a pure conversion that cannot fail.
+/// This is a pure conversion that cannot fail for context-free literals.
+/// Panics on ExternalPredicateHash — use `lower_literal_with_context` when
+/// external predicate references may appear (e.g. inside containers).
 pub fn lower_literal(lit: &LiteralValue) -> Value {
     match lit {
         LiteralValue::Int(i) => Value::from(i.value),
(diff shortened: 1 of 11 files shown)


src/lang/grammar.pest
```diff
@@ -49,7 +49,14 @@ custom_predicate_def = {
 
 statement_list = { statement+ }
 
-statement_arg = { literal_value | anchored_key | identifier }
+// Predicate hash literals: resolve to the predicate's identity hash as a value.
+// @native_predicate and @external_predicate are in literal_value (usable in containers).
+// @self_predicate is only in statement_arg (not in containers — deferred resolution).
+predicate_hash_native = { "@native_predicate" ~ "(" ~ identifier ~ ")" }
+predicate_hash_external = { "@external_predicate" ~ "(" ~ identifier ~ "," ~ identifier ~ ")" }
+predicate_hash_self = { "@self_predicate" ~ "(" ~ identifier ~ ")" }
```
ed255, line 57: In the PR message you wrote this example:
```
@native(Equal) // Refers to the predicate hash for the `Equal` statement
@external(tx, TxMutated) // refers to the predicate hash for the `TxMutated` statement from the `tx` imported module
@self(some_pred) // refers to the predicate hash for the `some_pred` hash contained within the same module
```
but in the grammar the keywords are different (longer).
I actually prefer the ones defined in the grammar (the longer ones)

src/lang/frontend_ast_validate.rs
```diff
@@ -588,13 +593,91 @@ impl Validator {
...
     }
+
+    /// Validate a @self_predicate reference: the name must be a custom predicate in this module.
+    fn validate_self_predicate_hash(
+        &self,
+        id: &Identifier,
+        wildcard_context: Option<(&str, &WildcardScope)>,
+    ) -> Result<(), ValidationError> {
+        // @self_predicate only makes sense inside module predicate definitions
+        if wildcard_context.is_none() {
```
ed255, line 616: If this is not a predicate definition then it's a request right?
The error `ValidationError::UndefinedPredicate` seems misleading.

src/lang/frontend_ast_lower.rs
```diff
@@ -157,8 +157,10 @@ fn resolve_local_predicate(
 
 /// Lower a literal value from AST to middleware Value.
 ///
-/// This is a pure conversion that cannot fail.
-pub fn lower_literal(lit: &LiteralValue) -> Value {
+/// This is a pure conversion that cannot fail for context-free literals.
+/// Panics on ExternalPredicateHash — use `lower_literal_with_context` when
+/// external predicate references may appear (e.g. inside containers).
+pub(crate) fn lower_literal(lit: &LiteralValue) -> Value {
```
ed255, line 163: Did you consider merging this function and `lower_literal_with_context` and taking as argument `Option<&ResolutionContext>`?
I think it would simplify the code a bit, and this is a pattern you're already using in other places of `lang`.

src/lang/frontend_ast_lower.rs
```diff
@@ -190,13 +192,85 @@ pub fn lower_literal(lit: &LiteralValue) -> Value {
...
+            .ok_or_else(|| LoweringError::PredicateNotFound {
+                name: format!("{}::{}", module.name, predicate.name),
+            })?;
+            let pred = match pred_or_wc {
+                crate::frontend::PredicateOrWildcard::Predicate(p) => p,
+                crate::frontend::PredicateOrWildcard::Wildcard(_) => {
+                    return Err(LoweringError::PredicateNotFound {
+                        name: format!("{}::{}", module.name, predicate.name),
+                    });
+                }
```
ed255: Given that you're passing `PredicateRef::Qualified` to `resolve_predicate_ref` this case should be unreachable: `resolve_predicate_ref` always returns `PredicateOrWildcard::Predicate` on `PredicateRef::Qualified`.
So you could just do 
```rust
_ => unreachable!("`resolve_predicate_ref` always returns `PredicateOrWildcard::Predicate` on `PredicateRef::Qualified`"),
```

src/lang/frontend_ast_lower.rs
```diff
@@ -190,13 +192,85 @@ pub fn lower_literal(lit: &LiteralValue) -> Value {
...
     }
 }
 
 /// Lower a statement argument from AST to BuilderArg.
 ///
-/// This is a pure conversion that cannot fail.
-pub fn lower_statement_arg(arg: &StatementTmplArg) -> BuilderArg {
+/// Context-free for most arg types. Panics on ExternalPredicateHash inside literals —
+/// use `lower_statement_arg_with_context` when external predicate references may appear.
+pub(crate) fn lower_statement_arg(arg: &StatementTmplArg) -> BuilderArg {
```
ed255, line 271: By merging `lower_literal` and `lower_literal_with_context` it becomes easy to merge `lower_statement_arg` and `lower_statement_arg_with_context`

ed255: Ah, I just noticed that the `_with_context` methods return Results but the regular ones don't, so perhaps merging them complicates things?

ed255 (approved):
I've left some small comments/suggestions but this looks good to me generally so I'm approving already :)

src/lang/frontend_ast_lower.rs
robknight: I think the difference between context-dependent and context-independent lowering is useful, especially as the former can fail. I'd prefer to keep them separated.

ed255: sounds good to me!

PR #496, opened 2026-03-25
Fix pod builder

ed255:
This PR includes several fixes and code simplifications:
- MainPodBuilder
  - Fix: It was not tracking Contains statements inherited via input pods (via public statements) when automatically generating Contains statements for Entry arguments.
  - Enhancement: Deduplicate statements
- MultiPodBuilder
    - Simplify: Remove the "statement groups" logic and instead deduplicate statements in the MainPodBuilder (which is much simpler to do)
    - Remove the "anchored key" explicit dependency tracking and instead rely on regular dependency tracking by using all the implicit operations and statements generated by MainPodBuilder as input to the solver.
    - Fix: Count and constrain custom predicates used in a pod instead of batches used

### DIFF
--- a/src/frontend/mod.rs
+++ b/src/frontend/mod.rs
@@ -137,7 +137,7 @@ pub struct MainPodBuilder {
     pub operations: Vec<Operation>,
     pub public_statements: Vec<Statement>,
     // Internal state
-    dict_contains: Vec<(Value, Value)>, // (root, key)
+    contains: Vec<(RawValue, RawValue)>, // (root, key)
 }
 
 impl fmt::Display for MainPodBuilder {
@@ -171,10 +171,16 @@ impl MainPodBuilder {
             statements: Vec::new(),
             operations: Vec::new(),
             public_statements: Vec::new(),
-            dict_contains: Vec::new(),
+            contains: Vec::new(),
         }
     }
+    pub fn len(&self) -> usize {
+        self.operations.len()
+    }
     pub fn add_pod(&mut self, pod: MainPod) -> Result<()> {
+        for st in &pod.public_statements {
+            self.track_contains(st);
+        }
         self.input_pods.push(pod);
         match self.input_pods.len() > self.params.max_input_pods {
             true => Err(Error::too_many_input_pods(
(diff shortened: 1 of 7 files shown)


src/frontend/multi_pod/mod.rs
```diff
@@ -845,33 +786,13 @@ mod tests {
...
-            "POD 1 should contain statement 2 (b_out), got {:?}",
-            solution.pod_statements[1]
-        );
+        //   Solution A:
+        //   - POD 0 (intermediate): public statements 0 (contains)
+        //   - POD 1 (output): inherits statement 0 (contains) from POD0, statement 1 (a_out),
+        //   public statement 2 (b_out)
+        //   Solution B:
+        //   - POD 0 (intermediate): statements 0 (contains), public statement 1 (a_out)
+        //   - POD 1 (output): inherits statement 1 (a_out) from POD0, public statement 2 (b_out)
```
ed255, line 794: I've removed some asserts because they were assuming a particular solution but the input has 2 valid solutions; and I think we should be happy with any of them.  In particular after the changes in this PR the solution obtained is A, whereas previously it was B.

src/frontend/mod.rs
```diff
@@ -650,9 +660,12 @@ impl MainPodBuilder {
         self.add_entries_contains(&op)?;
         let op = Self::fill_in_aux(Self::lower_op(op)?)?;
         let st = self.op_statement(wildcard_values, op.clone())?;
-        self.insert(public, (st, op))?;
+        // Skip adding the statement and operation if it already exists
+        if !self.statements.contains(&st) {
```
robknight, line 655: If the statement were first added as private, I think this might incorrectly avoid making it public.

src/frontend/multi_pod/deps.rs
```diff
@@ -5,7 +5,7 @@
 
 use std::collections::HashMap;
 
-use super::cost::AnchoredKeyId;
+// use super::cost::AnchoredKeyId;
```
robknight: Remove rather than comment this out.

src/frontend/multi_pod/solver.rs
```diff
@@ -588,13 +549,13 @@ fn try_solve_with_pods(
...
     // MainPodBuilder::add_entries_contains auto-inserts these, and we must account
     // for them in the statement count.
-    let anchored_key_used: Vec<Vec<Variable>> = (0..input.all_anchored_keys.len())
-        .map(|_| {
-            (0..target_pods)
-                .map(|_| vars.add(variable().binary()))
-                .collect()
-        })
-        .collect();
+    // let anchored_key_used: Vec<Vec<Variable>> = (0..input.all_anchored_keys.len())
```
robknight: Can this just be removed? And the comment above it?

src/frontend/multi_pod/mod.rs
```diff
@@ -217,6 +213,7 @@ impl SolvedMultiPod {
         let mut pods: Vec<MainPod> = Vec::with_capacity(solution.pod_count);
 
         for pod_idx in 0..solution.pod_count {
+            // println!("DBG SolvedMultiPod.prove pod {}", pod_idx);
```
robknight: This can be removed.

src/frontend/multi_pod/mod.rs
```diff
@@ -488,8 +464,10 @@ impl MultiPodBuilder {
     ) -> Result<Statement> {
         let stmt = self.add_operation(wildcard_values, op)?;
         if public {
-            // Index is always new (just added), so push without duplicate check
-            self.output_public_indices.push(self.statements.len() - 1);
+            let index = self.stmt_index(&stmt);
+            if !self.output_public_indices.contains(&index) {
+                self.output_public_indices.push(self.stmt_index(&stmt));
```
robknight: Do we need to call `self.stmt_index` again here given that we already have the result in `index`?

robknight (approved):
I think there's a genuine small bug here, but otherwise it looks good. Some other minor cosmetic comments. Approved assuming you fix the public statement issue!

src/frontend/mod.rs
ed255: you're right!

ed255:
Thanks for pointing out the issue https://github.com/0xPARC/pod2/pull/496#discussion_r2989183781
I've addressed it and the other comments in https://github.com/0xPARC/pod2/pull/496/commits/6394cca3536ad790e362d2f8df36361eb2ae7d3a

PR #498, opened 2026-03-31
Allow entries as args in custom statements

### LINKED ISSUE #434: Custom statements arguments can't be AnchoredKey
Currently we can write this:
```
my_pred(d) = AND(
  Equal(d.foo, 1)
  Lt(d.bar, 100)
)
```

But I can't write this:
```
my_pred(d) = AND(
  Equal(d.foo, 1)
  Lt(d.bar, 100)
  my_other_pred(d.baz)
)
```

That's because arguments to custom statements can only be literals in the current implementation.  That feels quite inconvenient.

The same happens with introduction statements.

What would be needed to support AnchoredKeys as arguments to custom statements?
Currently the circuit resolves 4 AnchoredKeys for each statement slot.  This is because native predicates use up to 4 arguments (`DictContains(new_root, old_root, key, value`).
We would need to change that to `Params::max_statement_args` which is 5 in `Params::default()`.  Then in the circuit that verifies custom predicate operations we could use the resolved AnchoredKeys.
Note that if we bump the number of resolved AnchoredKeys to 5 we could add a new argument to the `DictionaryUpdate` to include the old value which is currently missing.

The circuit cost of this is adding an extra AnchoredKey resolution for each statement slot that is unused most of the time

ed255:
Resolve https://github.com/0xPARC/pod2/issues/434

We have a limit on the number of wildcards a custom predicate can have.  This means that sometimes we want to express some logic that uses more wildcards than available.  Ideally we want a design where hard limits set by the `Params` don't limit us by declaring our intention in an unconstrained environment and having a smart compiler splitting and chaining things around.  One natural solution to overcome the limit on the number of wildcards is to pack multiple values in a dictionary and use entries instead.  For example instead of doing:
```
Foo(a, b, c) = AND(
  SumOf(a, b, c)
)
```
We do
```
Foo(d) = AND(
  SumOf(d.a, d.b, d.c)
)
```

Previously we couldn't apply this pattern on custom statements because they didn't support entries as arguments.  This PR makes the necessary changes to support that, so now we can do the following:

```
PredB(d) = AND(
  PredA(d.x)
)
```

A question remains: to create a statement of `PredB({x: 42})` we need a statement `PredA({x: 42}.x)`, but how do we generate such statement?

The solution I propose is the following:
- The "Custom" operation works as before, and it always outputs a custom statement where all the arguments are literals
- Introduce a new operation `ReplaceValueWithEntry` that allows taking any statement and replacing literal arguments with entries given a matching `Contains` statement.

With the new operation we can spend 1 statement/operation slot to transform the statement and make it use references instead, making it suitable for a custom predicate that requires it with entries as arguments.
For example given these two statements:
```
PredA(42)
Contains({x: 42}, "x", 42)
```
Using `ReplaceValueWithEntry` we can generate `PredA({x: 42}.x)`.
The API for using this operation looks like this:
```rust
// st is `Lt(5, 42)`
// d is `{a: 42}`
frontend::Operation::replace_value_with_entry(
  vec![None, Some((&d, "a"))], // Don't touch `arg[0]`, replace `arg[1]` with reference `d.a`
  st, // Statement to be transformed
) // output statement is `Lt(5, d.a), 
``` 

To use entries as arguments in custom predicates we'll spend extra slots for these transformations.  Fortunately these transformations are cheap so we can do up to `params.max_priv_statements()` of them.

The `ReplaceValueWithEntry` should be able to replace any of the arguments of a statement, which means it can take up to `BASE_PARAMS.max_statement_args` `Contains`, but it also needs to take the statement to be transformed.  For our default params this meas a bump of `max_operation_args` from 5 to 6.  This means that the circuit needs to do 6 random access to previous statement per verified operation instead of 5.  This increases the number of gates and made the circuit go to 2^17, so I decreased the `max_statements` from 48 to 46 to stay below 2^16.

**Update**: I found a simple optimization that removes gates that allows us to stay with 48 `max_statements`: For the public statements slots in the circuit we only support None and Copy which take at most 1 argument; but we were still doing `max_statement_args` random accesses per slot; so I reduced that to just 1 random access to a previous statement.

MainPodCircuit usage: degree_bits = 16, free_gates = 935
```
Gate count:
- RecCircuit: 1 x 51370.0 = 51370
- RecCircuit/MainPodVerify: 1 x 40287.0 = 40287
- RecCircuit/MainPodVerify/OpVerifyPub: 8 x 69.5 = 556
- RecCircuit/MainPodVerify/OpVerifyPub/OpCopy: 8 x 13.5 = 108
- RecCircuit/MainPodVerify/OpVerifyPub/OpNone: 8 x 11.9 = 95
- RecCircuit/MainPodVerify/OpVerifyPub/ResolveOpArgs: 8 x 44.0 = 352
- RecCircuit/MainPodVerify/OpVerifyPriv: 40 x 719.3 = 28772
- RecCircuit/MainPodVerify/OpVerifyPriv/OpCustom: 40 x 94.5 = 3778
- RecCircuit/MainPodVerify/OpVerifyPriv/MerkleDeleteOp: 40 x 15.4 = 616
- RecCircuit/MainPodVerify/OpVerifyPriv/MerkleUpdateOp: 40 x 16.4 = 654
- RecCircuit/MainPodVerify/OpVerifyPriv/MerkleInsertOp: 40 x 17.8 = 710
- RecCircuit/MainPodVerify/OpVerifyPriv/OpSignedBy: 40 x 14.1 = 562
- RecCircuit/MainPodVerify/OpVerifyPriv/OpPublicKeyOf: 40 x 14.4 = 578
- RecCircuit/MainPodVerify/OpVerifyPriv/OpNotContainsFromEntries: 40 x 14.1 = 562
- RecCircuit/MainPodVerify/OpVerifyPriv/OpContainsFromEntries: 40 x 15.8 = 632
- RecCircuit/MainPodVerify/OpVerifyPriv/OpReplaceValueWithEntry: 40 x 27.2 = 1090
- RecCircuit/MainPodVerify/OpVerifyPriv/OpMaxOf: 40 x 19.9 = 794
- RecCircuit/MainPodVerify/OpVerifyPriv/OpProductOf: 40 x 40.5 = 1618
- RecCircuit/MainPodVerify/OpVerifyPriv/OpSumOf: 40 x 20.9 = 838
- RecCircuit/MainPodVerify/OpVerifyPriv/OpHashOf: 40 x 15.3 = 612
- RecCircuit/MainPodVerify/OpVerifyPriv/OpLtToNeq: 40 x 12.7 = 508
- RecCircuit/MainPodVerify/OpVerifyPriv/OpTransitiveEq: 40 x 16.6 = 666
- RecCircuit/MainPodVerify/OpVerifyPriv/OpLtEqFromEntries: 40 x 24.3 = 972
- RecCircuit/MainPodVerify/OpVerifyPriv/OpEqNeqFromEntries: 40 x 17.1 = 686
- RecCircuit/MainPodVerify/OpVerifyPriv/OpCopy: 40 x 13.6 = 542
- RecCircuit/MainPodVerify/OpVerifyPriv/OpNone: 40 x 11.8 = 472
- RecCircuit/MainPodVerify/OpVerifyPriv/GetTaggedTblEntry: 40 x 46.0 = 1840
- RecCircuit/MainPodVerify/OpVerifyPriv/ResolveOpArgs: 40 x 249.2 = 9970
- RecCircuit/MainPodVerify/CalculateStsHash: 1 x 44.0 = 44
- RecCircuit/MainPodVerify/BuildOpAuxTbl: 1 x 9223.0 = 9223
- RecCircuit/MainPodVerify/BuildOpAuxTbl/CustomPredVerify: 8 x 258.6 = 2069
- RecCircuit/MainPodVerify/BuildOpAuxTbl/CustomPredVerify/HashTaggedTblEntry: 8 x 41.0 = 328
- RecCircuit/MainPodVerify/BuildOpAuxTbl/CustomPredVerify/CustomOpVerify: 8 x 184.6 = 1477
- RecCircuit/MainPodVerify/BuildOpAuxTbl/CustomPredVerify/CustomOpVerify/PredFromTmpl: 40 x 1.3 = 51
- RecCircuit/MainPodVerify/BuildOpAuxTbl/CustomPredVerify/CustomOpVerify/StArgFromTmpl: 40 x 21.7 = 869
- RecCircuit/MainPodVerify/BuildOpAuxTbl/MerkleProofStateTransition_32: 6 x 340.2 = 2041
- RecCircuit/MainPodVerify/BuildOpAuxTbl/MerkleProofStateTransition_32/MerkleProofExist_32: 6 x 107.2 = 643
- RecCircuit/MainPodVerify/BuildOpAuxTbl/MerkleProofStateTransition_32/MerkleProof_32: 6 x 109.0 = 654
- RecCircuit/MainPodVerify/BuildOpAuxTbl/SignedBy: 4 x 187.5 = 750
- RecCircuit/MainPodVerify/BuildOpAuxTbl/SignedBy/HashTaggedTblEntry: 4 x 2.0 = 8
- RecCircuit/MainPodVerify/BuildOpAuxTbl/SignedBy/SignatureVerify: 4 x 183.5 = 734
- RecCircuit/MainPodVerify/BuildOpAuxTbl/PublicKeyOf: 2 x 1050.0 = 2100
- RecCircuit/MainPodVerify/BuildOpAuxTbl/PublicKeyOf/HashTaggedTblEntry: 2 x 2.0 = 4
- RecCircuit/MainPodVerify/BuildOpAuxTbl/MerkleProof_32: 20 x 108.9 = 2178
- RecCircuit/MainPodVerify/BuildOpAuxTbl/HashTaggedTblEntry: 27 x 3.1 = 85
- RecCircuit/MainPodVerify/BuildCustomPredTbl: 1 x 1480.0 = 1480
- RecCircuit/MainPodVerify/BuildCustomPredTbl/CustomPred: 8 x 185.0 = 1480
- RecCircuit/MainPodVerify/BuildCustomPredTbl/CustomPred/MerkleProof_16: 8 x 56.8 = 454
- RecCircuit/MainPodVerify/VerifyInPod: 2 x 106.0 = 212
- RecCircuit/MainPodVerify/VerifyInPod/MerkleProof_6: 2 x 25.0 = 50
- RecCircuit/MainPodVerify/VerifyInPod/CalculateStsHash: 2 x 44.0 = 88
- RecCircuit/VerifyProof: 2 x 5442.5 = 10885
- CalculateStsHash: 1 x 6.0 = 6
```


ed255:
I've requested a review from @ax0 expecting a full review (with emphasis on the circuits).
I've also requested @robknight with the goal of following/discussing the design decisions and following up with the frontend/middleware changes.

src/middleware/mod.rs
```diff
@@ -815,7 +815,7 @@ impl Default for Params {
             max_input_pods_public_statements: 8,
             max_statements: 48,
             max_public_statements: 8,
-            max_operation_args: 5,
+            max_operation_args: BASE_PARAMS.max_statement_args + 1,
```
robknight: Should this be part of the `BASE_PARAMS` if we are relying on this relationship always holding? In `verify_replace_value_by_entry_circuit` we have `let st_in = &cache.op_args[BASE_PARAMS.max_statement_args];` which would be out-of-bounds if `max_operation_args` is <= `max_statement_args`.

src/backends/plonky2/mainpod/statement.rs
```diff
@@ -96,15 +98,12 @@ impl TryFrom<Statement> for middleware::Statement {
...
-                let vs: Vec<Value> = proper_args
+                let args: Vec<ValueRef> = proper_args
                     .into_iter()
-                    .filter_map(|arg| match arg {
-                        SA::None => None,
-                        SA::Literal(v) => Some(v),
-                        _ => unreachable!(),
-                    })
+                    // ValueRef from StatementArg gives Err on None, which is used for padding
+                    .filter_map(|arg| arg.try_into().ok())
```
robknight: Is there a risk that this silently swallows other errors?

robknight (approved):
Looks good to me, just one question about the params changes.

src/middleware/mod.rs
ed255: This is a good point.  Ideally we'd have a sanity check to verify that `max_operation_args` is at greater or equal than `BASE_PARAMS.max_statement_args + 1`, to guarantee that `ReplaceValueByEntry` works.

A reason why `max_operation_args` may be higher would be to allow custom predicates to have more statements.  But in hindsight this is a "freedom" we've never used (in fact this is not a current "freedom" because the arity of custom predicates is defined in `BaseParams`).

I'll move this parameter to `BaseParams`.

src/backends/plonky2/mainpod/statement.rs
ed255: There is no risk here, but from reading the code it's not obvious unless you see the implementation of `TryFrom`.  Considering that it's just a few lines I'll rewrite this to not use `try_into` so that it becomes more clear.

src/middleware/mod.rs
ed255: Changes applied in https://github.com/0xPARC/pod2/pull/498/commits/4a8b05a223707df03f1c0f85f06c339484d99ae7

src/backends/plonky2/mainpod/statement.rs
ed255: Changes applied in https://github.com/0xPARC/pod2/pull/498/commits/4a8b05a223707df03f1c0f85f06c339484d99ae7

src/middleware/operation.rs
```diff
@@ -219,6 +221,10 @@ pub enum Operation {
         /*  key    */ Statement,
         /*  proof  */ MerkleTreeStateTransitionProof,
     ),
+    ReplaceValueByEntry(
```
ax0: I think `ReplaceValueWithEntry` rolls off the tongue a bit better. 😁

src/frontend/mod.rs
```diff
@@ -566,6 +567,37 @@ impl MainPodBuilder {
                         // TODO: validate proof
                         Statement::ContainerDelete(r1, r2, r3)
                     }
+                    (ReplaceValueByEntry, &args, _) => {
```
ax0: Could you (open an issue to) document the new op in the book?

ax0 (approved):
LGTM!

src/middleware/operation.rs
ed255: Thanks for the suggestion!  Renamed in https://github.com/0xPARC/pod2/pull/498/commits/bc031672dbc2f2c4d8342af41bedce21d2553c3e

src/frontend/mod.rs
ed255: Good idea: https://github.com/0xPARC/pod2/issues/499

PR #500, opened 2026-04-17
Diagnostics for MultiPodBuilder

robknight:


src/frontend/multi_pod/diagnostics.rs
```diff
@@ -0,0 +1,519 @@
...
+                        name: "public_key_of",
+                        used: public_key_of,
+                        limit: params.max_public_key_of,
+                    },
+                    UtilizationRow {
+                        name: "distinct custom predicates",
+                        used: custom_pred_ids.len(),
+                        limit: params.max_custom_predicates,
+                    },
+                ];
```
ed255: This code looks very similar to the code in `ResourceSummary::from_costs`.
Did you consider implementing this once with `UtilizationRow` and then for `ResourceSummary` calculate `min_pods` dynamically in the `Display` impl with `lower_bound(row.used, row.limit)`?

ed255 (approved):
Overall looks good to me but I think some code could be deduplicated, take a look at my comment.

src/frontend/multi_pod/diagnostics.rs
robknight: I've refactored this to remove the duplication. The original branch was just the quickest thing that could give me feedback during txlib testing, but with some cleanup it is much less redundant.

PR #501, opened 2026-04-19
Use windowed ECAddXuGate for PublicKeyOf

robknight:
This copies the cheaper scalar multiplication approach used in `SignedBy`, saving something like 1,800 gates.


ax0 (approved):
LGTM!

PR #502, opened 2026-04-23
Add caching for MultiPodBuilder solutions

### LINKED ISSUE #489: Reuse work done by MultiPodBuilder's solver
If we are building PODs with the same shape repeatedly, then running MultiPodBuilder each time is inefficient, as solving for the distribution of statements over multiple PODs is non-trivial.

Instead, we should consider some alternatives:
- Solve for a generic "solution shape" then reuse that solution
- Solve at compile time
- Cache for identical proof tree shapes (if we can identify them)

robknight:
Closes #489 

Uses the existing POD2 memory or disk cache to store solution layouts for MultiPodBuilder. This can avoid the overhead of repeatedly solving for the layout of PODs that only differ in their concrete inputs.

Many of the PODs we want to build in real-world applications have standard layouts: we know exactly which operations will be needed, in which order, and how many input PODs will be required. This means that we should be able to reuse multi-POD layout solutions in those cases.


src/frontend/multi_pod/mod.rs
```diff
@@ -547,51 +560,193 @@ impl MultiPodBuilder {
...
+    /// solver and reuse the cached assignment. With the `disk_cache` feature
+    /// the cache persists across runs (keyed additionally by git commit);
+    /// with `mem_cache` it persists for the process lifetime.
+    ///
+    /// Robust against cache failures: a cache subsystem error (read failure,
+    /// CBOR deserialize error) or a key-mismatched entry (e.g. corrupted on
+    /// disk) is logged and falls back to a direct solve, so callers always
+    /// get either a valid solution or a typed solver error - never a panic
+    /// from the cache layer.
+    #[cfg(any(feature = "mem_cache", feature = "disk_cache"))]
```
ed255: The current code-base assumes one cache is enabled.  I don't think the crate compiles with no cache enabled, so this line seems redundant.

src/frontend/multi_pod/layout.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Two builders with equal `LayoutKey`s always produce the same solver output,
+/// regardless of concrete values, dict contents, or external pod hashes.
+/// Marked `#[non_exhaustive]` so callers must obtain instances via
+/// [`MultiPodBuilder::layout_key`] (or by deserializing) rather than struct
+/// literals, which keeps the field invariants in this module's hands.
+///
+/// [`MultiPodBuilder::layout_key`]: super::MultiPodBuilder::layout_key
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[non_exhaustive]
+pub struct LayoutKey {
```
ed255: From this PR we now have:
- SolverInput: this defines the shape of the problem and is passed to the solver to find a solution
- LayoutKey: this is the symbolic shape of the problem

LayoutKey is required because `SolverInput` has some values that are independent of the shape, like the hashes of input pods.
Then you have ways to move between the two.

The solver should be capable of finding a solution with a symbolic problem definition.  So I would like to consider a slightly different implementation: instead of having LayoutKey and SolverInput, make SolverInput symbolic.
Then there's no need to make synthetic SolverInput from LayoutKey, and I think the implementation will be more straight forward.

It would look something like this:
- MultiPod builds up the problem into MultiPodInput
- From MultiPodInput we generate InputShape
- The solver works with InputShape and returns OutputShape.  This is cached (key: InputShape, value: OutputShape)
- From OutputShape and MultiPodInput now we can build MultiPodOutput

ed255: A different way of expressing my proposal is: instead of adding `layout.rs` with new types and methods, try to rework the implementation of the solver and SolvedMultiPod so that they naturally connect with types that encode the sape without data (so the input and output of the solver is purely symbolic)

PR #503, opened 2026-04-26
Use projected statement lookup for op arg resolution

robknight:
This PR picks up a TODO from the code for cheaper op-arg resolution. Statements are 44 fields wide, and resolving an op argument used to do a column-wise random access across all 44 fields. Instead, we can precompute a 4-field Poseidon hash of each statement, do the random access over those hashes, and let a witness generator reveal the actual row (constrained by a hash-equality check). This saves around 6,000 gates, which seems like a worthwhile change.


ax0 (approved):
LGTM!

PR #504, opened 2026-04-29
Split aux tables based on size

robknight:
The aux table holds supporting data for certain operations, including Merkle container and state transition ops, `SignedBy`, `PublicKeyOf`, and custom predicate operations. It has a fixed row size, based on the largest possible value. Most aux table entries are quite small - `MerkleTreeStateTransitionClaimTarget` is 18 field elements, and all others are smaller. However, `CustomPredicateVerifyQueryTarget` is 314 field elements. Because of this outlier, the table rows are much larger than is needed for typical cases, making lookups more expensive than they need to be.

This change stores only a query hash in the aux table, for every operation kind:

- The aux table now carries a single query hash per row (`HashOutTarget`, 4 field elements) instead of a full entry, so the row size is the same for every operation and no longer driven by the `CustomPredicateVerifyQueryTarget` outlier — narrower rows, fewer Poseidon constraints per lookup.
- At table-build time each operation hashes its query inputs into that one entry: Merkle `Contains` / `NotContains`, state transition, `OpenInputStatement`, `PublicKeyOf`, `SignedBy`, and CustomPredVerify.
- Each verify circuit looks up the stored hash and compares it to a freshly-computed hash of the operation's inputs (e.g. `verify_custom_circuit` hashes (statement, op_type, op_args)), instead of unhashing a wide row and doing flattened equality.

The hashed input is prefixed with a domain-separation tag (`OperationAuxQueryKind`), distinct from the row tag (`OperationAuxTableTag`): the row tag labels the physical table segment, while the query kind separates semantic shapes that share a row tag, such as `Contains` / `NotContains` or transition Insert / Update / Delete.

### DIFF
--- a/src/backends/plonky2/circuits/mainpod/mod.rs
+++ b/src/backends/plonky2/circuits/mainpod/mod.rs
@@ -373,22 +373,14 @@ fn append_container_proofs_operation_aux_table_circuit(
         verify_merkle_state_transition_circuit(builder, merkle_transition_proof);
         let entry = MerkleTreeStateTransitionClaimTarget::from(merkle_transition_proof.clone());
 
-        tables.push_general(
-            builder,
-            OperationAuxTableTag::MerkleTransitionProof,
-            &entry,
-        );
+        tables.push_general(builder, OperationAuxTableTag::MerkleTransitionProof, &entry);
     }
     // Medium Merkle state transition proofs: verify op proof (insert/update/delete)
     for merkle_transition_proof in &merkle_transition_proofs.medium {
         verify_merkle_state_transition_circuit(builder, merkle_transition_proof);
         let entry = MerkleTreeStateTransitionClaimTarget::from(merkle_transition_proof.clone());
 
-        tables.push_general(
-            builder,
-            OperationAuxTableTag::MerkleTransitionProof,
-            &entry,
-        );
+        tables.push_general(builder, OperationAuxTableTag::MerkleTransitionProof, &entry);
     }
 }
 
--- a/src/backends/plonky2/circuits/mainpod/mod.rs
+++ b/src/backends/plonky2/circuits/mainpod/mod.rs
@@ -205,8 +205,8 @@ enum OperationAuxTableTag {
 // This is distinct from `OperationAuxTableTag`: the table tag labels the physical
 // row segment, while this kind separates semantic query shapes that share a row tag,
 // such as `Contains`/`NotContains` or transition `Insert`/`Update`/`Delete`.
-// Custom verification uses `CustomPredicateVerifyQueryTarget::hash`; its row tag is
-// sufficient because its input shape is structurally unique among aux kinds.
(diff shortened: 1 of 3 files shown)


ax0 (approved):

ed255:
If I understand this correctly, the saving comes from replacing:
- an `is_equal_flattenable` over 314 elements (CustomPredicateVerifyQueryTarget)
by
- an `is_equal_flattenable` over 4 elements (HashOutTarget) + an extra lookup over a table of hashes

I wonder if we can apply this change for all cases, and we can move back to a single table.

src/backends/plonky2/circuits/mainpod/mod.rs
```diff
@@ -466,10 +589,11 @@ fn build_operation_aux_table_circuit(
             op_type,                        // output
             op_args: entry.op_args.clone(), // input
         };
+        let query_hash = query.hash(builder);
```
ed255: All previous cases use a new method called `hash_*_query` but here we're still using the type `*QueryTarget` with its `hash` method.  Could we just add a `hash_custom_predicate_verify_query` method so that it all looks uniform?

src/backends/plonky2/circuits/mainpod/mod.rs
```diff
@@ -700,27 +841,19 @@ fn verify_contains_from_entries_circuit(
     cache: &StatementCachePriv,
 ) -> BoolTarget {
     let measure = measure_gates_begin!(builder, "OpContainsFromEntries");
-    let (aux_tag_ok, resolved_merkle_claim) =
-        aux.as_type::<MerkleClaimTarget>(builder, OperationAuxTableTag::MerkleProof as u32);
+    let (aux_tag_ok, resolved_query_hash) =
+        aux.as_type::<HashOutTarget>(builder, OperationAuxTableTag::MerkleProof as u32);
```
ed255, line 862: The `TableEntryTarget` is made generic so that you can unflatten a resolved entry into different types.  But in this PR all the resolved table entries are `HashOutTarget`.

For this reason I propose changing the implementation of `MuxTableTarget` and `TableEntryTarget` so that instead of working with `T: Flattenable` and `flattened_entry: &[Target]` it works with plain `HashOutTarget`.  This will simplify the code.
Then there's no need to keep the type `TableEntryTarget`, the result of `MuxTableTarget::get` can already be a `HashOutTarget`

ed255 (commented):
Overall looks good! This is a great optimization which removes 4433 gates!

I've made two comments about consistency and simplification after the new change, please take a look.

src/backends/plonky2/circuits/mainpod/mod.rs
ed255: I've opened a PR for this https://github.com/0xPARC/pod2/issues/527

ed255 (approved):
LGTM!

PR #506, opened 2026-05-04
Improved predicate splitter with MILP-based equivalence test

robknight:
This branch started out as a replacement of the earlier "greedy" splitter with a MILP-based version, because the greedy version was unable to find feasible solutions to some of the longer custom predicates we've been testing with recently.

However, the MILP version also struggled with performance, sometimes taking 15 seconds or longer to return a solution.

What is implemented in this PR is a dynamic programming-based approach which is able to find feasible solutions in all cases I've tested with, with a MILP version implemented for testing purposes. There is a test which randomly generates sample predicates, and then verifies that the DP approach produces feasible solutions in all cases where MILP does. In addition, there are several new tests involving longer custom predicate chains.

The DP code is much faster than the MILP code for every scenario that I've tested. From the test:

```
Parity sweep: checked=1200, milp_feasible=1177, dp_divergences=0
  MILP total=73.158783697s mean=60.965653ms max=6.204155675s
  DP   total=4.329626493s mean=3.608022ms max=25.858044ms
```

So DP is ~16x faster at the mean, and ~240x faster at the maximum.


src/lang/frontend_ast_split.rs
```diff
@@ -1,26 +1,55 @@
...
-//! When splitting a predicate, we try to group statements that use the same
-//! wildcards together. However, if a private wildcard must be used across a
-//! split boundary, it must be promoted to a public argument in the latter
-//! predicate, to ensure that it is bound to the same value in both predicates.
+//! The split is computed by an MILP that, for a given number of links K:
 //!
-//! A wildcard is "live" at a split boundary if it is used in a statement on both
-//! sides of the boundary. We want to minimize the number of live wildcards at
-//! split boundaries, to minimize the number of promotions required.
+//! - Assigns each statement to exactly one link.
```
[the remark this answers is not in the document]
ed255: I see that in variables and comments below you use the term `statement` for statement template.  Perhaps I'm being too pedantic and just saying `statement` is clear enough, so maybe ignore my suggestion.

robknight:
test

ed255:
test

src/lang/frontend_ast_split.rs
```diff
@@ -103,520 +113,717 @@ pub fn split_predicate_if_needed(
...
-    if n <= Params::max_custom_predicate_arity() {
-        return OrderingResult {
-            statements,
-            reorder_map: (0..n).collect(),
-        };
+/// Compute [`WildcardLifetimes`] for a given ordering.
+fn wildcard_lifetimes(
+    ordering: &[usize],
+    statements_using: &[Vec<usize>],
+    num_wildcards: usize,
```
ed255: I don't understand how this variable is used

src/lang/frontend_ast_split.rs
```diff
@@ -103,520 +113,717 @@ pub fn split_predicate_if_needed(
...
-    // Metric 2: Closure - prefer statements that close active private wildcards
-    // (wildcards that won't be needed by any remaining statements)
-    let closes_count = stmt_wildcards
-        .intersection(active_wildcards)
-        .filter(|w| !needed_later.contains(*w))
-        .count();
+    // Main prefix DP over non-terminal links (cap max_arity - 1).
+    let non_last_cap = max_arity - 1;
+    let mut dp: Vec<Option<(usize, usize)>> = vec![None; n + 1];
+    dp[0] = Some((0, 0));
```
ed255: This is the same DP algorithm implemented in https://github.com/0xPARC/pod2/pull/515 right?  Could you use similar names in the variables?  For example in #515 you added a struct named `DpEntry` and used names like `start`, `end`, `count` instead of `a`, `p`, `k`.

src/lang/frontend_ast_split.rs
```diff
@@ -103,520 +113,717 @@ pub fn split_predicate_if_needed(
...
-/// Calculate which wildcards are live at a split boundary
-fn calculate_live_wildcards(
-    before_split: &[StatementTmpl],
-    after_split: &[StatementTmpl],
-) -> HashSet<String> {
-    let before: HashSet<_> = before_split
+/// Numeric encoding of a predicate's wildcard graph, ready for the DP
+/// partitioner.
+pub(super) struct SplitInput {
+    pub(super) n: usize,
```
ed255: could you document what is `n`?  Or alternatively give a more comprehensive name, like `num_sts`.

src/lang/frontend_ast_split.rs
```diff
@@ -103,520 +113,717 @@ pub fn split_predicate_if_needed(
...
+        }
+    }
 
-/// Generate a refactor suggestion for wildcards crossing a boundary
-fn generate_refactor_suggestion(
-    crossing_wildcards: &[String],
-    ordered_statements: &[StatementTmpl],
-) -> Option<crate::lang::error::RefactorSuggestion> {
-    use crate::lang::error::RefactorSuggestion;
+    let mut is_original_public = vec![false; wildcard_names.len()];
```
ed255, line 704: Note: `is_original_public.len()` is the total number of wildcards

src/lang/frontend_ast_split.rs
```diff
@@ -103,520 +113,717 @@ pub fn split_predicate_if_needed(
...
+    for name in &original_public_args {
+        is_original_public[wildcard_index[name]] = true;
+    }
 
-    if crossing_wildcards.is_empty() {
-        return None;
+    SplitInput {
+        n: pred.statements.len(),
```
ed255: Note: `n` is the number of statements.

src/lang/frontend_ast_split.rs
```diff
@@ -103,520 +113,717 @@ pub fn split_predicate_if_needed(
...
+                &input.statements_using,
+                &input.is_original_public,
+                max_args,
+                REFINE_ITERATIONS,
+            );
+            let c = cost(&r);
+            (r, c)
+        })
+        .collect();
+    // Refined orderings go first so DP probes the lowest-bandwidth candidates
```
ed255: Question: First you define `orderings` and put some solutions there.
Then you make `refined_seeds` by cloning the orderings, truncating, adding shuffled orderings, and later store that in `refined`.
Finally you add the entries in `refined` back to `orderings`, but orderings already had some stuff there.
So the first entries in `orderings` may be repeated (albeit processed by `refine_ordering`).  Is this intentional?

src/lang/frontend_ast_split_milp.rs
```diff
@@ -0,0 +1,398 @@
...
+
+/// MILP variables. All binary; constraints C1..C7 in `add_structural_constraints`
+/// pin every variable other than `assign` to be an exact function of the
+/// assignment.
+struct MilpVars {
+    n: usize,
+    k: usize,
+    num_wildcards: usize,
+    assign: Vec<Vec<Variable>>,
+    u: Vec<Vec<Variable>>,
```
ed255: It would be great if `n, k, u` have self describing names.

src/lang/frontend_ast_split.rs
ed255: What I've figured out:
- `num_wildcards` is the total number of wildcards
- `statements_using` maps a wildcard to the statements that use it.

So `statements_using` must have length equal to the number of wildcards right?
Then why pass `num_wildcards`?
The thing that seems most odd to me is doing this: `statements_using.iter().enumerate().take(num_wildcards)`, why crop the tail of `statements_using`?  If my guess is correct, the cropped tail is empty.

ed255: In `ordering_excess_cost` you also do `statements_using.iter().enumerate().take(num_wildcards)`

ed255 (approved):
LGTM!  The renaming of vars from https://github.com/0xPARC/pod2/pull/506/commits/b97ac3f1b9544c75608bf4bd1a04613b7a019044 helps a lot

PR #507, opened 2026-05-04
Support "records" in Podlang

robknight:
This PR adds support for "Records". A Record is an array with a set of named indices, allowing the user to write code which accesses the entries at those indices by name. Records can be declared as having a sequence of named entries. Wildcards can be annotated with the Record's name, which allows the Podlang parser to substitute named entry accesses with index-based accesses. This compiles down to a middleware anchored key pattern, with anchored keys extended to allow integer "index" keys.

Syntax:

```!
use module 0xSOME_HASH as movies

record Song = (
  title,
  performer,
  composer,
  length
)

// Type annotation in the argument list:
IsPerformedBy(song Song, artist) = AND(
  // Anchored key:
  Equal(song.performer, artist)
  // Or we could check the contents of `song.performer` using a `Contains` statement:
  Contains(song, Song::performer, artist) // Song::performer gets desugared to the integer 1
)

// Contrived example to show literal records
IsAfricaByToto(song) = AND(
  // Entries can be given in any order and will be desugared to a sparse
  // array with the specified entries populated
  // Unknown entries cause parse errors
  Equal(song, Song(title: "Africa",
                   performer: "Toto",
                   length: 295,
                   composer: "David Paich and Jeff Porcaro"))
)

// Records from imported modules can be used, with namespace prefix
IsInSoundtrack(movie movies::Movie, song Song) = AND(
  // The `movies` module must define a `Movie` record with a `soundtrack` entry
  SetContains(movie.soundtrack, song)
)
```

Records are totally desugared, meaning that the middleware has no concept of Records.

Record *literals* are desugared to sparse POD2 arrays. Once the shallow Merkle tree feature is added, record literals will be desugared to arrays with the shallower depth option. I've provisionally added a field to `Params` to represent this depth, with a default of 8, but this is not used except for calculating the maximum number of named entries a Record can have.

Records can have at most `2 ^ max_depth_mt_records` named entries, which should align with the maximum number of entries in our shallow Merkle tree.

### DIFF
--- a/src/lang/frontend_ast_lower.rs
+++ b/src/lang/frontend_ast_lower.rs
@@ -206,6 +206,11 @@ pub(crate) fn lower_literal(lit: &LiteralValue) -> Value {
                 "ExternalPredicateHash must be lowered with context via lower_literal_with_context"
             )
         }
+        LiteralValue::RecordEntryIndex { .. } => {
+            unreachable!(
+                "RecordEntryIndex must be lowered with context via lower_literal_with_context"
+            )
+        }
     }
 }
 
@@ -279,6 +284,14 @@ pub fn lower_literal_with_context(
                 arr.insert(idx, value)?;
             }
             Ok(Value::from(arr))
+        }
+        LiteralValue::RecordEntryIndex { record, entry } => {
+            let schema = symbols
+                .records
+                .get(&record.to_string())
+                .expect("record schema validated");
+            let idx = schema.entry_index[&entry.name];
+            Ok(Value::from(idx as i64))
         }
         // All other variants are context-free
         other => Ok(lower_literal(other)),
@@ -1113,6 +1126,24 @@ mod tests {
                     }
                 }
             }
+        }
+    }
+
+    #[test]
+    fn test_record_entry_index_lowers_to_integer() {
+        // `R::Bar` resolves to integer 1 (Bar is the second entry of R).
+        let input = r#"
+            record R = (Foo, Bar, Baz)
+            my_pred(A) = AND(Contains(A, R::Bar, 7))
+        "#;
+        let module = parse_validate_and_lower_module(input, &Params::default()).unwrap();
+        let pred = &module.batch.predicates()[0];
+        let stmt = &pred.statements()[0];
+        match &stmt.args()[1] {
+            middleware::StatementTmplArg::Literal(v) => {
+                assert_eq!(v.raw(), Value::from(1i64).raw());
+            }
+            other => panic!("expected Literal at arg 1, got {other:?}"),
         }
     }
 }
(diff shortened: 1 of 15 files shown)


src/middleware/mod.rs
```diff
@@ -622,58 +636,217 @@ where
...
+/// APIs that only make sense for one variant (e.g. `Dictionary::insert` for
+/// strings) take the inner type directly, lifting the variant check out to
+/// the call site where the `match` on `Key` makes the missing arm visible at
+/// compile time.
+#[derive(Clone, Debug, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(untagged)]
+pub enum Key {
+    Str(StrKey),
+    Index(IndexKey),
+}
```
ed255, line 749: I would like to suggest removing the `StrKey` and `IndexKey` type and instead defining this enum as:
```
pub enum Key {
    Str{name: String, hash: Hash},
    Index{value: i64},
}
```
The reason for this suggestion is because I don't think we'll be using the  `StrKey` and `IndexKey` types, we'll always be using `Key`, and the code is reduced with this change.  There many methods that are extremely similar for `StrKey` and `IndexKey` and `Key`: `hash`, `raw`, `PartialEq`, `hash::Hash`, `ToFields`, `Display`.

src/middleware/mod.rs
```diff
@@ -622,58 +636,217 @@ where
...
+
+impl Key {
+    pub fn new(name: String) -> Self {
+        Key::Str(StrKey::new(name))
+    }
+    pub fn from_index(i: i64) -> Self {
+        Key::Index(IndexKey::new(i))
+    }
```
ed255: This seems redundant with `impl From<i64> for Key`

src/middleware/mod.rs
ed255: Update: I see how you use `StrKey` extensively in the Dictionary implementation in `middleware/containers.rs` and in `SignedDictBuilder` in the frontend so perhaps my suggestion would not play out nicely.
Perhaps those two uses should be using `String` instead, but that change would be out of the scope of this PR.

Please share your thoughts but I'm not requesting any action here.  In case you agree with my suggestion we can just take a note and do it in a future PR.

src/lang/frontend_ast_lower.rs
```diff
@@ -511,15 +560,55 @@ impl<'a> Lowerer<'a> {
...
     }
+
+    /// Rewrite `r.foo` to `r[i]` when `r` is a typed wildcard, using the
+    /// record schema's entry-index map. Untyped wildcards keep
+    /// `Dot`/`Bracket` keys unchanged (POD-string-key semantics).
+    fn rewrite_typed_dot_access(&self, pred: &mut CustomPredicateDef) {
+        let symbols = self.validated.symbols();
+        let Some(scope) = symbols.wildcard_scopes.get(&pred.name.name) else {
+            return;
+        };
```
ed255: could you tell me what's happening here? what's this case for early return?

ed255 (approved):
Overall LGTM!  I've left a few minor comments

PR #511, opened 2026-05-07
Fix apply_predicate_with handling of empty chain links

robknight:
`apply_predicate_with` did not work properly with long `OR` chains - it did not account for the fact that the later parts of the chain can be skipped entirely if they are empty.


ed255 (approved):
LGTM!

PR #512, opened 2026-05-12
Remove legacy wildcard key support

robknight:
This PR removes legacy support for wildcard keys, which was interfering with the new integer "index" keys.

### DIFF
--- a/src/backends/plonky2/circuits/common.rs
+++ b/src/backends/plonky2/circuits/common.rs
@@ -448,39 +448,6 @@ impl PredicateTarget {
     }
 }
 
-/// Mirrors `middleware::KeyOrWildcard`
-#[derive(Clone)]
-pub struct LiteralOrWildcardTarget {
-    pub elements: [Target; VALUE_SIZE],
-}
-
-impl LiteralOrWildcardTarget {
-    fn from_slice(v: &[Target]) -> Self {
-        Self {
-            elements: v.try_into().expect("len is VALUE_SIZE"),
-        }
-    }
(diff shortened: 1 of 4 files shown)


ed255 (commented):
Thanks for finding this!  I'm afraid that this fix is not enough.

This is the encoding of StatementArg:
https://github.com/0xPARC/pod2/blob/a2f427e7ecd7d63f16ebb67e623a84d7d0012f91/src/middleware/statement.rs#L568-L571
Since the `key` in AnchoredKey can now be an integer, it could be 0.  And then we can have an AnchoredKey that is encoded like a Literal.

The circuit uses this function to determine whether a StatementArg is a literal, and would treat an anchored key with key=0 as a literal https://github.com/0xPARC/pod2/blob/a2f427e7ecd7d63f16ebb67e623a84d7d0012f91/src/backends/plonky2/circuits/common.rs#L1557-L1562

We need to figure out a way to distinguish the encoding of StatementArg::Literal and StatementArg::Key with ak.key = integer.
One option is to add a new field element to store a tag.  But will affect performance, as it grows the number of fields required to encode a statement :(

Perhaps changing the encoding of integers isn't a bad idea after all.

Update: I've just seen you opened an issue about this https://github.com/0xPARC/pod2/issues/513

ed255 (approved):
LGTM!  As mentioned before, this is a partial fix, it allows using integer keys but not the 0 value.  We have the workaround of not using the 0 entry of an array to circumvent this issue.

PR #515, opened 2026-05-20
New multipod solver

robknight:
Replaces the existing MILP-based MultiPodSolver with a heuristic solver that uses three initial orderings, variations on Kahn's topological sort:

1) A bin-packing algorithm which tries to fit as many statements as possible into a POD, preferring statements closely related to statements in the current segment in order to avoid cross-POD publishing
2) A DFS-from-sinks topological order that places each statement's full dependency-closure immediately before it, so producer/consumer pairs stay contiguous
3) A topological sort which randomly prioritizes the next statement to add (cheap to compute and so easy to generate candidates)

This is followed by a "partitioner" which uses dynamic programming to find points at which to cut the sequence of operations, seeking to minimize the number of PODs used. Because most cuts require us to spend operations on input statement openings in the next POD, we need to be able to experiment with different cut points to find the best one.

We also have a MILP-based version which implements the same rules, and which sometimes out-performs (and never under-performs) the heuristic solver. In many cases, however, the heuristic solver is close enough to the MILP solver.

To test real-world use-cases, a test fixture from ZK-craft is included. However, this may be unnecessary given the new debugging tools in ZK-craft. I have included it for reference here but we could choose to remove it.


src/frontend/multi_pod/cost.rs
```diff
@@ -1,20 +1,32 @@
...
+    },
 };
 
-/// Unique identifier for a custom predicate in a module.
-///
-/// Uses the predicate's cryptographic hash as identifier. Two predicates with the same
-/// hash are considered identical for resource counting purposes.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+/// Depth of each POD's Merkle statement tree. Capacity is `2^TREE_DEPTH`.
+pub const TREE_DEPTH: usize = 10;
```
ed255: For consistency better do:
```
pub const TREE_DEPTH: usize = BASE_PARAMS.max_depth_public_statements_mt;
```

src/frontend/multi_pod/cost.rs
```diff
@@ -23,99 +35,246 @@ impl From<&CustomPredicateRef> for CustomPredicateId {
...
+/// Eligibility rules (see `ParamsContainers` comments):
+/// - state (proofs): small slots only support `Contains` (existence).
+///   `NotContains` always counts as medium.
+/// - transition: small slots only support `Update`. `Insert` / `Delete`
+///   always count as medium.
+/// - Additionally, any proof on a tree deeper than
+///   `params.containers.max_depth_small` counts as medium regardless of
+///   op kind.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct OperationCost {
```
ed255, line 48: I guess you treat the cost of opening input statements internally, based on the partitioning?

src/frontend/multi_pod/mod.rs
```diff
@@ -1,585 +1,368 @@
-//! Multi-POD builder for automatic statement packing.
+//! Multi-POD builder for the Merkle statement tree design.
 //!
-//! This module provides [`MultiPodBuilder`], a higher-level alternative to [`MainPodBuilder`]
-//! that automatically handles cases where statements exceed per-POD resource limits by
-//! splitting across multiple PODs.
+//! See `docs/multipod_merkle_statement_tree.md` for the design overview.
```
ed255: I don't see this file

src/frontend/multi_pod/mod.rs
```diff
@@ -1,585 +1,368 @@
...
 //!
-//! # Problem
+//! This is a parallel implementation alongside `super::multi_pod`. The old
+//! solver remains in place during development; once the new POD circuit
+//! lands, this module replaces `multi_pod` and that module is removed.
```
ed255: Is this still the case?

src/frontend/multi_pod/mod.rs
```diff
@@ -1,585 +1,368 @@
...
+        }
+        output.pod_republished_externals = new_republished;
+
+        let public_sets = intermediate_public_sets(&shape, &output);
+        let chain_total: usize = public_sets
+            .iter()
+            .take(output.pod_count.saturating_sub(1))
+            .map(|s| s.len())
+            .sum();
+        let chain_capacity = 2 << BASE_PARAMS.max_depth_public_statements_mt;
```
ed255: woops, is this supposed to be `1 << BASE_PARAMS.max_depth_public_statements_mt`?

ed255 (commented):
First pass review.  Missing:
- src/frontend/multi_pod/partition.rs
- src/frontend/multi_pod/partition_milp.rs
- src/frontend/multi_pod/shape.rs

src/frontend/multi_pod/mod.rs
robknight: Oh, yes. I thought I had fixed this already - that's how I noticed this calculation was wrong the first time.

src/frontend/multi_pod/partition_milp.rs
```diff
@@ -0,0 +1,775 @@
...
+        // when no imports are taken.
+        let sum: Expression = (0..n).map(|d| v.import_from[d][p]).sum();
+        model.add_constraint(constraint!(v.uses_chain[p] <= sum));
+    }
+
+    // (10) Input-pod cap: uses_chain + external pods <= max_input_pods.
+    for p in 0..k {
+        let ext_sum: Expression = (0..num_ext_pods).map(|e| v.ext_used[e][p]).sum();
+        model.add_constraint(constraint!(
+            v.uses_chain[p] + ext_sum <= input.params.max_input_pods as f64
```
ed255, line 410: I would assume that `uses_chain` is always `[0, 1, ... 1]`. Do you think that may not always be the case?

src/frontend/multi_pod/partition_milp.rs
```diff
@@ -0,0 +1,775 @@
...
+
+/// Random `InputShape` generator. Covers internal dep graphs (up to 2
+/// deps per statement), occasional external deps (drawn from 0-2
+/// external pods with up to 3 input statements each), and
+/// per-statement resource costs (signed_by / merkle_proofs).
+///
+/// Shared between the DP-vs-MILP parity sweep and the DP-vs-greedy
+/// sweep in `partition.rs`; both consume the same RNG seed so the
+/// two sweeps see the same input distribution.
+pub(super) fn random_input(
```
ed255, line 482: could this be a `#[cfg(test)]` function?

src/frontend/multi_pod/shape.rs
```diff
@@ -0,0 +1,146 @@
...
+/// computed from. See the helper methods for derivations.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub struct OutputShape {
+    /// Number of PODs in the chain. The last POD (`pod_count - 1`) is the
+    /// output POD and publishes a fresh tree of `output_public_indices`.
+    pub pod_count: usize,
+    /// For each POD, the statements produced there. Indices are into the
+    /// `InputShape`'s statement list. Within each POD the list is in
+    /// ascending index order.
```
ed255, line 119: Is the ascending index order a way to get determinism?  Or is it to do efficient search?

src/frontend/multi_pod/partition.rs
```diff
@@ -0,0 +1,1369 @@
...
+//! - **Picking an ordering**. Combinatorial: there's no realistic way
+//!   to search every topological order. We sample a small set of
+//!   candidates: one bin-packing ordering ([`kahn_bin_packing`]), the
+//!   DFS-from-sinks ordering ([`build_dfs_topo_order`]), and ten
+//!   random-priority orderings ([`kahn_with_priority`]).
+//!
+//! - **Cutting the ordering into segments**. Once the order is fixed
+//!   this collapses to a 1D problem: where do POD boundaries go?
+//!   Dynamic programming over prefixes solves it optimally in
+//!   O(n * W^2) where W = `max_priv_statements` (see [`run_dp`]). It
```
ed255: `max_priv_statements` is now `max_statements`, could you rename it everywhere in this PR?

src/frontend/multi_pod/partition.rs
```diff
@@ -0,0 +1,1369 @@
...
+    /// Scratch buffers used inside [`can_extend`] for deduplication.
+    /// Cleared on each call to avoid churning allocations.
+    scratch_new_producers: Vec<usize>,
+    scratch_new_ext_imports: Vec<usize>,
+    scratch_new_ext_pods: Vec<usize>,
+    /// Scratch buffer for [`can_extend`]'s tentative export check:
+    /// holds distinct in-segment dep targets of the candidate to avoid
+    /// double-decrementing `unfinished_consumers` when a candidate
+    /// names the same producer twice.
+    scratch_seen_segment_deps: Vec<usize>,
```
ed255, line 320: what is the advantage of defining these scratch buffers here instead of declaring them in `GreedyState::can_extend`?  Is it to reduce the number of memory allocations?

src/frontend/multi_pod/partition.rs
```diff
@@ -0,0 +1,1369 @@
...
+}
+
+/// Generate a per-statement priority vector by ranking a random
+/// permutation. The resulting `prio_of[s]` is `s`'s rank in the shuffle.
+fn random_priority(rng: &mut ChaCha20Rng, n: usize) -> Vec<usize> {
+    let mut shuffled: Vec<usize> = (0..n).collect();
+    shuffled.shuffle(rng);
+    let mut prio_of = vec![0_usize; n];
+    for (rank, &s) in shuffled.iter().enumerate() {
+        prio_of[s] = rank;
```
ed255: `shuffle` is a random permutation of `0..n`
And `prio_of` is a vector of size n, that has values 0..n assigned at a random ordering.
Isn't `prio_of` also a random permutation of `0..n` then?

src/frontend/multi_pod/partition.rs
```diff
@@ -0,0 +1,1369 @@
...
+
+    orderings.push(build_dfs_topo_order(input));
+
+    let mut rng = ChaCha20Rng::seed_from_u64(RANDOM_SEED);
+    for _ in 0..NUM_RANDOM_ORDERINGS {
+        let prio = random_priority(&mut rng, n);
+        if let Some(o) = kahn_with_priority(input, &prio) {
+            orderings.push(o);
+        }
+    }
```
ed255, line 544: How's the quality of results without these candidates?
I'm a bit surprised that doing a topological ordering with randomization and nothing more gives results that may be better than `kahn_bin_packing` and `build_dfs_topo_order`.

The randomized algorithms I'm familiar with work quite differently.  They would start with a valid solution (any topological ordering) and then do a loop like this:
- for N steps
  - try M random permutations while staying in the valid solution space.  Keep the best one
  
And then you can do extra things like for the first steps be more aggressive in the randomization, or allow degrading the quality (with the hope of exiting a local minima zone).

src/frontend/multi_pod/partition.rs
ed255: Also, after you have the shuffle you could call `build_pos_in_ordering` and get the same.  But anyways, I think returning `shuffled` would be equivalent.

src/frontend/multi_pod/partition.rs
```diff
@@ -0,0 +1,1369 @@
...
+    let max_segment_len = input.params.max_statements;
+
+    // dp[p] = Some((min_k, prev_a)) iff `[0..p]` partitions into `min_k`
+    // non-terminal segments, with the last segment being `[prev_a..p]`.
+    let mut dp: Vec<Option<(usize, usize)>> = vec![None; n + 1];
+    let mut workspace = DpWorkspace::default();
+
+    dp[0] = Some((0, 0));
+
+    for p in 1..=n {
```
ed255: Note to myself: I have questions about what's going on here

ed255 (commented):
Overll looks good to me.  I've left a few comments, and I would like to discuss `run_dp` in a call.

src/frontend/multi_pod/partition.rs
```diff
@@ -857,97 +857,190 @@ fn segment_feasible_with(
...
+struct DpEntry {
+    pod_count: usize,
+    prev_start: usize,
+}
+
 /// Cut `ordering` into the fewest feasible PODs under per-POD caps.
-/// Returns the boundary list `[(a0, b0), (a1, b1), ..., (ak, n)]` (one
-/// tuple per POD, in chain order), or `None` if no valid partition
-/// exists for this ordering.
+/// Returns the cut list as a `Vec<Segment>` (one entry per POD, in
```
ed255, line 898: Thanks for extending the documentation of this function.  It is clear now to me what it is doing.

ed255 (approved):
LGTM!

src/frontend/multi_pod/partition_milp.rs
robknight: It is possible (though unlikely) for a POD > 0 in the chain to perform operations which do not rely on any earlier statements. Imagine a proof tree with many leaf statements, which are independent of each other. It is possible that POD 1 does not depend on any of the statements published by POD 0, even though statements published by both are depended on by POD 2.

ed255: But `uses_chain` is used to track whether input pod slot 0 is taken by a previous pod in the chain right?  So even if a POD > 0 in the chain doesn't require past statements for an operation, it still needs the previous pod as input in order to extend the statement tree.  If it doesn't have the previous pod as input, then we have two disconnected graphs of statements, and only one of them has user-requested public statements, right?

ed255: Or maybe I misunderstood and `uses_chain` means that a previous pod is used as input AND a statement from it is required in an operation in pod `p`?

src/frontend/multi_pod/shape.rs
robknight: This ensures that two InputShapes which contain the same statements produce equal OutputShapes, so we can use an InputShape as a cache key.

src/frontend/multi_pod/partition.rs
robknight: Done

src/frontend/multi_pod/partition_milp.rs
robknight: Yes, changed

src/frontend/multi_pod/mod.rs
robknight: No, this comment is outdated. I'll remove it.

src/frontend/multi_pod/partition.rs
robknight: Good point, I will change this to simply return the shuffled version.

src/frontend/multi_pod/partition.rs
robknight: They are less useful, but in the tests there are a small number of cases where they produce a better result.

I did try a "refinement" approach which would permute an existing valid solution, but I couldn't find any cases in which this would improve an existing valid solution. The way I understand this is that it's quite hard to improve some local optimum, but still possible to find other local optima, including some which may be globally optimal via a more randomized process.

src/frontend/multi_pod/partition.rs
robknight: I think we discussed this offline, and I have added more comments which explain what is going on.

src/frontend/multi_pod/partition.rs
ed255: thanks for the explanation!  If it works in some cases, then I have no objection :)

PR #516, opened 2026-05-25
fix: merkle state transition circuit

ed255:
The test I added doesn't pass (fails in the circuit witness assignment) without the fix


src/backends/plonky2/primitives/merkletree/circuit.rs
```diff
@@ -572,13 +572,16 @@ pub fn verify_merkle_state_transition_circuit(
...
 
-        // 5.3) assert that if old_key!=empty, both old_leaf_path&new_leaf_path
-        //   should diverge at the inputted divergence level. We can check it
-        //   without having into account old_key!=empty, since if
-        //   old_key==empty, the paths would still diverge.
+        // 5.3) assert that if other_leaf!=empty, both old_leaf_path&new_leaf_path
+        //   should diverge at the inputted divergence level.
         let paths_eq_at_d = builder.is_equal(old_leaf_path[i].target, new_leaf_path[i].target);
-        builder.conditional_assert_eq(
+        // case_ii is 1 when other_leaf is not empty, 0 when it is empty
```
ed255, line 578: When old_key==empty I don't think we can guarantee that the paths diverge, because the `old_leaf_path = [0, ..., 0]` and the `new_leaf_path` can be 0 or 1 at the divergence level.

robknight (approved):
lgtm!

ed255:
After this fix with default params we have degree_bits = 16, free_gates = 365, so it's all good

PR #518, opened 2026-05-28
Make SolvedMultiPod and AbstractDep public

robknight:
These are necessary for the `pexe inspect` command to be able to inspect more details of what MultiPodBuilder returns.


ed255 (approved):
LGTM!

PR #519, opened 2026-06-01
Allow anchored keys in intro statements

robknight:
Although anchored keys cannot be used as arguments to intro operations, they can appear in intro *statements* via `ReplaceValueWithEntry`. This PR allows intro statement arguments to be `ValueRef`s, as is the case for custom statements. This makes the `IncorrectStatementArgs` error obsolete, as any statement can have any kind of `StatementArg` at any position.


ed255 (approved):
LGTM!

PR #520, opened 2026-06-03
fix ak with key=0 and literal collision

### LINKED ISSUE #513: StatementArg encoding with integer keys can allow collisions
`StatementArg::to_fields` encodes its three variants (`None`, `Literal(v)`, and `Key(root, key)`) into eight field elements with no discriminator. Two semantically different `StatementArg` values can produce identical field encodings, and the attacker chooses both sides. Because `MainPod::verify` binds only the statement hash to the proof's public input, a single proof verifies against either interpretation of the public statement.

`StatementArg::to_fields` at encodes the variants as:

```
None              -> [0, 0, 0, 0, 0, 0, 0, 0]
Literal(v)        -> [v.raw().0[0..4], 0, 0, 0, 0]
Key(root, key)    -> [root.0[0..4], key.to_fields()[0..4]]
```

The second half is zero for `Literal(v)` (padding) and also for `Key(root, IndexKey(0))`, because `IndexKey(0).to_fields()` is `[0, 0, 0, 0]`. That comes from `IndexKey::hash() = Hash::from(self.raw())` together with `RawValue::from(0_i64) = [0, 0, 0, 0]`.

Both of the following encode to the same eight field elements:

- `StatementArg::Key(r, IndexKey(0))`
- `StatementArg::Literal(v)` where `v.raw().0 == r.0`

The attacker controls `v`, so the match is constructible for any `r` they can observe. No grinding or hash inversion is needed.

`MainPod::verify` binds only the `sts_hash` (computed by flat-mapping `StatementArg::to_fields`) to the proof's public input. The same proof therefore verifies against either interpretation of the public statement: a verifier reconstructing the statement as `Key(r, IndexKey(0))` accepts it, and a verifier reconstructing it as `Literal(v)` with matching limbs also accepts it.

ed255:
Change the encoding of integers as Value to include a non-zero tag value a the highest limb (previously 0) so that it doesn't collide with the empty value.  This fixes the collision between `StatementArg::Literal(x)` and `StatementArg::Key((x, 0))`

Resolve #513

### DIFF
--- a/src/backends/plonky2/circuits/mainpod/tests.rs
+++ b/src/backends/plonky2/circuits/mainpod/tests.rs
@@ -1360,10 +1360,7 @@ fn test_custom_operation_verify_gadget_positive() -> frontend::Result<()> {
         Statement::equal(AnchoredKey::new(dict, Key::from("key")), Value::from(1234)),
     ];
     let args = vec![Value::from(dict), Value::from(1234)];
-    let expected_st = Statement::Custom(
-        custom_predicate.clone(),
-        vec![value_ref(args[0].clone()), value_ref(0)],
-    );
+    let expected_st = Statement::Custom(custom_predicate.clone(), vec![value_ref(args[0].clone())]);
 
     helper_custom_operation_verify_gadget(
         &params,
@@ -1381,10 +1378,7 @@ fn test_custom_operation_verify_gadget_positive() -> frontend::Result<()> {
         Statement::None,
     ];
     let args = vec![Value::from(dict), Value::from(0)];
-    let expected_st = Statement::Custom(
-        custom_predicate.clone(),
-        vec![value_ref(args[0].clone()), value_ref(0)],
-    );
+    let expected_st = Statement::Custom(custom_predicate.clone(), vec![value_ref(args[0].clone())]);
 
     helper_custom_operation_verify_gadget(
         &params,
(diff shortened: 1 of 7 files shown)


src/backends/plonky2/circuits/mainpod/tests.rs
```diff
@@ -1360,10 +1360,7 @@ fn test_custom_operation_verify_gadget_positive() -> frontend::Result<()> {
         Statement::equal(AnchoredKey::new(dict, Key::from("key")), Value::from(1234)),
     ];
     let args = vec![Value::from(dict), Value::from(1234)];
-    let expected_st = Statement::Custom(
-        custom_predicate.clone(),
-        vec![value_ref(args[0].clone()), value_ref(0)],
-    );
+    let expected_st = Statement::Custom(custom_predicate.clone(), vec![value_ref(args[0].clone())]);
```
ed255, line 1363: All these changes were from bugs in the test.
The custom predicates in this test only have 1 public argument, but the expected statement was taking 2 arguments, the second one being the literal 0.
Previously a literal 0 and None were indistinguishable, now they are not.

This was only a bug in the test.  Eventhough an unused argument was indistinguishable from the literal 0, the predicate is part of the statement and it already defines how many arguments the statement has, so it was unambiguous whether an argument needed to be interpreted as 0 or as None.

src/middleware/basetypes.rs
```diff
@@ -105,11 +105,13 @@ impl PartialOrd for RawValue {
     }
 }
 
+pub const VALUE_INT_TAG: F = F::NEG_ONE;
```
ed255, line 108: I chose this value for a tag because it's easily noticable if you encounter it while debugging.
This is 18446744069414584320 = 0xffffffff00000000

robknight (approved):
lgtm!

We have some content in the book which describes encodings (book/src/backendtypes.md) and which is outdated by this PR. It would be good to keep that up-to-date. Other than that, this looks good.

ed255:
> lgtm!
> 
> We have some content in the book which describes encodings (book/src/backendtypes.md) and which is outdated by this PR. It would be good to keep that up-to-date. Other than that, this looks good.

thanks for pointing this out! I've updated the docs in https://github.com/0xPARC/pod2/pull/520/commits/edf4c68f9756bfb176c331bc680c309dbfd61e9f

PR #521, opened 2026-06-03
Change order of arguments in operantions and statements

### LINKED ISSUE #509: Predicates that model functions should have the output after the input
Currently we define native predicates that model deterministic functions as having the output first and then the inputs.

For example `HashOf(c, a, b)` corresponds to `c = hash(a, b)`.  Perhaps when thinking about imperative style code this order looks fine, but if we think about the signature of the function the order is different: `hash: a -> b -> c`.

Languages that have first class predicates use inputs first, so that the inputs and ouputs go from left to right, matching type signatures of the corresponding function:
- Roqc: https://softwarefoundations.cis.upenn.edu/lf-current/Imp.html#AExp.aevalR_first_try.aevalR
- Lean4: https://brandonrozek.com/blog/functions-inductive-propositions-lean4/

The proposal of this issue is to rewrite all native predicates to have inputs first.  The same should be done for custom predicates in the examples which have arguments that could be mapped to function input/outputs or before/after concepts.

ed255:
Resolve https://github.com/0xPARC/pod2/issues/509

To follow the convention of functional and languages, we change the order of statement arguments for statements that have function equivalence so that the output comes last.

I've also taken this opportunity to make native operation and statement names uniform.

The affected statements/operations now look like this:
```
Sum(a, b, a+b)
Product(a, b, a*b)
Max(a, b, max(a, b))
Hash(a, b, hash(a, b))
PublicKey(sk, pk)
ContainerInsert(old_root, key, value, new_root)
ContainerUpdate(old_root, key, value, new_root)
ContainerDelete(old_root, key, new_root)
// Syntactic sugar for Dict, Set and Array also follow the pattern above
```


src/backends/plonky2/circuits/mainpod/mod.rs
robknight: I think these should have been switched around?

examples/main_pod_points.rs
```diff
@@ -151,7 +151,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let mut builder = MainPodBuilder::new(&params, vd_set);
     builder.add_pod(pod_alice_lvl_1_points)?;
     builder.add_pod(pod_alice_lvl_2_points)?;
-    let st_points_total = builder.priv_op(Operation::sum_of(3512 + 5771, 3512, 5771))?;
+    let st_points_total = builder.priv_op(Operation::sum(3512 + 5771, 3512, 5771))?;
```
robknight: The arguments have not been reordered here, so this fails when run.

src/backends/plonky2/circuits/mainpod/tests.rs
```diff
@@ -956,7 +956,7 @@ fn test_operation_verify_merkle_delete() -> Result<()> {
 }
 
 #[test]
-fn test_operation_verify_publickeyof_ok() -> Result<()> {
+fn test_operation_verify_publickey_ok() -> Result<()> {
```
robknight, line 959: These tests use the old (pk, sk) ordering rather than the documented (sk, pk) ordering. This works, but only because the circuit is also using the old ordering.

book/src/statements.md
```diff
@@ -55,12 +55,12 @@ The frontend also exposes the following syntactic sugar predicates.  These predi
 | 1004 | ArrayContains | `ArrayContains(root, idx, val) -> Contains(root, idx, val)` |
 | 1005 | GtEq | `GtEq(a, b) -> LtEq(b, a)`|
 | 1006 | Gt | `Gt(a, b) -> Lt(b, a)` |
-| 1009 | DictInsert | `DictInsert(new_root, old_root, key, val) -> ContainerInsert(new_root, old_root, key, val)` |
-| 1010 | DictUpdate | `DictUpdate(new_root, old_root, key, val) -> ContainerUpdate(new_root, old_root, key, val)` |
-| 1011 | DictDelete | `DictDelete(new_root, old_root, key) -> ContainerDelete(new_root, old_root, key)` |
-| 1012 | SetInsert | `SetInsert(new_root, old_root, val) -> ContainerInsert(new_root, old_root, val, val)` |
-| 1013 | SetDelete | `SetDelete(new_root, old_root, val) -> ContainerDelete(new_root, old_root, val)` |
-| 1014 | ArrayUpdate | `ArrayUpdate(new_root, old_root, idx, val) -> ContainerUpdate(new_root, old_root, idx, val)` |
+| 1009 | DictInsert | `DictInsert(old_root, key, val, new_root) -> ContainerInsert(new_root, old_root, key, val)` |
```
robknight: I think the right-hand side here is incorrect - ContainerInsert should have the same argument order as DictInsert? Same applies to the lines below.

book/src/operations.md
robknight: These are now incorrect.

book/src/statements.md
robknight: This is now incorrect.

src/frontend/multi_pod/partition.rs
robknight: Tiny nit but this comment should be updated with the new field name

robknight (changes requested):
Looks good overall but I think the `PublicKey` changes may be incomplete, and `main_pod_points` is broken. I flagged some comment/docs drifts as well.

examples/main_pod_points.rs
ed255: thanks!  I didn't try running this hoping that the github action would find errors.  But apparently we don't have a github action to run this example.  I'll add one.

ed255: Addressed in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

src/backends/plonky2/circuits/mainpod/mod.rs
ed255: Yes!  The swap in the public key operation was missing in a few places.  I've addressed this in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

src/backends/plonky2/circuits/mainpod/tests.rs
ed255: fixed in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

book/src/statements.md
ed255: thanks for spotting all the missing updates in the docs, fixed in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

book/src/operations.md
ed255: fixed in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

book/src/statements.md
ed255: fixed in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

src/frontend/multi_pod/partition.rs
ed255: fixed in https://github.com/0xPARC/pod2/pull/521/commits/bf8eb9d194276c4aa0a49d89f29c16edb19a7d5a

robknight (approved):
lgtm!

PR #522, opened 2026-06-06
Add tagging to discriminate between container serialized forms

robknight:
Prior to #493, container types were specified as untagged variants of `TypedValue`. This worked because they could be disambiguated by their internal structure, with separate struct fields (`kvs`, `set` and `array`) in dictionary, set and array respectively.

In #493 the internal structure changed so that each has an `inner` field of type `Container`. This means that serde cannot disambiguate the type using the shape of the serialized object. As a result, containers may be deserialized incorrectly (e.g. a set as a dictionary). This is not picked up by tests because operations on the containers don't care about the Rust-level wrapper type.

This PR restores tagged variants, and adds a custom deserialize function for each container, ensuring that the same invariants applied during construction are applied during deserialization (e.g. dictionary keys must be strings, set entries must have identical keys and values).

### DIFF
--- a/src/frontend/serialization.rs
+++ b/src/frontend/serialization.rs
@@ -87,7 +87,7 @@ mod tests {
             (Value::from(true), r#"{"Int":"1"}"#),
             (
                 Value::from(Array::new(vec![Value::from("foo"), Value::from(false)])),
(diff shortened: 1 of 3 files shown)


ed255 (approved):
LGTM!

As a future serialization optimization, the Set could have a custom serialization that avoids having duplicate data: for each (key,value) pair, `key == value`.  This was the previous behavior before the containers refactor.  This would come at an increased code complexity.

PR #523, opened 2026-06-10
Fix panic during parser diagnostics

robknight:
The Podlang parser generates diagnostic messages for parser errors. However, the span sizing has an off-by-one error which is triggered when the error is at the end of the file, e.g. an unexpected EOF.

### DIFF
--- a/src/lang/diagnostics.rs
+++ b/src/lang/diagnostics.rs
@@ -58,12 +58,14 @@ fn render_with_span(
 fn clamped_range(source: &str, span: &Span) -> std::ops::Range<usize> {
     let end = span.end.min(source.len());
     let mut start = span.start.min(end);
-    while start == end && start > 0 {
-        start -= 1;
-        if !source.is_char_boundary(start) {
-            continue;
+    if start == end {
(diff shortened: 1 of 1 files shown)


src/lang/diagnostics.rs
```diff
@@ -45,12 +45,29 @@ fn render_with_span(
...
+fn clamped_range(source: &str, span: &Span) -> std::ops::Range<usize> {
+    let end = span.end.min(source.len());
+    let mut start = span.start.min(end);
+    while start == end && start > 0 {
+        start -= 1;
+        if !source.is_char_boundary(start) {
+            continue;
+        }
+        break;
+    }
```
ed255: I believe the loop condition will be only hit once.
It requires `start == end`, but the first line in the body does `start -= 1`, so a `continue` will just exit the loop.
Perhaps you meant this?
```rust
    if start == end {
    while start > 0 {
        start -= 1;
        if !source.is_char_boundary(start) {
            continue;
        }
        break;
    }
    }
 ```

ed255 (approved):
LGTM!

PR #524, opened 2026-06-15
Add non-coercive container kind accessor and raw container reads

robknight:
This makes two things possible:

1) For a given value, detect what kind of container it is before using the `as_XXX` method to coerce the value, which can sometimes result in an unintended conversion
2) Given a RawValue key, get the value associated with that key (if any)

This comes from trying to update the POD request solver to the latest codebase and noticing that some things have become more difficult to do due to API changes.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -530,6 +530,19 @@ impl Value {
             _ => None,
         }
     }
+    /// The Value's actual container kind, or None for non-containers.
+    ///
+    /// Unlike `as_dictionary`/`as_set`/`as_array`, which intentionally
+    /// coerce any container into the requested wrapper, this reports
+    /// the typed variant without conversion.
+    pub fn container_kind(&self) -> Option<containers::ContainerKind> {
+        match &self.typed {
+            TypedValue::Dictionary(_) => Some(containers::ContainerKind::Dictionary),
+            TypedValue::Set(_) => Some(containers::ContainerKind::Set),
+            TypedValue::Array(_) => Some(containers::ContainerKind::Array),
+            _ => None,
+        }
+    }
     pub fn as_dictionary(&self) -> Option<Dictionary> {
         match &self.typed {
             TypedValue::Set(s) => Some(Dictionary {
--- a/src/middleware/containers.rs
+++ b/src/middleware/containers.rs
@@ -302,6 +302,16 @@ impl Container {
     }
 }
 
+/// The kind of container a `Value` holds. Returned by
+/// `Value::container_kind`, which reports the typed variant without the
+/// kind coercion that `as_dictionary`/`as_set`/`as_array` perform.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
+pub enum ContainerKind {
+    Dictionary,
+    Set,
+    Array,
+}
+
 /// Dictionary: the user original keys and values are hashed to be used in the leaf.
 ///    leaf.key=hash(original_key)
 ///    leaf.value=hash(original_value)
@@ -353,6 +363,9 @@ impl Dictionary {
     }
     pub fn get(&self, key: &StrKey) -> Result<Option<Value>> {
         self.inner.get(key.raw())
+    }
+    pub fn get_raw(&self, key: RawValue) -> Result<Option<Value>> {
+        self.inner.get(key)
     }
     pub fn prove(&self, key: &StrKey) -> Result<(Value, MerkleProof)> {
         self.inner.prove(key.raw())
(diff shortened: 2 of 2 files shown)


src/middleware/mod.rs
```diff
@@ -530,6 +530,19 @@ impl Value {
             _ => None,
         }
     }
+    /// The Value's actual container kind, or None for non-containers.
+    ///
+    /// Unlike `as_dictionary`/`as_set`/`as_array`, which intentionally
+    /// coerce any container into the requested wrapper, this reports
+    /// the typed variant without conversion.
+    pub fn container_kind(&self) -> Option<containers::ContainerKind> {
```
ed255, line 538: This will not be reliable once the value does round trips through the database.
The reason is that in the database, each container is stored by root, and we don't store duplicates.  So if we store an array and a set in the same db with the same root, on retrival of each container we'll get either two arrays or two sets.
That's why the API no longer exposes the container type publicly, so that the user is forced to remember whether the stored value was a container or an array.

This overwrite happens in two particular scenarios:
1. When using a persistent db, if we store multiple containers with the same root only one will be really stored.
2. If we're using memory db, usually we have a different db for each container.  So two independent containers would use two different databases.  But if we have nested containers they'll share the same db: imagine a dictionary that contains two entries, an array and a set: the dictionary, array and set will be stored in the same db, so if the array and set have the same root only one of them will be stored.

Because of this I'd prefer not having this method, because it can get very confusing when encountering these edge cases.

ed255: I would find it OK if this was more like a hint: you get the guarantee that this container can be interpreted as a ContainerKind, but you don't get the guarantee that it always was this container kind.  So if you really need this method I'd ask for a different name that makes this clear.

As a reminder, here are examples of containers with the same root:
- Dict{"a": "a"} == Set{"a"}
- Dict{"a": "a", "b": "b"} = Set{"a", "b"}
- Array[0] == Set{0}
- Array[0, 1] = Set{0, 1}
- Dict{} == Array[] == Set{}

robknight: Hmmm. But back before the DB change, we used to have distinct types for the different containers, and they serialized differently. Why is this no longer possible?

ed255: During serialization-deserialization roundtrip there's no loss of information.
But during db storage and retrival there is.
Previously we didn't have shared db for the data in a container, each single container had an independent in-memory store.

So now, if you make a memory container, insert some shallow values and serialize + deserialize everything will be fine.

But if you make a memory container, insert some containers that have equivalent roots, and then iterate over entries you'll see that internally something is different.  Let me prepare a test that shows this.

ed255: This is the test:
```rust
    #[test]
    fn same_roots() {
        let mut a = Dictionary::new(HashMap::new());
        a.insert(&StrKey::new("a".to_string()), &Value::from("a".to_string()))
            .unwrap();
        let a = Value::from(a);
        let mut b = Set::new(HashSet::new());
        b.insert(&Value::from("a".to_string())).unwrap();
        let b = Value::from(b);
        println!("a: {}", serde_json::to_string(&a).unwrap());
        println!("b: {}", serde_json::to_string(&b).unwrap());
        let top = Array::new(vec![a, b]);
        println!("top: {}", serde_json::to_string(&top).unwrap());
    }
```
This is the output
```
a: {"Dictionary":[["a"]]}
b: {"Set":[["a"]]}
top: [[{"Int":"0"},{"Set":[["a"]]}],[{"Int":"1"},{"Set":[["a"]]}]]
```
`a` and `b` have the same root, although one is a Dictionary and the other is a Set.  Once they're merged in the same database, they get stored under the same root, so one is treated as a duplicate of the other; and only one is stored.  The philosophy is that publicly they are the same, because the API doesn't allow you to distinguish them, on purpose.

robknight: Ok, but if I have some function which takes a Value and I want to know "is this Value a Dictionary" there is currently no way to do that. The information exists, inside the `TypedValue`, but this is private. I don't see why it should not be possible to read this information. If there is a mistaken assumption about the type of the container, this mistake has already happened by this point.

ed255: > Ok, but if I have some function which takes a Value and I want to know "is this Value a Dictionary" there is currently no way to do that.
That is currently intentional.  After https://github.com/0xPARC/pod2/pull/493 the interface of a Value no longer has the concept of having a type, instead it has the concept of being able to be interpreted as a type; and in some cases multiple interpretations are valid; so it relies on the user to figure out / remember which interpretation to make.
The information inside `TypedValue` is not 100% reliable as you can see in the edge cases I mentioned, and as you can see in the test I copied.

With the current design, the best we could do for containers is store and return a bitmask like {can_be_dict, can_be_set, can_be_array}.

If we just present a single container type to the user, with the current implementation that does deduplication, then we won't cover the edge cases, and I think that can be very confusing for the user.

PR #525, opened 2026-06-16
Fix SetInsert desugaring

robknight:
This is a small follow-on from #521, fixing a missed update to SetInsert desugaring.


ed255 (approved):
LGTM!

PR #526, opened 2026-06-17
Add ContainerKind

ed255:
Now containers store a bitmask in the `kind` field which contains a flag for valid container types that have been observed in the backing database so far.
With this information we can do proper type checking when casting a Value or a generic Container to a particular kind (Dictionary, Set or Array).

To store this information in the database we need a new key-value entry to associate the kind with the merkle root.  Because we can only store this association after we know the new merkle root, I had to refactor the database traits to work via transactions so that storing the root + bitmask happens atomically.  This corresponds to he biggest change in this PR:
- many merkle tree functions have been refactored to work on database transactions
- the merkle tree and container database-associated traits have been refactored to include transactions
- the memory and disk database-associated trait implementations have been refactored accordingly

In serialization, any of Dictionary, Set, Array and Container use the same encoding, which is a generic container encoding plus the kind which is encoded as a simple 3 character string, inspired by rwx permissions commonly displayed in Unix.

Before this PR it was responsibility of the user to remember the container kind before using it, and the user would be able to cast to the wrong type and perform insert operations leading to a Frankenstein container.  Now this is not possible because the casting operation will return an error if that container had never been seen as the casted type before.  Moreover the user now gets a very good hint as to what kind of container they have by inspecting the bitmask.

### DIFF
--- a/src/backends/plonky2/primitives/merkletree/mod.rs
+++ b/src/backends/plonky2/primitives/merkletree/mod.rs
@@ -17,7 +17,7 @@ use crate::middleware::{Hash, RawValue, EMPTY_HASH, EMPTY_VALUE, F};
 pub mod circuit;
 pub use circuit::*;
 pub mod db;
-pub use db::DB;
+pub use db::{DB, TX};
 pub mod error;
 pub use error::{TreeError, TreeResult};
 use error::{TreeError as Error, TreeResult as Result};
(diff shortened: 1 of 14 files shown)


src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -17,7 +17,7 @@ use crate::middleware::{Hash, RawValue, EMPTY_HASH, EMPTY_VALUE, F};
 pub mod circuit;
```
ed255, line 17: This diff looks really bad.  I didn't change any algorithm from this implementation.  The changes I made can be summarized by:
- Moving MerkleTree associated functions out of the struct (like verify methods)
- Moving the implementation of MerkleTree operations out of the struct and adding the arguments (db/tx and root) so that they can be called from the outside to perform merkle tree operations as part of a larger transaction.
  - The corresponding MerkleTree operations are now just wrappers to these.

src/middleware/mod.rs
```diff
@@ -155,56 +158,75 @@ impl fmt::Display for TypedValue {
...
-            TypedValue::Array(a) => {
-                write!(f, "[")?;
-                for (i, r) in a.iter().enumerate() {
-                    if i > 0 {
-                        write!(f, ", ")?;
-                    }
-                    if i == 8 {
-                        write!(f, "…")?;
-                        break;
+            TypedValue::Container(c) => {
```
ed255, line 161: This display implementation tries to be nice: if the kind has a single bit flag set, then the display will show as the underlying kind.  I thought this would be appreciated when debugging.

src/middleware/containers.rs
```diff
@@ -169,66 +288,90 @@ impl PartialEq for Container {
...
-        TypedValue::Set(Set { inner })
-        | TypedValue::Dictionary(Dictionary { inner })
-        | TypedValue::Array(Array { inner }) => {
-            if db.is_persistent() {
-                store_container_mt(db, inner)?;
-            }
-            db.store_value(v).map_err(Error::Database)?
-        }
-        _ => db.store_value(v).map_err(Error::Database)?,
+        TypedValue::Container(inner) if tx.is_persistent() => store_container_mt(tx, inner)?,
```
robknight: I think this fails to store the container kind for nested containers, if the inner container is backed by the in-memory DB. The top-level container gets its kind because its insert/update/delete sets it (e.g. `Dictionary::insert` calls `update_kind`), but a nested container reaches the DB as a value via `store_container_mt`, which copies its nodes but never calls `update_kind`.

The fix is probably `store_container_mt` to call `update_kind`, so we can't store a container without a kind.

src/middleware/db/rocks.rs
```diff
@@ -37,66 +43,169 @@ impl RocksDB {
...
+    fn load_kind(&self, root: Hash) -> anyhow::Result<Option<ContainerKind>> {
+        if root == EMPTY_HASH {
+            return Ok(Some(
+                *ContainerKind::default()
+                    .set_dictionary()
+                    .set_set()
+                    .set_array(),
+            ));
+        }
+        Ok(self.tx.get_for_update(kind_key(root), true).map(|opt| {
```
robknight: Is `get_for_update` necessary here since this is read-only? A plain `get` avoids locking/synchronization.

src/middleware/containers.rs
```diff
@@ -4,40 +4,171 @@
...
+            kind.set_set();
+        } else if s[1] != b'-' {
+            return Err(Error::custom("invalid char at pos 1"));
+        }
+        if s[2] == b'a' {
+            kind.set_array();
+        } else if s[2] != b'-' {
+            return Err(Error::custom("invalid char at pos 2"));
+        }
+        Ok(kind)
```
robknight, line 100: I think the string "---" would result in no kind bits being set here. I am not sure if this would happen in practice - should it be rejected as invalid?

src/middleware/db/rocks.rs
```diff
@@ -37,66 +43,169 @@ impl RocksDB {
...
+            opt.map(|bytes| {
+                assert_eq!(1, bytes.len());
```
robknight: I think the failure mode here is that we read a value from RocksDB which has a length other than 1 byte. Is panicking the correct response or should we return an error? Normally I would think that a failed/invalid DB read is an error but I'm not very familiar with RocksDB.

robknight (commented):
I have a few comments here, some of which might be because I don't understand the correct usage of the DB properly.

src/middleware/containers.rs
ed255: Let's consider the in-memory DB case.
We have a container with nested containers.  At storage, all entries of the top level container are stored as opaque values (that includes container entries, which are nested containers).
The non-persistent storage of values mutates the Container kind on overwrite, so the container kind should be correctly stored and OR-updated, it's just not in the kind key-value but in the Value key-value.

Nevertheless I acknowledge that having different paths for persistent and non-persistent storage is complicated.  The original idea behind this code path split was that for in-memory DB, we can optimize for performance by not traversing nested containers to merge them into a single DB, and just keep the entire nested container as an independent struct.  Do you think it would be better to remove this optimization for the sake of simplicity and better understanding of whats going on?  Given that the current code is a bit confusing I'm inclined to say yes, but I would like to hear your thoughts.
Note that this is an optimization for in-memory db, and with in-memory db you're not supposed to have massive containers, so traversing shouldn't be too bad.  Moreover, as a consumer you could manually reuse the in-memory db with the container constructors that take the db as argument to build the containers, and then you directly skip this merging.

src/middleware/db/rocks.rs
ed255: This one is tricky.  This is `load_kind` in a transaction.  `get_for_update` is the safe option (and if the assumption is wrong, you get a performance penalty).
Since it's a transaction in general I thought it's fair to think this may be read for a later update.
And indeed this is the case, this method is used `RocksTx::update_kind`, for update.

I agree it's not very clear why that is.  Do you think just adding a comment could help here?  Or something else is needed, like an API change?

src/middleware/containers.rs
ed255: I believe a `---` should not happen in practice because the only write functions for containers are under `Dictionary`, `Set` and `Array`, and each one sets one bit.

Reading a serialized container that has `---` would not give errors but you wouldn't be able to use the container as a `Dictionary` nor `Set` nor `Array`.  So I think it's safe to receive a `---` and not reject it because nothing breaks.  For this reason I would keep the implementation as is.

src/middleware/db/rocks.rs
ed255: My reasoning for the assert here is that this is an irrecoverable error.  This would be like a database corruption, or some other path writing an invalid entry to the database.  And for that the only solution I can think of is deleting the database and creating a new one.

But perhaps it's not nice to panic because pod2 is a library that can be used in many ways.  I'll change this to an error (`load_kind` already returns Result so it's an easy change)

ed255: Replaced assert by error in https://github.com/0xPARC/pod2/pull/526/commits/6249e973c5b7b9d74369ad869a2d5cf5f926722a

src/middleware/db/rocks.rs
ed255: Added clarifying comment in https://github.com/0xPARC/pod2/pull/526/commits/6e32f5830265732b8ca7d41389fb4be12c112231

src/middleware/containers.rs
ed255: I've unified both paths here https://github.com/0xPARC/pod2/pull/526/commits/6e32f5830265732b8ca7d41389fb4be12c112231
Now it should be simpler to follow what's going on and reason about the correctness of the implementation.

I've also added a test that moves two independent in-memory containers into entries in a container (so the two in-memory containers will become nested), which covers the code we discussed.  I believe there was a bug previously: in the case of persistent storage, the kind of nested containers was not being stored, I've fixed that.  Perhaps that's what you were mentioning!

robknight (approved):
lgtm!

PR #528, opened 2026-06-22
Simplify MuxTableTarget, cheaper query hashes

### LINKED ISSUE #527: Simplify MuxTableTarget
https://github.com/0xPARC/pod2/pull/504#discussion_r3414737889

robknight:
Closes #527 

Two related changes to the operation aux-table machinery, building on #504:

1. Cheaper query hashes. The custom-predicate and open-input-statement aux queries now hash over *already-materialized component hashes* instead of re-flattening wide statement structures. The custom query went from hashing a ~314-element flattening of `(statement, op_type, op_args)` to hashing `statement_hash (4) + op_type (6) + op_arg_hashes (N*4)`, where those component hashes are reused from work the circuit already does (`vec_ref_projected` now returns the hash it computes; `statement_hashes[i]` is already available).

2. Collapse the aux-table indirection. With every entry now a 4-element query hash, the table no longer needs the generic hash-and-unhash layer it inherited. `MuxTableTarget` stores query hashes directly, `get` returns a bare `HashOutTarget`, and `TableEntryTarget` is removed. This addresses issue #527.

Collapsing the indirection drops the per-row `OperationAuxTableTag` and the per-verify tag check. That is sound because domain separation now lives entirely in the query hash: every `hash_*_query` helper routes through `operation_aux_query_hash`, which prepends a unique `OperationAuxQueryKind`.

Removing this indirection reduces gate count by 2200 or so.

### DIFF
--- a/src/backends/plonky2/circuits/mainpod/mod.rs
+++ b/src/backends/plonky2/circuits/mainpod/mod.rs
@@ -212,9 +212,9 @@ enum OperationAuxQueryKind {
     MerkleTransition = 3,
     MerkleDelete = 4,
     OpenInputStatement = 5,
-    PublicKeyOf = 6,
-    SignedBy = 7,
-    CustomPredVerify = 8,
+    CustomPredVerify = 6,
+    PublicKeyOf = 7,
+    SignedBy = 8,
 }
 
 fn operation_aux_query_hash(
@@ -294,7 +294,10 @@ fn hash_merkle_delete_query(
     )
 }
 
-fn hash_statement_query(builder: &mut CircuitBuilder, st_hash: &HashOutTarget) -> HashOutTarget {
+fn hash_open_input_statement_query(
+    builder: &mut CircuitBuilder,
+    st_hash: &HashOutTarget,
+) -> HashOutTarget {
     operation_aux_query_hash(
         builder,
         OperationAuxQueryKind::OpenInputStatement,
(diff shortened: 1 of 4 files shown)


src/backends/plonky2/circuits/mainpod/mod.rs
```diff
@@ -216,6 +214,7 @@ enum OperationAuxQueryKind {
     OpenInputStatement = 5,
     PublicKeyOf = 6,
     SignedBy = 7,
+    CustomPredVerify = 8,
```
ed255: This is a nitpick, but could you move this one to position 5 (and value 5), so that it matches the order in the table?
See https://github.com/0xPARC/pod2/blob/10e8cc92af1a0795e97f2ee13e77c6077807635b/src/backends/plonky2/mainpod/operation.rs#L118

src/backends/plonky2/circuits/mainpod/mod.rs
```diff
@@ -295,11 +294,11 @@ fn hash_merkle_delete_query(
     )
 }
 
-fn hash_statement_query(builder: &mut CircuitBuilder, st: &StatementTarget) -> HashOutTarget {
+fn hash_statement_query(builder: &mut CircuitBuilder, st_hash: &HashOutTarget) -> HashOutTarget {
```
ed255: would you mind renaming this method to `hash_open_input_statement_query`?  I realized this method uses the `OperationAuxQueryKind::OpenInputStatement` so it's more specific than just hashing a statement.

ed255 (approved):
LGTM!  It's a very nice optimization to reuse the hashed statements and operation arguments for the queries to the aux table.

I've left 2 minor requests, please take a look.

ed255 (approved):

PR #529, opened 2026-07-28
improve containers api

ed255:
- Implement Display for Container (by moving the implementation from Value)
- Expose the inner container for Dict, Set and Array
- fix: Support casting a Raw EMPTY_VALUE to Container, Set, Dict and Array


robknight (approved):
lgtm!

PR #531, opened 2026-08-10
Share output-statement checks across native operations

robknight:
Each native-operation checker previously rebuilt an expected output statement and compared it independently, even though the argument checks were derived from the same cached inputs. This duplicated a significant number of constraints across the `MainPod` circuit.

During input resolution, `StatementCache` represents each operation argument as a normalized equation pairing the statement argument that identifies the input with the concrete value used by the operation. For a referenced container entry, for example, this is roughly `anchored_key(origin, key) = entry_value`; the left-hand side is what must appear in the output statement.

This change precomputes whether each output argument matches those left-hand sides and whether the remaining arguments are correctly padded with `None.` Individual operation checkers now only verify the predicate and select the shared argument check for that predicate's arity.

This reduces the total gate count by 6,282 gates.


ed255 (approved):
nice! LGTM!

PR #532, opened 2026-08-10
Speed up predicate splitting during module lowering

robknight:
Podlang modules commonly contain families of predicates with the same statement and wildcard-usage structure. Previously, the splitter repeated the complete ordering and partition search for every predicate.

This PR introduces a `SplitShape` representation containing the inputs that determine a link assignment and caches assignments by that shape. Module lowering shares one `SplitSearchCache`, allowing equivalent predicates to reuse the result.

It also reduces work within each uncached search by:

- accepting the source ordering immediately when it has zero excess cost and partitions at the arity-only lower bound
- testing the first cost-zero refined ordering immediately and stopping when it reaches that lower bound
- computing wildcard boundary crossings with a difference array instead of scanning every wildcard at every boundary
- factoring shape-dependent partition and cost operations into `SplitInput` helpers
- adding debug timing for parsing, AST construction, validation, lowering, and predicate splitting

The early exits only apply when an assignment reaches the arity-only minimum link count, so no later candidate could produce a better result.


src/lang/frontend_ast_split.rs
```diff
@@ -723,56 +841,68 @@ pub(super) fn prepare_split_input(pred: &CustomPredicateDef) -> SplitInput {
...
+    for seed in refinement_seeds {
+        let result = refine_ordering(
+            seed,
+            &input.shape.statements_using,
+            &input.shape.is_original_public,
+            max_args,
+            REFINE_ITERATIONS,
+        );
+        let result_cost = input.excess_cost(&result);
+        if result_cost == 0 && !shortcut_attempted {
```
ed255, line 896: This conditional is true only on the first occurrence of `result_cost = 0`.
Is there an intuition of why the first occurrence `result_cost = 0` may be valid, compared to other occurences?

ed255 (approved):
LGTM!

PR #533, opened 2026-08-13
More Podlang parser performance improvements

robknight:
This PR builds on #532 to speed up the Podlang parser using various tweaks and bottleneck removals.

Mostly this means parallelizing using Rayon where possible, eliminating some quadratic search behaviour in `CustomPredicateBatchBuilder`, and making sure that fast paths/early exits are taken whenever they can be.

This cuts `pexe build` by about 50% for the microverse example. The remaining cost is partly in Merkle tree building, which we can address by bulk-building Merkle trees rather than iteratively inserting each new entry. Since the Merkle tree implementation is a dependency for many parts of the library, I will leave this for another PR.


src/frontend/custom.rs
```diff
@@ -142,10 +148,24 @@ impl CustomPredicateBatchBuilder {
...
+    pub fn predicates(&self) -> &[CustomPredicate] {
+        &self.predicates
+    }
+
+    /// Append a prebuilt predicate, skipping the checks `predicate()`
+    /// performs.
+    pub fn push_predicate(&mut self, predicate: CustomPredicate) {
+        self.predicate_index_by_name
+            .entry(predicate.name.clone())
+            .or_insert(self.predicates.len());
```
ed255: maybe we should error on duplicate names, because then getting a predicate by name back can have surprising results, and I don't think there's a good reason to support duplicate names (it just makes debugging harder)

src/lang/frontend_ast_split.rs
```diff
@@ -124,6 +130,96 @@ pub fn split_predicate_if_needed(
...
+        .par_iter()
+        .map(|(name, input)| search_link_assignment(name, input, params))
+        .collect();
+    let assignments: Vec<LinkAssignment> = searched.into_iter().collect::<Result<_, _>>()?;
+
+    // Past this point every predicate was validated, so `inputs` and
+    // `rep_of_pred` have an entry for each of them.
+    if let Some((_, err)) = validation_failure {
+        return Err(err);
+    }
```
ed255: why doesn't this happen as early as possible?  Just right after assigning `validation_failure`

src/lang/frontend_ast_validate.rs
```diff
@@ -592,12 +593,6 @@ impl Validator {
         let pred_name = stmt.predicate.predicate_name();
         let pred_span = stmt.predicate.span();
 
-        let wc_names = match wildcard_context {
```
ed255, line 595: is this removed because this wildcard name validation already happens elsewhere?

ed255 (approved):
LGTM!  I've left a few comments/questions

src/lang/frontend_ast_validate.rs
robknight: Yes — moved, not removed. The collision check depends only on the predicate's wildcard scope, not on any individual statement, so it was hoisted from once-per-statement inside `validate_statement` to once-per-predicate in `validate_custom_predicate`.

src/lang/frontend_ast_split.rs
robknight: Good point, we can simplify this by returning earlier.

src/frontend/custom.rs
robknight: Done.

PR #534, opened 2026-08-17
Split disjunction trees

robknight:
This changes the Podlang predicate splitter, so that for `OR` predicates with `>= 2 * max_statement_args`, we split the predicate into a balanced tree of predicates rather than a chain.

This ensures a constant number of statements required, rather than being position-dependent. This is driven by cases of very large predicates, where many statements are required to reach the branches at the end of the chain.

In cases where some statements really are more likely than others, Podlang authors/tools can manually split the predicate so that high-priority branches appear first.


ed255 (approved):
LGTM!

When I first read the new `SplitTreeInfo` and how it's processed to construct the statement I noticed that it contains the full graph representing the tree.  I wondered if the same data structure can be used for chains so that:
- splitting takes different paths (one for chain, one for tree)
- constructing a statement takes a single path by processing a generic graph structure

But I think there are a few details that make this not direct in the current implementation, for example the chain contains a reorder_map but the tree doesn't.

It would be interesting to explore a structure that can hold both cases so that the statement-building code has a single path, but merging the implementation in this PR is fine by me.

PR #538, opened 2026-08-19
Test timings and effect of enabling avx2

robknight:
Enable `mimalloc` and `avx2` in CI, which results in ~13% faster proofs, and therefore faster tests.

Also adds some code for retrieving more detailed timing information from proofs, which is behind an existing feature flag, and configuration for enabling this in a CI workflow (off by default).


.github/workflows/scripts/summarise-timings.py
```diff
@@ -0,0 +1,86 @@
...
+if __name__ == "__main__":
+    args = sys.argv[1:]
+    if len(args) == 2:
+        with open(args[0]) as f:
+            base = parse(f)
+        with open(args[1]) as f:
+            other = parse(f)
+        if not base or not other:
+            sys.exit("no timing lines found in one of the logs")
+        compare(base, other, "base", "avx2")
```
ed255, line 77:
```suggestion
        compare(base, other, "base", "other")
```
I think this is supposed to be generic?

ed255 (approved):
LGTM

PR #539, opened 2026-08-19
Avoid races when concurrently building pod2 disk cache

robknight:
When running two processes which both use the same POD2, it's possible for them to race over the disk cache, not noticing that the other process is also trying to write to the cache. This PR ensures that if one process has opened the file to start writing, the other process will wait for it to finish.


src/cache/disk.rs
```diff
@@ -37,82 +90,63 @@ pub fn get<T: Serialize + DeserializeOwned, P: Serialize>(
...
     create_dir_all(&cache_dir)?;
 
     // Store the params.json if it doesn't exist for better debuggability
     let params_path = cache_dir.join("params.json");
-    loop {
-        if !params_path.try_exists()? {
-            // First write the file to .tmp and then rename to avoid having a corrupted file if we
-            // crash in the middle of the write.
-            let params_path_tmp = cache_dir.join("params.json.tmp");
-            let mut file = File::create(&params_path_tmp)?;
```
ed255, line 57: Was the issue that `File::create` was truncating the file before locking, and then destroying the contents that a different process could have been writing while holding the lock?

ed255 (approved):
LGTM! Thanks for fixing this!

PR #540, opened 2026-08-26
Fix MultiPodBuilder external input accounting and scheduling

robknight:
Correct several related issues in how MultiPodBuilder accounts for and schedules external inputs.

- Count every generated POD’s predecessor as an input POD, even when no statements are imported from it.
- Preserve concrete external POD identity instead of identifying PODs by their public-statement hash.
- Avoid counting external opening statements twice against the statement limit.
- Add an external-opening-first candidate ordering so the first POD can use its available external input slots and carry those statements forward through the chain.

Both the heuristic and MILP partitioners now use consistent input-POD and statement accounting. Diagnostics report total input-POD usage, and materialization asserts that its concrete input list agrees with the solver.


src/frontend/multi_pod/cost.rs
```diff
@@ -46,6 +46,9 @@ impl From<&CustomPredicateRef> for CustomPredicateId {
 ///   op kind.
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct OperationCost {
+    /// True for a statement copied directly from an external input POD.
+    #[serde(default)]
+    pub is_external_opening: bool,
```
ed255, line 51: Why is this boolean and the others are usize?
Since this is an operation, the usize fields will be 0 or 1, so they could be bool as well?  But then I guess adding is easier if they are usize; yet this new field is bool.

src/frontend/multi_pod/partition.rs
```diff
@@ -420,13 +420,11 @@ impl GreedyState {
         let n_producers = self.prev_pod_producers.len() + self.scratch_new_producers.len();
         let n_ext_imports = self.external_imports.len() + self.scratch_new_ext_imports.len();
         let n_ext_pods = self.external_pods.len() + self.scratch_new_ext_pods.len();
-        if !tree_imports_ok(n_producers, n_ext_imports, n_ext_pods, params) {
+        if !tree_imports_ok(self.a == 0, n_producers, n_ext_imports, n_ext_pods, params) {
```
ed255, line 423: could you rename this `a` field into something more meaningful?

src/frontend/multi_pod/partition.rs
```diff
@@ -522,9 +520,46 @@ fn random_priority(rng: &mut ChaCha20Rng, n: usize) -> Vec<usize> {
     prio_of
 }
 
+fn external_opening_pod(input: &InputShape, statement: usize) -> Option<usize> {
+    if !input.costs[statement].is_external_opening {
+        return None;
+    }
+    match input.dep_edges[statement].as_slice() {
+        [AbstractDep::External { pod, .. }] => Some(*pod),
+        _ => None,
```
ed255, line 529: is this case reachable?  Because we just saw that `is_external_opening = true`.  If it's unreachable I recommend using `unreachable!()` to be explicit about it.

ed255: In fact, is `input.costs[statement].is_external_opening` fully coherent with `input.dep_edges[statement].as_slice()`?  If so, wouldn't this be enough?
```rust
n external_opening_pod(input: &InputShape, statement: usize) -> Option<usize> {
    match input.dep_edges[statement].as_slice() {
        [AbstractDep::External { pod, .. }] => Some(*pod),
        _ => None,
    }
}
```

ed255 (approved):
LGTM! I've left some minor comments / questions, please take a look before merging.

PR #541, opened 2026-08-27
Support sparse array literals in Podlang

robknight:
POD2 arrays can be sparse, with arbitrary indices. Podlang literal array syntax only supports zero-indexed arrays, with no missing indices between elements.

This PR adds syntax for sparse arrays:

```
[10: 20, 5: 41, 11, 85]
```

Where an index is given, that index must be used. Where an index is not given, the index is that of the immediately preceding entry plus one. If no indices are given, the first entry has the index 0 and each counts up from there, which is the same behaviour as the existing array literal.


src/lang/pretty_print.rs
```diff
@@ -699,4 +701,29 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_round_trip_sparse_array_literal() {
```
ed255, line 706: notice that there's a canonical representation and for some arrays, non-canonical ones.
For example the array `[0: "a", 1: "b"]` would not be equal after the round trip, as it would become `["a", "b"]`.

src/middleware/containers.rs
```diff
@@ -326,8 +326,13 @@ impl fmt::Display for Container {
             write!(f, "]")
         } else if self.kind() == *ContainerKind::default().set_array() {
             let a = self.clone().as_array().expect("array");
+            // Merkle iteration is ordered by key hash. Read one entry beyond the
```
ed255, line 329: I don't think that's true.
Merkle iteration is ordered by key.  For some value types, the key is hashed (string), for others it's not (i64).

ed255: Did you observe this implementation showing arrays with elements in an unexpected order?

ed255 (approved):
Overall LGTM, but I've left a few comments, please take a look before merging.

src/lang/pretty_print.rs
robknight: I think it's acceptable that the pretty-printed output can change from a non-canonical form to a canonical form, so long as the parsed result is the same in both cases.

src/middleware/containers.rs
robknight: You're right, the comment was incorrect: integer keys aren't hashed. However, I verified that array iteration isn't numeric, because Merkle paths consume key bits least-significant first; indices 0–7 iterate as `0, 4, 2, 6, 1, 5, 3, 7`.

I've corrected the comment and removed sorting from the bounded `Display` preview, which only shows up to 8 elements anyway. Podlang rendering now has a separate, complete container printer, where arrays are sorted by index before rendering.

src/lang/pretty_print.rs
ed255: I also consider it acceptable; my comment was an observation :)

PR #542, opened 2026-08-27
Fix Podlang parsing and validation edge cases

robknight:
This tightens Podlang parsing and validation around several ambiguous or lossy cases.

- Treat reserved words and boolean literals as complete tokens, so identifiers such as `record_count`, `true_flag`, and `falsehood` parse correctly.
- Format `REQUEST` blocks as whitespace-separated statements so pretty-printed requests can be parsed again.
- Reject duplicate dictionary keys, including duplicates in nested literals, before lowering can discard a value.
- Reject empty `REQUEST` blocks.
- Reject custom and intro predicate names that collide with native or syntactic-sugar predicates.
- Reject duplicate module aliases before their predicates and records can be mixed in the symbol table.
- Correct container statement argument documentation.

This also removes the unused validation-diagnostics path now that invalid constructs such as empty requests are reported as validation errors, and adds targeted parser, round-trip, validation, and diagnostic coverage.

---


src/lang/grammar.pest
```diff
@@ -119,7 +120,11 @@ record_entry_index = {
 
 // Primitive literal types
 literal_int = @{ "-"? ~ ASCII_DIGIT+ }
-literal_bool = @{ "true" | "false" }
+// `word_end` matters here as much as in `reserved_identifier`: without it
+// `true_flag` parses as the bool `true` followed by junk, and `statement_arg`
+// has already committed to `literal_value` by then, so there is no fallback to
+// `identifier`.
+literal_bool = @{ ("true" | "false") ~ word_end }
```
ed255, line 127: I don't understand this.  Without this change, if we have `true_flag` I get that the parser will get the token `true` but afterwards there will be `_flag` which the grammar rejects, so we would get an error right?

src/lang/pretty_print.rs
```diff
@@ -159,41 +159,42 @@ impl PrettyPrint for Value {
...
     indent: usize,
     batch_context: Option<&CustomPredicateBatch>,
 ) -> std::fmt::Result {
-    let base_indent = " ".repeat(indent);
     let statement_indent = " ".repeat(indent + 4);
+    for statement in statements {
+        write!(w, "\n{}", statement_indent)?;
+        statement.fmt_podlang_with_batch_context(w, batch_context)?;
+    }
+    write!(w, "\n{})", " ".repeat(indent))
```
ed255, line 183: is this indent after the new line necessary?

ed255 (approved):
LGTM!

src/lang/grammar.pest
robknight: Yes, exactly - we would get an error, but the error is a bug: `true_flag` should be accepted as an identifier. Without the boundary check, `literal_bool` successfully consumes `true`, so `statement_arg` selects that alternative; the later failure on _flag doesn't cause it to retry as an identifier. With `word_end`, the boolean alternative fails immediately, allowing the identifier alternative to match the whole name.

src/lang/pretty_print.rs
robknight: It aligns the closing `)` with the beginning of the block when the caller requests nonzero indentation. With the default indent = 0, it adds no spaces.

ed255: oh, I didn't register the trailing `)` in the string.

PR #543, opened 2026-08-27
Preserve predicate wildcard calls when splitting Podlang predicates

robknight:
This PR fixes the handling of higher-order calls such as `P(A)` when a Podlang predicate is split into multiple predicates.

Previously, wildcard calls were parsed as ordinary local predicate references and inferred from their names during lowering. That became ambiguous after splitting introduced generated continuation predicates. Predicate-position uses were also missing from the splitter's wildcard dependency graph, so a split piece could fail to declare or promote the predicate-valued wildcard it used.

This change:

- tags validated higher-order calls explicitly as wildcard references
- tags generated continuation calls separately from source-level predicate references
- resolves these reference kinds directly during lowering instead of inferring them from names
- chooses continuation names that do not collide with source predicate names
- counts predicate-position wildcard uses during chain and tree splitting
- carries predicate-valued wildcards between split pieces when required

The change includes splitter, lowering, and end-to-end regression coverage for unconstrained predicate wildcards, generated-name collisions, and both chain and tree splits.

---


src/lang/frontend_ast_lower.rs
```diff
@@ -53,34 +51,28 @@ pub fn resolve_predicate_ref(
             )))
         }
         PredicateRef::Local(id) => resolve_predicate(&id.name, symbols, context),
+        PredicateRef::Wildcard(id) => Some(PredicateOrWildcard::Wildcard(id.name.clone())),
+        PredicateRef::Generated(id) => {
+            let ResolutionContext::Module { reference_map } = context else {
+                unreachable!("split continuations only exist while lowering a module")
+            };
+            resolve_local_predicate(&id.name, reference_map).map(PredicateOrWildcard::Predicate)
```
ed255, line 59: In this match, `::Local` uses `resolve_predicate` and `::Generated` uses `resolve_local_predicate`.
Maybe the name `PredicateRef::Local` is misleading, and instead should be `PredicateRef::Unqualified` or `::Simple` or `::General`?  Because reading the code I see that it can also be a native predicate.

ed255 (approved):
LGTM!  Take a look at my comment about the enum case name.

src/lang/frontend_ast_lower.rs
robknight: I've changed it to `Unqualified`.

PR #545, opened 2026-08-28
Make MainPodBuilder mutations atomic

robknight:
Stage generated statements and operations before committing them, and pre-check limits so failed builder calls leave existing state unchanged.

---


ed255 (approved):
LGTM!

PR #546, opened 2026-08-28
Make frontend lookups fallible

### LINKED ISSUE #401: Use TryFrom instead of From in conversions that can panic
These two implementations can panic if the dictionary doesn't contain the key:
https://github.com/0xPARC/pod2/blob/b11dd9317bb623467ce95c9d8d06292491bb83b6/src/frontend/operation.rs#L71-L87

Unfortunately just replacing the `From` with `TryFrom` breaks the usage of the operation functions as they take arguments which are `Into<OperationArg>`, so we should find a solution that still allows us to have ergonomic examples.

See https://github.com/0xPARC/pod2/pull/394#discussion_r2302131919

robknight:
Propagate SignedDict storage errors, bounds-check input POD access, and replace panicking operation-entry conversions with the fallible entry helper.

---


ed255 (approved):
LGTM!  Those TryFrom were long due! https://github.com/0xPARC/pod2/issues/401

PR #547, opened 2026-08-28
Validate custom predicate construction

robknight:
Check statement-template arity, batch limits, and pushed predicates before mutating the batch, and propagate construction errors during lowering.

---


src/frontend/custom.rs
```diff
@@ -455,10 +492,135 @@ mod tests {
...
+                assert!(builder.predicates().is_empty());
+            }
+        }
+    }
+
+    #[test]
+    fn push_predicate_enforces_batch_size() {
+        let params = Params::default();
+        let mut builder = CustomPredicateBatchBuilder::new(params.clone(), "batch".into());
+        for index in 0..Params::max_custom_batch_size() {
```
ed255, line 546: `Params::max_custom_batch_size()` is 2^16 = 65k.  At first I thought this may make this test slow, but after running it I see it runs very quickly even on debug (0.10s on M1).
So this comment is just an anecdote.

ed255 (approved):
LGTM!

PR #548, opened 2026-08-28
Enforce public statement limits during construction

robknight:
Check per-POD publication limits and inherited public-tree capacity in insert, operation, reveal, and extension paths without partially mutating the builder.

---


src/frontend/mod.rs
```diff
@@ -215,10 +215,46 @@ impl MainPodBuilder {
...
+            self.input_pods
+                .first()
+                .map_or(0, |pod| pod.public_statements.len())
+        } else {
+            0
+        };
+        let extended_public_count = inherited_public_count + new_public_count;
+        let extended_capacity = 2usize.pow(BASE_PARAMS.max_depth_public_statements_mt as u32);
+        if extended_public_count > extended_capacity {
+            return Err(Error::too_many_public_statements(
```
ed255: This error and the previous one look the same.  Will that be confusing when encountering it?  Because you'll see a max capacity but you'll not know the value corresponds to max_public_statements or max leafs of the public merkle tree

ed255: Maybe having two error types, or adding an extra field in the error type to say the name of the parameter that is exceeded

ed255 (approved):
Overall LGTM but take a look at my suggestion.

PR #550, opened 2026-08-31
Build Merkle trees directly from sorted leaves

robknight:
This PR changes how Merkle trees are built from `HashMap`s. Previously, we would iterate over the `HashMap`, inserting each node into a full Merkle tree, repeatedly traversing existing paths and writing intermediate nodes that were superseded by later insertions. This is expensive when we have to build a Merkle tree containing thousands of nodes, e.g. when building a large custom predicate module.

Building from sorted leaves performs one pass over the final tree structure and stores each leaf and intermediate node exactly once, substantially reducing constructor work and database traffic.

- Replace repeated insertion in MerkleTree::new_with_db with direct bulk construction.
- Sort leaves by path and recursively construct the canonical tree.
- Store only nodes reachable from the final root, avoiding transient nodes created by incremental insertion.
- Add tests comparing roots, iteration results, and proofs with incremental construction.
- Cover empty trees, randomized input order, maximum shared prefixes, and stored-node counts.


src/backends/plonky2/primitives/merkletree/mod.rs
```diff
@@ -620,6 +621,31 @@ impl MerkleTree {
     }
 }
 
+/// Build the canonical subtree for path-sorted leaves and store every
+/// reachable node exactly once.
+fn build_subtree(tx: &mut dyn TX, leaves: &[Leaf], level: usize) -> Result<Hash> {
```
ed255, line 626: Very smart, I like how small this function is

ed255, line 626: For bonus points, you could make this function parallel:
Let's say, if we're at a level between 0 an N, and we have more than M leaves, run each `build_subtree` in a different thread.
Where N would be log2(number of CPU threads) and M some fixed parameter tuned for the threshold when the overhead of spawining and joining is lower than the gain of parallelization.

ed255 (approved):
LGTM!

src/backends/plonky2/primitives/merkletree/mod.rs
robknight: I've added a follow-up issue https://github.com/0xPARC/pod2/issues/551

